PR #2, opened 2024-09-16
connect extension: process monitor

edavidaja:


reaper/app.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def whoami():
+    res = client.get("/v1/user")
+    me = res.json()
+    return me
+
+
+def get_content_guids():
+    owner = whoami()["guid"]
```
tdstein, line 23:
```suggestion
    owner = client.me["guid"]
```

reaper/app.py
```diff
@@ -0,0 +1,72 @@
...
+    res = client.content.find(owner_guid = owner)
```
tdstein, line 24:
```suggestion
    res = client.me.content.find()
```

edavidaja: thanks!

PR #7, opened 2024-09-23
Guide for Styling Connect Extensions

edavidaja:


nealrichardson:
Are you aware of the brand.yml project that Carlos and Garrick have been working on for Quarto, Shiny, etc.? Perhaos extensions should hook into that. https://connect.posit.it/brand-yml/

edavidaja:
I think that would be good to do, though it leaves open the question of how to style Dash, Streamlit, etc.

nealrichardson:
Streamlit isn't very customizable, IIRC.

PR #13, opened 2025-01-10
Swap Vanity URLs [DRAFT]

toph-allen:
Note: This app is a draft.


toph-allen:
It's currently hard-coded to show a single user's content. When Connect can use viewer API keys, it should be updated to use that. However, for admins, who can edit vanity URLs for all users, we could default to showing all content items. To do this, however, I'll need to add pagination to the list view (while storing the selected items), so that'll probably need some refactoring. In that case, might also want to add a toggle between "all items" and "your items".

PR #15, opened 2025-01-22
feat: publisher dashboard

tdstein:
The *Publisher Dashboard* is an experimental project that tries a different way of managing Connect projects.

The server application (`app.py`) is built using FastAPI. The server logic utilizes Connect OAuth to present the viewer with their own information instead of relying on the Content owner's credentials. When running outside of Connect (e.g., development), the environment variables `CONNECT_SERVER` and `CONNECT_API_KEY` are utilized to authenticate with Connect.

The client application  (`src/index.js`) is developed using the Mithril.js framework. Mithril is a modern client-side JavaScript framework ["used by companies like Vimeo and Nike, and open source platforms like Lichess."](https://mithril.js.org).

See `README.md` for additional developer instructions.

A production build is available at https://rsc.radixu.com/content/cd4a05a2-992a-43b9-83ab-30c4de7f302a/.

The application is intended to be utilized in "Open Solo." Ultimately, it could serve as the [`DashboardPath`.](https://docs.posit.co/connect/admin/appendix/configuration/#Server.DashboardPath)


extensions/content-manager/app.py
```diff
@@ -0,0 +1,128 @@
...
+
+app = FastAPI()
+
+client = connect.Client()
+
+
+@app.get("/api/contents")
+async def contents(posit_connect_user_session_token: str = Header(None)):
+    if posit_connect_user_session_token:
+        viewer = client.with_user_session_token(posit_connect_user_session_token)
```
mconflitti-pbc, line 15: This call should probably be cached. Saves a round trip once the app has made at least one call for that specific user session token.

```python
from cachetools import TTLCache, cached

# Create cache with TTL=1hour and unlimited size
client_cache = TTLCache(maxsize=float('inf'), ttl=3600)

@cached(client_cache)
def get_visitor_client(token: str) -> connect.Client:
    """Create and cache API client per token with 1 hour TTL"""
    return client.with_user_session_token(token)
```

mconflitti-pbc (approved):
Would be great to address the caching comment, but otherwise looks great!

extensions/content-manager/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# requirements.txt auto-generated by Posit Publisher
+# using /Users/me/Projects/connect-extensions/extensions/content-manager/.venv/bin/python
```
schloerke, line 2: Can you demo for me on Posit Publisher?

extensions/content-manager/app.py
```diff
@@ -0,0 +1,128 @@
...
+
+    content = viewer.content.get(content_id)
+    job = content.jobs.find(process_id)
+    if job:
+        job.destroy()
+        for _ in range(30):
+            job = content.jobs.find(process_id)
+            if job["status"] != 0:
+                return
+            sleep(1)
```
schloerke, line 82: If this used an async sleep, would other routes be able to work while sleeping?

schloerke: In shiny, the reactive graph is typically locked up. So there isn't a real advantage within Shiny unless you use extended tasks.

extensions/content-manager/.posit/publish/Publisher Dashboard-V7KE.toml
```diff
@@ -0,0 +1,20 @@
...
+  '/.posit/publish/Publisher Dashboard-V7KE.toml',
+  '/.posit/publish/deployments/deployment-VF7C.toml',
+  '/dist'
+]
+title = 'Publisher Dashboard'
+
+[python]
+version = '3.12.5'
+package_file = 'requirements.txt'
+package_manager = 'pip'
```
mconflitti-pbc, line 20: Does it make sense to to check in the .posit folder generally here?

Is this for user adoption or internal deployments?

extensions/content-manager/app.py
mconflitti-pbc: I think turning this into a background task would be a good optimization though maybe unnecessary for this example? But agree that at very least, this is switched to `asyncio.sleep(1)` since fastapi is using an async event loop for these requests. Pretty sure a sleep will block other stuff

extensions/content-manager/src/models/Releases.js
```diff
@@ -0,0 +1,32 @@
...
+    if (this.data) {
+      return Promise.resolve(this.data);
+    }
+
+    if (this._fetch) {
+      return this._fetch;
+    }
+
+    this._fetch = m
+      .request({ method: "GET", url: `api/contents/${id}/releases` })
```
toph-allen, line 17: This is probably a dumb question, but the Python FastAPI app is super simple — is concrete technical constraint that that the JavaScript app couldn't make requests to Connect directly? I guess in that case it'd be published as a "Static" content item, i.e. just a set of web pages available at a URL under Connect.

I guess the FastAPI app allows you to have a simple surface that you're programming against and lets you do some reshaping of the data, and I know the whole goal of this project is to use the SDKs — I'm just asking out of curiosity. :)

toph-allen (approved):
Looks v nice! I think we shouldn't check in `.posit/` probably, but I'm not going to raise a stink about it if we do. :)

extensions/content-manager/.posit/publish/Publisher Dashboard-V7KE.toml
jonkeane: > IMHO if connect-extensions is intended as a private set of example extension apps for Connect devs, this could be checked in. If this is meant to be a public-facing repo, we shouldn't check in our deployment records, as they won't be useful to anyone.

It is (and is meant to be) a  public facing repo.

tdstein: Yep, it should be removed here.

extensions/content-manager/README.md
```diff
@@ -0,0 +1,102 @@
+
+# Publisher Dashboard
+
+**Publisher Dashboard** is a web-based application designed to help publishers manage and track their content.
```
jonkeane, line 4:
```suggestion
# Publisher Content Manager

**Publisher Content Manager** is a web-based application designed to help publishers manage and track their content.
```

I know it's "just" the name but `dashboard` is so overloaded in Connect already that I don't think we want to add _yet another_ log on that fire

extensions/content-manager/app.py
tdstein: Thanks! Will do.

extensions/content-manager/requirements.txt
tdstein: Download it for VSCode or Positron. It has a walkthrough which should get you up and running.

PR #18, opened 2025-02-07
add oauth integration validator

dethmasque:
This R Shiny extension can be used with any Posit Connect viewer-based OAuth integration to validate a GET request against a given endpoint. 

Error handling around the credentials exchange request ensures that the application does not fail to start on initial deploy, when the user's refresh token expires, etc.

The results are only displayed to the user if the request succeeds; on failure a simple error message is output along with the HTTP response code, and the full error is logged for debugging.

oauth integration validator extension

### DIFF
--- a/extensions/oauth-integration-validator/app.R
+++ b/extensions/oauth-integration-validator/app.R
@@ -72,7 +72,7 @@ server <- function(input, output, session) {
       # remind the user that they don't have a token 
       output$results <- renderUI({
         HTML('<span style="color:red; font-size:larger;">No access token found.',
-        '<br/>Try logging out and logging back in to the integration if you haven not ',
+        '<br/>Try logging out and logging back in to the integration if you have not ',
         'visited this content recently.<br/>If an integration is not associated with ',
         'the content then ask your publisher to add one.</span><br/>')
       })
--- a/extensions/oauth-integration-validator/app.R
+++ b/extensions/oauth-integration-validator/app.R
@@ -1,0 +1,118 @@
+library(shiny)
+library(httr2)
+library(bslib)
+library(connectapi)
+
+ui <- page_sidebar(
+  title = "OAuth Integration Validation in R",
+  sidebar = sidebar(
+    title = "Request an endpoint",
+    textInput("endpoint", "GET endpoint to request:", placeholder = "https://example.com/api/list"),
+    actionButton("request", "Request")
+  ),
+  layout_columns(
+    card(
+      card_header("Results"),
+      htmlOutput("results")
+    )
+  )
+)
+
+server <- function(input, output, session) {
+  
+  # check if running on Posit Connect
+  if (Sys.getenv("RSTUDIO_PRODUCT") == "CONNECT") {
+    # initialize Connect API client
+    client <- connect()
+    
+    # read the user-session-token header
+    user_session_token <- session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
+    
+    # wrap client credentials exchange request for error handling
+    # and so the app does not fail to start on first deploy
+    credentials <- reactive({
+      tryCatch({
+        # exchange user session token for an access token using connectapi client
+        # checking for a client error in
+        credentials <- get_oauth_credentials(client, user_session_token)
+      }, error = function(e) {
+          return(list(error = "Unable to obtain access token: ", message = e$message))
+        })
+      })
+   
+    observeEvent(credentials(), {
+      creds <- credentials()
+      
+      # check if the credentials exchange was successful
+      if (is.null(creds$error)) {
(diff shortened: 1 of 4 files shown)


extensions/oauth-integration-validator/app.R
```diff
@@ -0,0 +1,122 @@
+library(shiny)
+library(httr2)
+library(bslib)
+library(connectapi)
+
+ui <- page_sidebar(
+  title = "OAuth Integration Validation in R",
+  sidebar = sidebar(
```
toph-allen, line 8: Perhaps a different layout like [`page_fluid()`](https://rstudio.github.io/bslib/reference/page.html) would be better? My thinking is, the sidebar's constrained width means that the pasted URL, plus the dropdowns, is always cut off. But a fluid page (which is a single column the width of the viewport) would just put the URL bar directly above the result, which means it can be wider, and since it's just a line or two it won't take up too much space. :)

extensions/oauth-integration-validator/app.R
```diff
@@ -0,0 +1,122 @@
...
+      
+      # output the unexpected HTTP code
+      if (http_code != 200) {
+        output$results <- renderUI({
+          # only display the HTTP error code to the user
+          error <- paste0("Received non-200 HTTP response code: ", http_code)
+          HTML('<span style="color:red; front-size:larger;">', error, '</span>')
+        })
+        # log the results for debugging
+        print(httr2::resp_body_json(resp))
```
toph-allen, line 115: Do you think it'd be worth displaying the full error in the app, in addition to logging it? I feel like… there's no _harm_, right? And the thing that most people are probably going to do in the case of an error is go to the log pane anyway, so we might as well show the error in the UI?

Feel free to disregard — there may be good reason to just display the error code. (Even, like, it would just be an overload of irrelevant information!)

extensions/oauth-integration-validator/app.R
```diff
@@ -0,0 +1,122 @@
...
+    endpoint <- input$endpoint
+    
+    # do not allow request without an access token
+    if (is.null(token())) {
+      # remind the user that they don't have a token 
+      output$results <- renderUI({
+        HTML('<span style="color:red; font-size:larger;">No access token found.',
+        '<br/>Try logging out and logging back in to the integration if you have not ',
+        'visited this content recently.<br/>If an integration is not associated with ',
+        'the content then ask your publisher to add one.</span><br/>')
```
toph-allen, line 77: Nice! I like the interactive debugging message. :)

I was initially going to suggest putting the bit about checking that there is an extension associated with the content first, but I guess if the app is targeted towards people who aren't viewing it in the Dashboard, this might be less relevant.

Perhaps "the content owner" instead of "your publisher" on the last line? Soft suggestion.

extensions/oauth-integration-validator/connect-extension.toml
```diff
@@ -0,0 +1,4 @@
+name = "oauth-integration-validator"
+title = "Connect OAuth Integration Validator"
+description = "Allows Connect content viewers to perform a GET request to a specified endpoint using the access token obtained by Connect."
```
toph-allen, line 3:
```suggestion
description = "Test an OAuth integration by performing a GET request to a specified URL, using the access token obtained by Connect."
```

Just took a shot at rewriting — feel free to reject or only take some parts! I was thinking… this app doesn't _just_ work for viewers, and might in fact be more useful to an admin testing that they've set up an OAuth integration correctly, or a publisher exploring how the feature works.

toph-allen (approved):
Looks good! 🦆

I left a few comments, mostly just thinking about the UI and use case of the app in a few ways. I don't think any of them are *required* to merge, just things that crossed my mind as I was reading the code and imagining the use cases. Hope they're helpful

extensions/oauth-integration-validator/app.R
dethmasque: I had imagined that this might be used by viewers of content to validate their access to given endpoints, e.g. a publisher is receiving reports that their OAuth integration content is not pulling in resources for some viewers. Instead of adding debugging functionality to their content they send this R Shiny app to the viewers experiencing issues to determine what the problem is. I thought the logs location made sense for error response output because it's accessible by the publisher and doesn't require a copy / paste of the results from the viewer to the publisher.

extensions/oauth-integration-validator/app.R
dethmasque: Modified to use `page_fluid()` and resized text input field and request button:

oauth integration validator feedback

Also incorporated your [verbiage feedback on missing access token help](https://github.com/posit-dev/connect-extensions/pull/18/files#r1950035268) and [toml file extension description](https://github.com/posit-dev/connect-extensions/pull/18/files#r1950036611).

PR #23, opened 2025-02-24
Create example of workflow with a build step

dotNomad:
This PR introduces a new GitHub Workflow for the Publisher Commander Center extension that illustrates how we can integrate a build job prior to the package job.

This is very similar to the workflows introduced in #20 with a few differences:
- There are now two jobs `build -> package`
- The build job does whatever it needs to do to create the extension code (in this case `npm run build`) and uploads an artifact for the next job to use.
- Rather than TARing up the extension's directory it TARs up the artifact from the build job

The PR also includes some changes to the `publisher-command-center` extension like:
- adding a `connect-extension.toml` which is required to install the extension via the TAR bundle
- removing the `index.html` from the `manifest.json` of the extension since it isn't needed (it uses `dist/index.html`)

### To Test

1. Grab the `publisher-command-center.tar.gz` Artifact from the "Publisher Command Center Extension" workflow.
    - This can be found by clicking "Checks" above and then "Publisher Command Center Extension" on the left. Artifacts are on the bottom of the page.
2. Install the extension using the "Add an Extension" via bundle feature on Dogfood verifying the installation and the content is working.

### DIFF
--- a/.github/workflows/publisher-command-center.yml
+++ b/.github/workflows/publisher-command-center.yml
@@ -1,0 +1,57 @@
+name: Publisher Commander Center Extension
+
+on:
+  pull_request:
+    paths:
+      - 'extensions/publisher-command-center/**'
+
+env:
+  EXTENSION_NAME: publisher-command-center
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./extensions/${{ env.EXTENSION_NAME }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+          cache: 'npm'
+          cache-dependency-path: extensions/${{ env.EXTENSION_NAME }}/package-lock.json
+
+      - run: npm ci
+      - run: npm run build
+
+      - name: Upload built extension
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.EXTENSION_NAME }}
+          path: |
+            extensions/${{ env.EXTENSION_NAME }}/dist/
+            extensions/${{ env.EXTENSION_NAME }}/requirements.txt
+            extensions/${{ env.EXTENSION_NAME }}/app.py
+            extensions/${{ env.EXTENSION_NAME }}/manifest.json
+            extensions/${{ env.EXTENSION_NAME }}/connect-extension.toml
+
+  package:
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+            name: ${{ env.EXTENSION_NAME}}
+            path: ${{ env.EXTENSION_NAME }}
+
+      - name: Create tar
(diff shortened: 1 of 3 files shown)


dotNomad:
Setting this one back to a draft, the TAR file I'm creating is larger than the 10 MB limit and we should be able to get it much smaller 👀

jonkeane:
>  is larger than the 10 MB limit and we should be able to get it much smaller 👀

Where / what is this limit?

dotNomad:
> > is larger than the 10 MB limit and we should be able to get it much smaller 👀
> 
> Where / what is this limit?

It appears when trying to use a bundle higher than 10 MB.

CleanShot 2025-02-25 at 14 12 04@2x

A bit of digging: it is based on the Content Images setting [`MaxAppImageSize`](https://docs.posit.co/connect/admin/appendix/configuration/#Applications.MaxAppImageSize).

According to the [Connect Admin Guide](https://docs.posit.co/connect/admin/content-management/index.html#content-images):

> Posit Connect rejects uploads of images larger than the [Applications.MaxAppImageSize](https://docs.posit.co/connect/admin/appendix/configuration/#Applications.MaxAppImageSize) configuration setting value. By default, the server accepts images up to 10MB in size.

aronatkins:
@dotNomad - It looks like we have incorrectly applied `Applications.MaxAppImageSize` to bundle uploads. The setting is intended to limit the size of _THUMBNAIL_ uploads, not bundles.

The `FileInput` component uses `Applications.MaxAppImageSize` as a size limit for all uses of `FileInput`. Instead, the limit should be different for different file upload contexts. Looks like we had not uncovered that problem, yet.

jonkeane:
Ha! Sneaky. And for posterity (cause I went to look): that check is in [the frontend](https://github.com/posit-dev/connect/blob/049d8c71d3134e9156adc78b65d2a0bab3cf4b69/ui/src/components/FileInput.vue#L85-L91). So presumably uploading with the API would have been fine.

You're totally right Jordan that this particular extension _also_ shouldn't be >10MB, but it's very possible that folks upload large bundles to Connect (since they might have datasets alongside them for example).

Could we spin off fixing the FileInput check for extensions into a separate issue? We _should_ fix it, but it's technically not critical path for the extensions work we're doing if they get installed with API call(s).

aronatkins:
Yes. The API does not restrict bundle upload sizes.

dotNomad:
> You're totally right Jordan that this particular extension _also_ shouldn't be >10MB, but it's very possible that folks upload large bundles to Connect (since they might have datasets alongside them for example). But this extension definitely doesn't need to be that large so yeah we should make sure we're not pulling in extraneous files, etc.

Having the method to select only necessary files is very nice so glad to have that in. Drastically lowering the file size is nice.

> The FileInput component uses Applications.MaxAppImageSize as a size limit for all uses of FileInput. Instead, the limit should be different for different file upload contexts. Looks like we had not uncovered that problem, yet.
> Could we spin off fixing the FileInput check for extensions into a separate issue? We should fix it, but it's technically not critical path for the extensions work we're doing if they get installed with API call(s).

I'll create an issue on the Connect issue board 👍

fizzboop (approved):
LGTM 🐱

PR #26, opened 2025-02-27
Audit API Extension TAR bundle

fizzboop:


dotNomad (approved):
Very nice, another TARed extension to use!

edavidaja (commented):
This seems mostly generic enough that you might just want to do it for each folder in extensions?

dotNomad:
> This seems mostly generic enough that you might just want to do it for each folder in extensions?

Some folders require different set-ups and instructions, but we will definitely be looking into making the workflows / actions more generic where possible.

fizzboop:
> This seems mostly generic enough that you might just want to do it for each folder in extensions?

> Some folders require different set-ups and instructions, but we will definitely be looking into making the workflows / actions more generic where possible.

This has been noted in the following [ticket](https://github.com/posit-dev/connect/issues/30007). Thank you for pointing it out!

PR #29, opened 2025-03-04
Extension TAR bundles

fizzboop:


.github/workflows/oauth-integration-validator.yaml
dotNomad, line 1: I wasn't able to get this TAR to work because the extension didn't have a `manifest.json`

.github/workflows/datadog-prometheus-metrics.yaml
dotNomad, line 1: This extension installed without issue, but I'm seeing the `Invalid authentication token for DD-API-KEY` error. Let's get this merged, but we should make a follow-up task to add instructions to the README, and fallback behavior for those installing it.

.github/workflows/oauth-integration-debug.yaml
dotNomad, line 1: I installed the TAR without issue, but didn't quite understand how to utilize the extension. Could be another case for follow-up for documentation.

dotNomad (commented):
Lets get a `manifest.json` for the `oauth-integration-validator` extension, and remove the workflow triggers then this should be ready to merge.

.github/workflows/oauth-integration-debug.yaml
fizzboop: Will be addressed in [#30371](https://github.com/posit-dev/connect/issues/30371)

.github/workflows/oauth-integration-validator.yaml
dotNomad: With the new manifest this is working now 🎉

dotNomad (commented):
Fantastic. Now that all of the artifacts have been checked lets remove the worklflow-triggering file changes and I'll get this approved.

PR #30, opened 2025-03-06
who deploys most often (table edition)

toph-allen:
Prototype metrics extension.

Loads data from Connect to display the following:

- Username
- full name
- number of deployments
- number of unique items deployed
- timestamp of last deploy

No reactivity or filters or anything fancy yet, just loading and processing the data.

Currently "number of deployments" uses fake data because the API does not support administrators loading bundle metadata for each content item yet.

The app is deployed on Dogfood and connect.posit.it:
- https://rsc.radixu.com/who-deploys-most-often/
- https://connect.posit.it/who-deploys-most-often/


jonkeane:
The loading time isn't super long, but it is long enough that it's perceivable and might lead someone to move on. Would it be possible (and is it advisable) to use something like https://shiny.posit.co/r/articles/improve/caching/ to improve loading performance? We can kick this to a different issue if you would like, but thinking about the experience we want for demos like this, if pulling the raw data takes this long, we want to do something to make it nicer for the people who will give us feedback.

extensions/who-deploys-most-often/app.R
```diff
@@ -0,0 +1,50 @@
...
+  )
+)
+
+server <- function(input, output, session) {
+  client <- connect()
+
+  # Load data
+  content <- get_content(client) |>
+    # Fake data, to be replaced with real data later.
+    mutate(n_bundles = rpois(n(), 3))
```
jonkeane, line 30: 💯 for making up data here. It might be a good idea to put a note on the dashboard that says this column is fake so that folks don't dig into it too much.

And to confirm my understanding, this generates a column of integers centered around 3, so on average in the resulting summary there will be 3x the number of bundles as there are unique items yeah?

extensions/who-deploys-most-often/app.R
```diff
@@ -0,0 +1,50 @@
...
+
+  user_table <- content |>
+    # The `owner` column is a nested list-column.
+    # We extract the requisite metadata up to be first-class atomic vector columns.
```
jonkeane, line 34: 💯 on doing the ugly thing here. Is this how we want this to be always? If no, do we already have an issue in `connectapi` about making this better? We don't need to (and should not) fix it before we move on with this, but I want to make sure we're capturing if we are finding new things that we need to fix in this project/proces.

extensions/who-deploys-most-often/app.R
```diff
@@ -0,0 +1,50 @@
...
+    mutate(
+      username = map_chr(owner, "username"),
+      user_full_name = paste(map_chr(owner, "first_name"), map_chr(owner, "last_name"))
+    ) |>
+    group_by(username) |>
+    summarize(
+      user_full_name = first(user_full_name),
+      n_bundles = sum(n_bundles),
+      n_content_items = n(),
+      last_deploy = format(max(last_deployed_time, na.rm = TRUE), "%Y-%m-%d at %I:%M %p")
```
jonkeane, line 44: Do we want to do the `format` here, or should we do that as part of the table presentation? Also totally fair to _not_ format in this PR at all since we will eventually make this table nicer / make it into a plot later. But generally transforms for formatting should be done at the presentation later not this layer, yeah?

toph-allen: Yeah, I only added it here because I'm not using DT to render the table yet, so it's a one-shot. You did remind me to move the format in my parallel PR to the DT layer though :)

extensions/who-deploys-most-often/app.R
toph-allen: Yeah, I think that these columns should probably just be at top level in `connectapi`. I'll make an issue to that effect.

toph-allen: I think there may be some nestedness in the `get_content()` output that is helpful because it'd add unnecessary width, but — I'll investigate that as part of that issue.

and really, I also feel like nested list columns is probably _always_ unhelpful because… like, I have to remember the best way to unnest them each time

extensions/who-deploys-most-often/app.R
toph-allen: Yeah, exactly — I used a Poisson distro because it's good for modeling count data of independent events over fixed units of observation. Poisson distribution means that it's realistically skewed towards most content items having fewer deploys and some having many more — I figured 3 is probably a reasonable mean.

Agreed about adding a label if the data are fake.

Marcos's PR for admin access to the bundles API was just added. I imagine that fetching the real data will add intense slowness (one request per content item!).

That suggests that I really _should_ investigate caching.

toph-allen:
@jonkeane I think the experience of loading the app would also be improved if there was a "loading data" dialog, as in the other demo dashboards I built. I was trying to keep this one as minimal as possible, though.

I'll look into using Shiny caching for this.

toph-allen:
@jonkeane I've made some updates based on your feedback.

- There is now a note above the table stating that `n_bundles()` uses synthetic data. Our goal is not to have this use synthetic data for long, so I didn't spend any time making it look professional or prod-ready.
- I added caching.
  - To use Shiny's `bindCache()` function, the data loading functions needed to be reactive.
  - As there are no inputs to the reactive function yet, I use a static cache key. In the future, the caches will key off date range inputs, to speed up operations.
  - Because we want caches to persist across launches of the app and different sessions, I used `cache::cache_disk()`, following the Shiny website's specific instructions for using a cache on Connect (i.e. put the cache inside the app's directory). The cache expires every 12 hours.

I also deployed the content to connect.posit.it: https://connect.posit.it/who-deploys-most-often/

jonkeane:
> As there are no inputs to the reactive function yet, I use a static cache key. In the future, the caches will key off date range inputs, to speed up operations.
> ...
> This approach to caching is so much better than the hacky pins approach I tried before, SO much simpler.

YAS this is great. The date hash key is a great idea + gets us the kind of behavior we want with the least moving pieces. If we do keep this type of caching around in later editions we might also add a button to let someone refresh the cache live, but defaulting it to the date is 💯 the right move.

extensions/who-deploys-most-often/app.R
jonkeane: I'm fine to leave this here in this PR, but if we were being strict about this PR is only what we need we probably would 🚮 this format as not needed yet. But we can 🚮 it in the next one if you would rather.

marcosnav:
Minor comment that can be addressed in subsequent PRs since this one is already closed.

The table headers names (`username, user_full_name, n_bundles, n_content_items, last_deploy`), to be more user centric.
e.g `Username, Full Name, Total Bundles Deployed, Total Content Items, Last Deploy`

PR #31, opened 2025-03-06
most used content (table edition)

toph-allen:
A prototype version of a "most used content" table.

Notes:
- This prototype includes a sortable and filterable table. Filtering is currently handled by DT's built-in string filtering.
- This content does not yet include date filters in its UI, but in the interests of running on Connect servers without the firehose endpoint, the data displayed is limited to the last week (otherwise load times are intractable).
- The code in the extension itself is parsimonious, there is a `get_usage.R` sidecar file that allows the extension to run using the dev firehose endpoint or the legacy "shiny" and "static" usage endpoints.
- Based on early feedback to https://github.com/posit-dev/connect-extensions/pull/30, I added caching using `cachem::cache_disk()`. The caches expire every 8 hours. The cache from the content API is static within that time window, and the cache from the metrics API is keyed on the day passed into the API (which has no effect now, but will when we add date filters to the UI).

The latter two points mean that the content runs successfully on Dogfood and connect.posit.it:
- https://rsc.radixu.com/most-used-content/
- https://connect.posit.it/most-used-content/

Some thoughts that are top of mind for me:
- Looking at the connect.posit.it table, I think it makes a good argument for separating anonymous and logged-in hits. In particular, look at the divergence between "Total Views" and "Unique Logged-in Users".
- The way that `DT` handles sorting, every time you click a new column, it defaults to sorting it ascending, which is not the behavior I think fits best. From my research, there aren't any built-in options to change this on a per-column basis, and I don't want to get hacky at this early stage.


toph-allen:
Looking at the connect.posit.it table, I think it makes a good argument for separating anonymous and logged-in hits.

extensions/most-used-content/get_usage.R
```diff
@@ -0,0 +1,60 @@
...
+    connectapi:::unversioned_url("instrumentation", "content", "hits"),
+    query = list(
+      from = from,
+      to = to
+    )
+  )
+
+  # FIXME for connectapi: This is slow, it's where most of the slowness is with
+  # the new endpoint.
+  usage_parsed <- connectapi:::parse_connectapi_typed(usage_raw, usage_dtype)
```
marcosnav, line 32: Interesting

marcosnav (approved):
Looks good!

extensions/most-used-content/app.R
```diff
@@ -0,0 +1,98 @@
...
+    get_content(client)
+  }) |> bindCache("static_key")
+
+  usage_data <- reactive({
+    get_usage(
+      client,
+      from = format(date_range()$from_date, "%Y-%m-%dT%H:%M:%SZ"),
+      to = format(date_range()$to_date + hours(23) + minutes(59) + seconds(59), "%Y-%m-%dT%H:%M:%SZ")
+    )
+  }) |> bindCache(date_range()$from_date, date_range()$to_date)
```
jonkeane, line 58: It feels wrong to format these dates here and not inside of `get_usage()`. Also, strictly speaking making this reactive now is extra complexity that is not needed right now. We know we will at some point do it, yes, but we should not take on the maintenance burden today for something we will do in the future. (Yes, in this sequence of PRs we expect to get to it quickly, but this is the same pattern that makes YAGNI so pernicious)

extensions/most-used-content/get_usage.R
```diff
@@ -0,0 +1,60 @@
+library(connectapi)
+
+# This file contains functions that ultimately will more likely be part of
+# connectapi. As such, I'm not using dplyr or pipes here.
```
jonkeane, line 4: Be careful with this: constraining yourself today might make putting this into connectapi easier, but that only counts for the code we do put into conenctapi. So if prototyping and getting things working is easier with dplyr/pipes/etc. do that here now and we can decide to refactor that later.

extensions/most-used-content/app.R
toph-allen: Hmm, interesting. Yes, I can see how the formatting should take place inside the function. I was lumping the formatting and date math together, but I'll separate them.

The only reason it's reactive here is so that it can be cached (only reactives can be passed into `bindCache()`).

extensions/most-used-content/get_usage.R
```diff
@@ -0,0 +1,60 @@
...
+
+NA_datetime_ <- # nolint: object_name_linter
```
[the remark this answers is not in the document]
toph-allen: Nope! I'll remove this.

extensions/most-used-content/get_usage.R
toph-allen: Fair! It took only a minute or two to switch from using pipes etc. to laying it out like this, but if it became a limiting factor I'd discard it :)

extensions/most-used-content/app.R
toph-allen: I do wonder whether these should be in the time zone of the Connect server.

toph-allen:
I wound up adding a "clear cache" button to make my own dev life easier…

extensions/most-used-content/app.R
jonkeane: > I do wonder whether these should be in the time zone of the Connect server.

This is going to end up being messy almost anyway you split it. Our API should either accept fully specified timestamps (with timezones) or only accept UTC (and our client code should convert to that form whatever is local) and then do what it needs to to make that work on the server (i.e. if the server is non-utc, then make sure the comparisons are timezone-aware and not comparing a UTC to some other timezone)

PR #32, opened 2025-03-10
who deploys most often (table edition)

toph-allen:
This PR

### DIFF
--- a/extensions/who-deploys-most-often/app.R
+++ b/extensions/who-deploys-most-often/app.R
@@ -3,12 +3,13 @@ library(bslib)
 library(dplyr)
 library(purrr)
 library(connectapi)
+library(DT)
 
 shinyOptions(
   cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 12)
 )
 
-ui <- page_fluid(
+ui <- page_fillable(
   theme = bs_theme(version = 5),
 
   card(
@@ -18,9 +19,9 @@ ui <- page_fluid(
         title = "No Filters Yet",
         open = FALSE
       ),
-      card_body("Note: `n_bundles` currently uses synthetic data."),
+      card_body("Note: \"Number of Deploys\" is using synthetic data.", fill = FALSE),
       card(
-        tableOutput("user_table")
+        DTOutput("user_table")
       )
     )
   )
@@ -36,7 +37,7 @@ server <- function(input, output, session) {
       mutate(n_bundles = rpois(n(), 3))
   }) |> bindCache("static_key")
 
-  user_table <- reactive({
+  content_by_user <- reactive({
     content() |>
       # The `owner` column is a nested list-column.
       # We extract the requisite metadata up to be first-class atomic vector columns.
@@ -47,14 +48,28 @@ server <- function(input, output, session) {
       group_by(username) |>
       summarize(
         user_full_name = first(user_full_name),
-        n_bundles = sum(n_bundles),
         n_content_items = n(),
-        last_deploy = format(max(last_deployed_time, na.rm = TRUE), "%Y-%m-%d at %I:%M %p")
-      ) |>
-      arrange(desc(n_content_items))
+        n_bundles = sum(n_bundles),
+        last_deploy = max(last_deployed_time)
+      )
   }) |> bindCache("static_key")
 
-  output$user_table <- renderTable({user_table()})
+  output$user_table <- renderDT({
+    datatable(
+      content_by_user(),
+      options = list(
+        order = list(list(3, "desc")),
+        paging = FALSE
+      ),
+      colnames = c(
+        "User" = "user_full_name",
+        "Number of Content Items" = "n_content_items",
+        "Number of Deploys" = "n_bundles",
+        "Time of Last Deploy" = "last_deploy"
+      )
+    ) |>
+      formatDate(columns = "Time of Last Deploy", method = "toLocaleString")
+  })
 }
 
 shinyApp(ui, server)
--- a/extensions/who-deploys-most-often/app.R
+++ b/extensions/who-deploys-most-often/app.R
@@ -20,18 +20,7 @@ ui <- page_fillable(
     layout_sidebar(
       sidebar = sidebar(
         title = "Filter Data",
-        open = FALSE,
-        checkboxInput("enable_date", "Filter Deploys", value = FALSE),
-
-        dateRangeInput(
-          "date_range",
-          label = "Date Range",
-          start = Sys.Date() - 7,
-          end = Sys.Date(),
-          min = "2020-01-01",
-          max = Sys.Date()
-        )
-
+        open = FALSE
       ),
       card_body("Note: \"Number of Deploys\" is using synthetic data.", fill = FALSE),
       card(
--- a/extensions/who-deploys-most-often/app.R
+++ b/extensions/who-deploys-most-often/app.R
@@ -1,23 +1,37 @@
 library(shiny)
 library(bslib)
+library(shinyjs)
+library(DT)
 library(dplyr)
 library(purrr)
 library(connectapi)
-library(DT)
 
 shinyOptions(
   cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 12)
 )
 
 ui <- page_fillable(
+  useShinyjs(),
+
   theme = bs_theme(version = 5),
 
   card(
     card_header("Who Deploys Most Often"),
     layout_sidebar(
       sidebar = sidebar(
-        title = "No Filters Yet",
-        open = FALSE
+        title = "Filter Data",
+        open = FALSE,
+        checkboxInput("enable_date", "Filter Deploys", value = FALSE),
+
+        dateRangeInput(
+          "date_range",
+          label = "Date Range",
+          start = Sys.Date() - 7,
+          end = Sys.Date(),
+          min = "2020-01-01",
+          max = Sys.Date()
+        )
+
       ),
       card_body("Note: \"Number of Deploys\" is using synthetic data.", fill = FALSE),
       card(
@@ -28,6 +42,18 @@ ui <- page_fillable(
 )
 
 server <- function(input, output, session) {
+  # Enable or disable the date range filter.
+  observe({
+    if (input$enable_date) {
+      print("enableing date range")
+      enable("date_range")
+    } else {
+      print("disabling date range")
+      disable("date_range")
+    }
+  })
+
+
   client <- connect()
 
   # Load data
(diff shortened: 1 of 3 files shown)


jonkeane:
> The bundles API does not store information about which user uploaded the bundle

If I'm understanding this correctly: this only really matters when multiple people have uploaded bundles to the same content item. Do we have any data on what percentage of items on connect.posit.it have more than one unique uploader? (even rough / not in a demo form). We might be looking at a lot of extra work for something that is functionally quite rare. It also might be the opposite and super common, but we don't currently know, right?

jonkeane:
Does DT (or any other sortable table library) allow you to filter by column? It might be nice to be able to filter things like "all people who have less than 5 uploads" rather than a global search.

toph-allen:
> Does DT (or any other sortable table library) allow you to filter by column? It might be nice to be able to filter things like "all people who have less than 5 uploads" rather than a global search.

@jonkeane Yes, it does! Thanks for prompting me to dig a little deeper.

Cc @dethmasque because you might find this useful: Adding `filter = "top"` *outside* of the `options` list adds these nice content-type-aware filter boxes on top of each column. I wound up adding `searching = FALSE` *inside* the `options` list because the "Search" field was kind of redundant at that point.

```r
    datatable(
      content_by_user(),
      options = list(
        order = list(list(3, "desc")),
        paging = FALSE,
        searching = FALSE
      ),
      filter = "top",
```
Screen Shot 2025-03-10 at 04 18 42 PM@2x

toph-allen:
> > The bundles API does not store information about which user uploaded the bundle
> 
> If I'm understanding this correctly: this only really matters when multiple people have uploaded bundles to the same content item. Do we have any data on what percentage of items on connect.posit.it have more than one unique uploader? (even rough / not in a demo form). We might be looking at a lot of extra work for something that is functionally quite rare. It also might be the opposite and super common, but we don't currently know, right?

Yeah, your statement about when this matters is correct. (On connect.posit.it we can't use this approach anyway, because it's only enabled on dev Connect, so on that server, the data are still randomly generated false data anyway.)

On Dogfood it's probably very rare. I'm not sure how common it is on connect.posit.it, and I don't think our APIs can tell us. My understanding is that we could learn how many users are collaborators on content items, but not how many bundles they have each uploaded.

On one hand, I don't think it would be a lot of extra work to use audit logs — I have the code to extract a sorted list of users by audit log activity already written in an older dashboard from last year. On the other hand, for prototyping _this_ extension… I think it might be fine to use simulated data, which might be enough to help folks imagine how useful this data might be. So I'm happy to punt on actually doing that work! :)

jonkeane:
> Yeah, your statement about when this matters is correct. ... On Dogfood it's probably very rare. I'm not sure how common it is on connect.posit.it

Do we have any sort of estimation? I agree it's probably very rare, but it would be really great to be data-driven about this rather than hunch-driven. Is there any way we could get something like this from connect.posit.it? I know that the APIs to do it the right way / the way we might need for our demo isn't possible — but are there other ways we could get at this? We should try to get some real data here if it is at all possible that we can. If that's really not possible, that's a conversation we can have with SEs who are slightly closer to our customers. @kellobri might also have thoughts.

> On Dogfood it's probably very rare. I'm not sure how common it is on connect.posit.it, and I don't think our APIs can tell us. My understanding is that we could learn how many users are collaborators on content items, but not how many bundles they have each uploaded.

FWIW, dogfood is not representative as a server generally. We don't actually want or need to analyze usage there (since it's so non-representative among other reasons). 

> On one hand, I don't think it would be a lot of extra work to use audit logs — I have the code to extract a sorted list of users by audit log activity already written in an older dashboard from last year. On the other hand, for prototyping this extension… I think it might be fine to use simulated data, which might be enough to help folks imagine how useful this data might be. So I'm happy to punt on actually doing that work! :)

If we have decided it's not useful, let's remove the madeup data (and frankly also the not-made up data in the dogfood one too?) since that will be more distracting than helpful (we want feedback on what we have, not spending time explaining why we put random data into this thing we're getting feedback on).

marcosnav:
@toph-allen Sounds good, anyway if I can help out with any server side adjustment to make things easier let me know!

I'm currently taking a look to have `uploaded_by` in bundles and `last_deployed_by` in apps, that might help, although the full history of deploys and uploads is on audit logs as you mention.

toph-allen:
> @toph-allen Sounds good, anyway if I can help out with any server side adjustment to make things easier let me know!
> 
> I'm currently taking a look to have `uploaded_by` in bundles and `last_deployed_by` in apps, that might help, although the full history of deploys and uploads is on audit logs as you mention.

I think it'd be good to have a user guid on the bundles and apps in those fields, as you mention. That would make the prior approach pretty handy.

The audit logs reveal aspects of the data that just querying the applications data wouldn't! For example, on Dogfood, there are automated tasks from `rstudio_ide_test` that seem to delete and create a new application each time, so querying the audit logs shows all the activity that just querying the current state of the application dataset doesn't! It's really interesting actually.

The audit logs are slow to load, and are fairly big, because they load lots of types of events. For them to be really useful for these types of metrics (and others) a date filter is baseline required, and being able to filter on `action` (i.e. event type) would also be good. Both would probably require an index on those fields to be performant but to a point @jonkeane made earlier, I don't actually know that! It'd also need the page size limit removed or raised, but I'm not sure if that's more complex with keyset pagination (don't want to go into the weeds in this comment but happy to discuss soon).

jonkeane:
> Updated the content. Rewrote it to drastically simplify the content itself and abstract away the audit logs caching to a separate piece of content, which I'm setting to run on a schedule. It caches 1000000 audit log entries (about 16 MB of data) to a pin.

> This means that the data only go as far back as the last 1000000 audit log entries, but it is all real data! And we have our "number of new applications deployed" and "number of individual deployments".

Nice, does this work on connect.posit.it too? Maybe I was wrong, but I thought we were limited to needing to be admin in order to get at the audit logs?

> The audit logs reveal aspects of the data that just querying the applications data wouldn't! For example, on Dogfood, there are automated tasks from rstudio_ide_test that seem to delete and create a new application each time, so querying the audit logs shows all the activity that just querying the current state of the application dataset doesn't! It's really interesting actually.

These are the automated integration tests that the ide team runs, right?

> The audit logs are slow to load, and are fairly big, because they load lots of types of events. For them to be really useful for these types of metrics (and others) a date filter is baseline required, and being able to filter on action (i.e. event type) would also be good. Both would probably require an index on those fields to be performant but to a point @jonkeane made earlier, I don't actually know that! It'd also need the page size limit removed or raised, but I'm not sure if that's more complex with keyset pagination (don't want to go into the weeds in this comment but happy to discuss soon).

Before we go and dig too deep down this rabbit hole, are there bits of information that we are using that we would rather put elsewhere in the API? Especially if the audit logs API is admin-required, and we want these to be useful to non-admins, we'll need them elsewhere anyway. It would be good to have an understand of what data in there is useful and if it has a natural home to expose elsewhere in the API.

marcosnav:
@jonkeane I'm taking a look on ways I can help making data more accessible, outside the audit logs, however there are a few ways we can approach this issue.

I have one question, can you elaborate a bit more on how publishers could benefit from "who deploys most often"? I've been thinking about this prototype as "Admin centric" since it is meant to be a ranking list of all publishers for all content items, I think naturally publishers wouldn't have access to all that data.

Asking because we could:
a) Extend (or create) our current `/content` and `/bundles` API endpoints
b) Modify audit logs endpoints to allow publishers to get the data for the content they have access to. (also adding from/to filters)

toph-allen:
@jonkeane:

- I *am* an admin on `connect.posit.it` 😎. Specifically for the purposes of running this type of content.
- Yes those are automated tests. There are other interesting discontinuities between number of new apps created and active apps — I found it really interesting as it speaks to the different ways of asking questions of the data on Connect.

Commenting on the discussion

I see Marcos's point of view — the question of "who publishes most often" is inherently going to require data available only to Admins, as it's kind of an Admin-focused dashboard. Even if we were gonna use the `bundles` API to count up information about bundles on Connect, the data on other users' bundles isn't available to Publishers, so the admin focus really is a property of this particular question.

To elaborate on the approaches that @marcosnav describes w/r/t the data that this extension shows:

- Extending and/or adding `/content` and `/bundles` endpoints:
  - Answers the question posed in this content from the perspective of "What's active on Connect right now?", which may or may not be what admins want to know — we'll have to get feedback on that.
  - Explicitly tied to content items that you can visit, get more data on, etc.
  - We really _should_ store the uploader's `user_guid` on each bundle, it seems like proper data to store ¯\_(ツ)_/¯
  - The current structure of the `bundles` endpoint, living below `content`, would be really slow to gather this kind of data, as you have to make a request for each content item. To make it performant, we'd need to add a top-level `bundles` endpoint, which is a larger task.
 
- Updating the `audit_logs` endpoint:
  - Answers the question posed in this content from the perspective of, "What is the historical activity on this Connect server", i.e. better reflecting the history of Connect rather than its current state. This might be more relevant, especially if admins are trying to track down e.g. "what caused all those issues x days ago?" but again we need that input.
  - Currently only available to admins. Since `user_guid` is one of the database fields, it'd be possible to provide relevant info to publishers, but this seems less aligned with the kind of data the audit logs offer, and it doesn't strike me as important.
  - For it to be really great to answer this question, we'd want to add date filters and possibly filters on the `action` field — again, all fields in the SQL DB schema already.
  - Only _implicitly_ tied to individual content items, and to do so would depend on possibly if I think if you have your audit log format configured as JSON there is other structured data logged, but I don't see that on Dogfood lol ([link](https://docs.posit.co/connect/admin/logging/#audit-log-format)) or textual munging of log data. For the purposes of this dashboard, that's totally fine, though — it's answering questions about aggregated user activity.

jonkeane:
> I have one question, can you elaborate a bit more on how publishers could benefit from "who deploys most often"? I've been thinking about this prototype as "Admin centric" since it is meant to be a ranking list of all publishers for all content items, I think naturally publishers wouldn't have access to all that data.
> [+ some of Toph's points above]

_nods_ yeah if one wanted a full view of the server, needing to be an admin makes sense. There still is value for publishers to be able to for the content they see get an idea of who is active. Though totally fair that this isn't a super big question-people-are-asking-now (so 💯 on buttoning the prototype up, getting it in front of SE and others and confirm "this is cool, but not all the useful yeah?" and moving on to the ones that do land better — but being open to learning that there's some other thing in this data we don't realize that we will get form feedback).

> Asking because we could:
> a) Extend (or create) our current /content and /bundles API endpoints
> b) Modify audit logs endpoints to allow publishers to get the data for the content they have access to. (also adding from/to filters)

Either is fine as long as the information is getable. Then it's a question of which lets us make the clients (and the code that uses the clients) neatest and most natural.

toph-allen:
@jonkeane @marcosnav I'm going to merge this shortly so I can continue on the next iteration. This conversation is valuable, but I haven't actually received any reviews on this PR yet.

toph-allen:
All good! I just wanna make sure I keep moving forward :)

PR #33, opened 2025-03-10
content with issues prototype: table view

dethmasque:
[Deployed to Dogfood](https://rsc.radixu.com/connect/#/apps/1cf75066-007f-4f54-af28-f21ea6e5e262) (cache builds in ~20m on new deploy and every 8h)
[Deployed to connect.posit.it](https://connect.posit.it/connect/#/apps/bc3dc19c-6c49-4d8b-9db0-c2e629a050b9/) (cache builds in ~1h on new deploy and every 8h)

Fixes https://github.com/posit-dev/connect/issues/30288 

See [posit-dev/connect issue](https://github.com/posit-dev/connect/issues/30288) for requirement details (further feedback in comments).

This R Shiny prototype extension checks all jobs on content available to the user associated with the provided API key, and compiles failed jobs data along with the last visit information into a static table.

### DIFF
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -117,7 +117,7 @@ server <- function(input, output, session) {
     map_dfr(content_list(), ~ get_failed_job_data(.x, usage()))
   }) |> bindCache("static_key")
   
-  # output the table on load, wrapping in render UI 
+  # output the datatable of failed jobs
   output$jobs <- renderDT({
     datatable(bad_content_df(), 
               rownames = FALSE, 
@@ -142,7 +142,7 @@ ui <- fluidPage(
     
   fluidRow(
     column(12,
-           titlePanel(tags$h6("All content with a failed job:")), 
+           titlePanel(tags$h6("All failed content jobs:")), 
            DTOutput("jobs"),
     )
   )
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -44,17 +44,29 @@ get_failed_job_data <- function(item, usage, client, connectServer) {
       get_jobs(item) |> 
         # filter successful jobs
         filter(exit_code != 0) |> 
-        # map job types to something more readable 
+        # map content job types to something more readable 
         mutate(tag = case_when(
-               tag == "run_app" ~ "Run R Application",
-               tag == "packrat_restore" ~ "Packrat Cache Restore",
+               tag == "unknown" ~ "Unknown Job Type",
                tag == "build_report" ~ "Build Report",
                tag == "build_site" ~ "Build Site",
                tag == "build_jupyter" ~ "Build Jupyter Notebook",
+               tag == "packrat_restore" ~ "Packrat Cache Restore",
                tag == "python_restore" ~ "Python Envrionment Restore",
+               tag == "configure_report" ~ "Parametrized Report Configure",
+               tag == "run_app" ~ "Run R Application",
+               tag == "run_api" ~ "Run Plumber API",
+               tag == "run_tensorflow" ~ "Run Tensorflow API Model",
+               tag == "run_python_api" ~ "Run Python API",
+               tag == "run_dash_app" ~ "Run Dash Application",
+               tag == "run_gradio_app" ~ "Run Gradio Application",
+               tag == "run_streamlit" ~ "Run Streamlit Application",
+               tag == "run_bokeh_app" ~ "Run Bokeh Application",
                tag == "run_fastapi_app" ~ "Run Python FastAPI",
-               tag == "run_gradio_app" ~ "Run Python Gradio",
                tag == "run_pyshiny_app" ~ "Run PyShiny Application",
+               tag == "render_shiny" ~ "Render Quarto or RMarkdown Interactive Document",
+               tag == "run_voila_app" ~ "Run Voila Application",
+               tag == "run_pyshiny_app" ~ "Run PyShiny Application",
+               tag == "ctrl_extraction" ~ "Extract Parameter Controls From Report",
                TRUE ~ tag)) |>
         # map exit codes to something more readable 
         mutate(exit_code = as.character(exit_code)) |>
@@ -64,6 +76,10 @@ get_failed_job_data <- function(item, usage, client, connectServer) {
                exit_code == "137" ~ "Application terminated due to low memory",
                exit_code == "2" ~ "Command encountered an error during execution",
                exit_code == "134" ~ "Terminated due to critical error, potential segfault",
+               exit_code == "255" ~ "Abnormal termination of process",
+               exit_code == "130" ~ "Process terminated by user-initiated signal",
+               exit_code == "13" ~ "File permissions issue",
+               exit_code == "15" ~ "Process terminated externally",
                TRUE ~ exit_code))
     },
     error = function(e) {
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -1,0 +1,178 @@
+library(shiny)
+library(bslib)
+library(DT)
+library(connectapi)
+library(dplyr)
+library(purrr)
+
+# cache data to disk with a refresh every 8h
+shinyOptions(
+  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
+)
+
+source("get_usage.R")
+
+# Hacky function to get a list of Content class objects without making a request
+# for each item. These objects differ from the ones created by `content_item()`
+# because they also include the full owner info as returned by `get_content()`.
+as_content_list <- function(content_df, client) {
+  cdf_split <- split(content_df, 1:nrow(content_df))
+  map(cdf_split, function(x) {
+    x <- x[, !(names(x) %in% c("tags"))]
+    x <- as.list(x)
+    Content$new(client, x)
+  })
+}
+
+# grab the username associated with a provided user_guid
+get_user_info <- function(client, user_guid, connectServer) {
+  if (is.na(user_guid)) {
+    # NA user_guid populates when ACL is no login required / public access
+    as.character("Anonymous")
+  } else {
+    user_endpoint <- paste0("v1/users/", user_guid)
+    user <- client$GET(user_endpoint)
+    user$username
+  }
+}
+
+# checks to see if a content item has failed jobs, grabs usage data if it does,
+# then compiles content, job, and usage data together, returning it.
+get_failed_job_data <- function(item, usage, client, connectServer) {
+  failed_jobs <- tryCatch(
+    {
+      get_jobs(item) |> 
+        #slice_max(start_time) |> # don't filter to only latest job
+        filter(status != 0 && exit_code != 0) |>
+        # map job types to something more readable 
+        mutate(tag = case_when(
+               tag == "run_app" ~ "Run R Application",
+               tag == "packrat_restore" ~ "Packrat Cache Restore",
+               tag == "build_report" ~ "Build Report",
+               tag == "build_site" ~ "Build Site",
+               tag == "build_jupyter" ~ "Build Jupyter Notebook",
+               tag == "python_restore" ~ "Python Envrionment Restore",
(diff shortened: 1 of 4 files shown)


extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+# cache data to disk with a refresh every 8h
+shinyOptions(
+  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
+)
+
+source("get_usage.R")
+
+# Hacky function to get a list of Content class objects without making a request
+# for each item. These objects differ from the ones created by `content_item()`
+# because they also include the full owner info as returned by `get_content()`.
```
jonkeane, line 17: This sounds like a feature request for `connectapi` have you raised an issue there?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+      slice_max(timestamp) %>%
+      select(user_guid, timestamp)
+    # some content with issues do not have visit data (ex: unpublished)
+    if (nrow(last_visit) == 0){
+      visit_timestamp <- as.POSIXct(0) 
+      visitor_username <- "No visit data exists for this content"
+    } else {
+      visit_timestamp <- last_visit$timestamp
+      visitor_username <- get_user_info(client, last_visit$user_guid, connectServer)
+    }
```
jonkeane, line 108: Do we need the last visited by? IF we don't add that here would this and the above if branch be much simpler?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+
+server <- function(input, output, session) {
+  # use API key environment variables for now
+  # we can add local run and connect API integration functionality later
+  api_key <- Sys.getenv("CONNECT_API_KEY")
+  connect_server <- Sys.getenv("CONNECT_SERVER")
+  # initialize Connect API client
+  client <- connect(server = connect_server, api_key = api_key)
+  # set dashboard URL, needed for call to v1/users for visitor user_guid replacement
+  connectServer <- client$get_dashboard_url()
```
jonkeane, line 139: Doesn't `connectapi` use these by default? Do we need to load them + pass them like this?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+               tag == "run_dash_app" ~ "Run Dash Application",
+               tag == "run_gradio_app" ~ "Run Gradio Application",
+               tag == "run_streamlit" ~ "Run Streamlit Application",
+               tag == "run_bokeh_app" ~ "Run Bokeh Application",
+               tag == "run_fastapi_app" ~ "Run Python FastAPI",
+               tag == "run_pyshiny_app" ~ "Run PyShiny Application",
+               tag == "render_shiny" ~ "Render Quarto or RMarkdown Interactive Document",
+               tag == "run_voila_app" ~ "Run Voila Application",
+               tag == "run_pyshiny_app" ~ "Run PyShiny Application",
+               tag == "ctrl_extraction" ~ "Extract Parameter Controls From Report",
```
jonkeane, line 69: I wonder if it would be better to bundle these into "Building", instead of "Build Report", "Build Site", and "Build Jupyter Notebook" and then: "Restoring environment", "Running" are there other meta categories we need?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+        mutate(exit_code = case_when(
+               exit_code == "1" ~ "Ended with errors",
+               exit_code == "127" ~ "Command could not be executed outside container",
+               exit_code == "137" ~ "Application terminated due to low memory",
+               exit_code == "2" ~ "Command encountered an error during execution",
+               exit_code == "134" ~ "Terminated due to critical error, potential segfault",
+               exit_code == "255" ~ "Abnormal termination of process",
+               exit_code == "130" ~ "Process terminated by user-initiated signal",
+               exit_code == "13" ~ "File permissions issue",
+               exit_code == "15" ~ "Process terminated externally",
```
jonkeane, line 82: Could we collapse some of these into ones that are more user-focused? I would think that something like: 1, 2, and 134 are all a group of "failed to run / error during running"; 137 is "out of memory"; 255, 130, and 15 are "process terminated by server"; and 13 and 127 are suggestive of configuration errors? Those are slightly more actionable for users (or at least that first and second group are, the others are "talk to your sysadmin")

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+      # content item does not have any jobs 
+      NULL
+    }
+  )
+  
+  if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
+    return(NULL)
+  } else {
+    print(item)
+    print(failed_jobs)
```
jonkeane, line 95: Do we need these prints?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,197 @@
...
+              )
+            )
+  })
+}
+
+ui <- fluidPage(
+  tags$head(
+    # icons for pretty display 
+    tags$link(rel = "stylesheet", href = "https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.3/css/all.min.css")
+  ),
```
jonkeane, line 179: Is this the right way to do something like this? I thought font awesome was bundled already? Also would `includeCSS()` be better than adding a tag to the head?

jonkeane: And, also, presentation layer like fonts like this is a premature optimization at this point, no?

extensions/prototype-content-with-issues-table/app.R
dethmasque: I'll check to make sure @toph-allen doesn't have an issue in the works, as I pulled this helpful function from him during code pairing, and will open one if not.

extensions/prototype-content-with-issues-table/app.R
dethmasque: Last visited can be only the timestamp of the visit, as admins and publishers could always dig deeper into visit data if they notice the failing content was visited. Removing considering the user GUID and resolving the associated username would clean up logic and get rid of get_user_info(), so let's do it.

extensions/prototype-content-with-issues-table/app.R
dethmasque: [Looks like it does!](https://github.com/posit-dev/connectapi/tree/ea0a1401d5ca30a5faabbbc48026a8b832c591ff?tab=readme-ov-file#client) `connectServer` is also cleaned up with the removal of `get_user_info()` as described in [another comment](https://github.com/posit-dev/connect-extensions/pull/33#pullrequestreview-2672118035), so this entire block will be tidied.

extensions/prototype-content-with-issues-table/app.R
dethmasque: Noted your feedback on bundled FA and `includeCSS()`, and will remove lingering styling.

dethmasque:
@jonkeane Done! Gmail automatically filtered the access request emails out of the Inbox and into the Updates category, fixing those settings now so I don't miss them in the future.

dethmasque:
Thank you @marcosnav! I will iterate on readable column headers in #34 as I already have to pull in feedback from this PR into it.

jonkeane:
> Thank you @marcosnav! I will iterate on readable column headers in https://github.com/posit-dev/connect-extensions/pull/34 as I already have to pull in feedback from this PR into it.

We should (have) resolved these comments in this PR. Also FWIW: I haven't actually had a change to even look at the example yet. I will come back here with any feedback I have from seeing it actually live.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,152 @@
...
+                          "run_dash_app",
+                          "run_gradio_app",
+                          "run_streamlit",
+                          "run_bokeh_app",
+                          "run_fastapi_app",
+                          "run_voila_app",
+                          "run_pyshiny_app") ~ "Running",
+               tag == "render_shiny" ~ "Rendering",
+               tag == "ctrl_extraction" ~ "Extracting parameters",
+               TRUE ~ tag)) |>
```
jonkeane, line 53: What do you think about moving this munging into the display code rather than the querying code?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,152 @@
...
+        # map exit codes to something more readable 
+        mutate(exit_code = as.character(exit_code)) |>
+        mutate(exit_code = case_when(
```
jonkeane, line 56: Do we need to turn this into characters at all?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,152 @@
...
+    x <- x[, !(names(x) %in% c("tags"))]
+    x <- as.list(x)
+    Content$new(client, x)
+  })
+}
+
+# checks to see if a content item has failed jobs, grabs usage data if it does,
+# then compiles content, job, and usage data together, returning it.
+get_failed_job_data <- function(item, usage) {
+  failed_jobs <- tryCatch(
```
jonkeane, line 30: What are we catching here?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,152 @@
...
+  if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
+    return(NULL)
+  } else {
+    # handle content without usage data, such as unpublished content
+    last_visit <- usage %>%
+      filter(content_guid == item$content$guid) %>%
+      slice_max(timestamp) %>%
+      select(timestamp)
+    if (is.na(item$content$title)) {
+      item$content$title <- "" # use empty strings when content is missing title
```
jonkeane, line 78: Would [`replace_na()`](https://tidyr.tidyverse.org/reference/replace_na.html) do what you're looking for here?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -0,0 +1,152 @@
...
+    }
+    # return required information from https://github.com/posit-dev/connect/issues/30288 
+    all_failed_jobs <- bind_rows(lapply(seq_len(nrow(failed_jobs)), function(i) {
```
jonkeane, line 81: I'm struggling to follow what you're doing here. You're going down the vector of failed jobs, and for each one pulling out details from that list and then enriching it with information about the content item itself, yeah? 

If yes (but I'm honestly not sure if this is true or not)... Would it be possible to start from `failed_jobs` and then use dplyr / vector recycling to do this instead?

extensions/prototype-content-with-issues-table/get_usage.R
```diff
@@ -0,0 +1,58 @@
...
+  shiny_usage_cols$timestamp <- shiny_usage$started
+
+  static_usage <- get_usage_static(client, limit = Inf, from = from, to = to)
+  static_usage_cols <- static_usage[c("user_guid", "content_guid")]
+  static_usage_cols$timestamp <- static_usage$time
+
+  bind_rows(shiny_usage_cols, static_usage_cols)
+}
+
+get_usage <- function(client, from = NULL, to = NULL) {
```
jonkeane, line 47: Here, but also elsewhere in these functions: you've got `from`/`to`, but are you using those anywhere here?

jonkeane (commented):
I've got a few changes requested here. I'll leave it up to you, @dethmasque  if you want to make a separate issue + PR to follow on or if you want to revert the merge and then reopen this PR.

extensions/prototype-content-with-issues-table/get_usage.R
toph-allen: (@jonkeane Just a side-note: I shared this function with @dethmasque `get_usage()` as it abstracts away the differences between the new firehose and older usage endpoints ([ref1](https://github.com/posit-dev/connect-extensions/pull/31), [ref2](https://github.com/posit-dev/connect-extensions/pull/31#discussion_r1985415947)). The `from` and `to` are there because in my extension, I decided to limit the query to the most recent week even before adding explicit date filters, for performance reasons.)

PR #34, opened 2025-03-10
content with issues prototype: interactive table 

dethmasque:
A continuation of #39 that follows the requirements outlined in https://github.com/posit-dev/connect/issues/30333 and includes feedback from SE to use the `gt` package, as well as [requests to simplify `purrr::map_dfr()` usage into a `dplyr` pipeline](https://github.com/posit-dev/connect-extensions/pull/39#discussion_r2005972992).

[Deployed to connect.posit.it](https://connect.posit.it/content-issues-interactive/) 

[Deployed to Dogfood](https://rsc.radixu.com/content-with-issues-interactive-table/)

Fixes https://github.com/posit-dev/connect/issues/30333

This R Shiny prototype extension checks all jobs for failures on content deployed within the last year and compiles information into an interactive table. 

<details><summary> GIF demo </summary>

content with issues interactive table prototype with help and filters

</details>


marcosnav (commented):
I'm still taking a look at this but I have a couple suggestions already:
1. The table headings, I'm a stubborn calf 😅 , it'll be nice for those to be in a human friendly format
2. Since this is the interactive table version, and considering previous feedback from SE, it'll be nice for the title and guid columns to be links to the failing content (maybe even the whole row? something like hovering on it hints that you can click and go?)

jonkeane:
Some app-level comments (I haven't yet dug through the code, but a few observations from poking at the app itself):

* Could we mark which of these errors are errors that are happening still / on the latest deployed bundle versus those that have had errors in the past, but are now just fine? This would also be something that would be fantastic to be able filter and toggle between "show me all errors" versus "show me errors that are still happening" are slightly different paths someone might go down.
* Could we collapse the rows by GUID (or maybe bundle?), so multiple failures for the same item (or same bundle) are together? This might also help with ^^^ (but also might complicate it, depending on how you implement it)
* Do we need the search box if we have the column by column search?
* Super minor: is there a way to reset the ordering after I click on a column? I assume this is something in {gt}, but I was a little surprised that I can't seem to get back to whatever the default order is without a reload of the page

jonkeane:
I'm a little bit surprised that this is in a separate folder from the original table edition. I was hoping that this would be in the same folder in this repository so that we had the diff between the static table and the interactive table more obvious (for now, it looks like we wrote all of this code bespoke for this ticket which isn't how we did it, but also part of our somewhat-artificial breakdown of these tasks was to use this as practice for iterating on the same piece of code). Would it be possible to have these changes be against the same files as https://github.com/posit-dev/connect-extensions/tree/main/extensions/prototype-content-with-issues-table ?

dethmasque:
Thank you @jonkeane and @marcosnav! I addressed your feedback in a recent commit and redeployed the example to both Dogfood and connect.posit.it. 

> Could we mark which of these errors are errors that are happening still / on the latest deployed bundle versus those that have had errors in the past, but are now just fine? This would also be something that would be fantastic to be able filter and toggle between "show me all errors" versus "show me errors that are still happening" are slightly different paths someone might go down.

We could highlight failed jobs that are the latest job ran for a given content item, noting that the content has not recovered.

dethmasque:
@jonkeane Content that has not recovered (latest job is a failure) is now marked with a red warning icon. 

Regarding grouping by bundle instead of content GUID: bundle information isn't surfaced in jobs data, so we would need to compare audit log `deploy_application` actions to see which bundle ID was active at the time the failed job began. There has been [a discussion around including bundle data, as well as only surfacing failed jobs on content that hasn't recovered, on the `v1/content/failed` issue](https://github.com/posit-dev/connect/pull/30414/files#r2004109077)

dethmasque:
> This is nice, but is it possible to make it so that I can filter on that? Either in a column, or with a toggle switch outside the table that does the filtering. For most / many use cases, folks only care about things that are right now a problem. If it failed int he past but is fine now they can (generally) ignore it.

Sure thing!

> On the grouping by-item front: I like the collapsing that is there, but will admit that at first I didn't totally understand that how they were connected. I wonder if what we should do is actually have a table of items (e.g. one row per GUID) and then the row expands to include details about the failed jobs associated with that item. The details of those failures don't need to be the same as the columns in the larger table — though some of the details might get summarised up to that level like number of total failures or time of last failure.

I'll look into collapsing the grouped rows by default. 

Additional feedback from SE that will be incorporated into this PR:

- Add information boxes that describe:
  - what jobs are tracked here (environment restores, report renders, etc), define them 
  - what jobs are not tracked here (Git, failed deploys, etc), define them
- Filter by jobs where the content owner was not notified via email
- Filter by jobs not initiated by the content owner
- Filter by level of visibility of failure
- Remove Last Visited and Last Deployed data

jonkeane (commented):
I know you're still working on the broader/product/UX level feedback, but in the mean time, here are a few code comments

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -31,35 +30,52 @@ as_content_list <- function(content_df, client) {
...
-        # filter out successful and running jobs
-        filter(exit_code != 0 & status != 0 & !(is.na(end_time)))
-  
+  jobs <- tryCatch(
+    {
+      get_jobs(item) 
+    }, error = function(e) {
+      print(paste("Error encountered with item: ", item))
+      NULL
+    })
```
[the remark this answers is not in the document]
dethmasque: 404s when trying to gets jobs from now-deleted content items, as well as 403s when the ACL is updated to exclude the user, are caught with this. There were some malformed content items on Dogfood but this `tryCatch()` was brought back when a 404 was encountered while retrieving jobs data on connect.posit.it following duplicated content removal post-deploy.

I can improve the error here and have it reflect that the content item was deleted or access was restricted after it was returned by `get_content()`, including the returned message in the logline.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -31,35 +30,52 @@ as_content_list <- function(content_df, client) {
...
+    # check is the content is still failing
+    if (latest_job$key %in% failed_jobs$key) {
+      recovered = FALSE  
+    } 
+      recovered = TRUE
+    }
   if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
     # content item does not have failed jobs
-    return(NULL)
+    NULL
```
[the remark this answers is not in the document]
dethmasque: We execute this function against a list of content items that are iterated over using `purrr::map_dfr()` in order to compile the dataframe of all failed content jobs. If an item does not have any failed jobs we return `NULL` here and nothing is added to the dataframe.

extensions/prototype-content-with-issues-table/app.R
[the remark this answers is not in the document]
jonkeane: > 404 was encountered while retrieving jobs data on connect.posit.it following duplicated content removal post-deploy.

Is this a common state that content is in?

 > I can improve the error here and have it reflect that the content item was deleted or access was restricted after it was returned by get_content(), including the returned message in the logline.
 
This is a bit of a surprising state that we are in: having access at one point and then not at another. I know functionally how this can happen, but practically speaking this is not something that should be common. Is there something about where we are getting these two sets of information from that is making this more likely? Or maybe we have the content info in a long-lived cache and then we try much later to poke at it and don't have access?

We might need to keep this here for the purposes of having a responsive demo, but we should comment our findings to ^^^ heavily, because this is a funk we don't want to bring with us as we make these less prototypes and more things that will be part of the gallery.

extensions/prototype-content-with-issues-table/app.R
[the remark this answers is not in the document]
jonkeane: Right, but if I'm reading this correctly with the change in this PR this function *does not* return `NULL` here any more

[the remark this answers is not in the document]
jonkeane: Oh, actually no, I see now. I had to pull this code to see what's actually going on here: you've got the `if`/`else` with each branch returning a thing (implicitly, as the last line of the function) right? That's what you're aiming for?

If so, I would recommend we do one of the following to make this clearer (I missread this code and was confused twice when looking at it, that's a sign something is off!):

1. return early in the if, then have the main logic follow (and frankly storing as `all_failed_jobs` versus having the dply pipe be at the end is a stylistic thing, either is fine.
```
...
  if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
    # content item does not have failed jobs
    return(NULL)
  }
  
  all_failed_jobs <- failed_jobs %>% 
    mutate(
      content_title = item$content$title,
      content_guid = item$content$guid,
      content_owner = item$content$owner[[1]]$username,
      content_url = item$content$dashboard_url
      )
        
  all_failed_jobs
} 
``` 

2. assign either of these explicitly, then return that (you can also do `all_failed_jobs <- ` inside the branches

```
...
  all_failed_jobs <- if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
    # content item does not have failed jobs
    NULL
  } else {
    failed_jobs %>% 
      mutate(
        content_title = item$content$title,
        content_guid = item$content$guid,
        content_owner = item$content$owner[[1]]$username,
        content_url = item$content$dashboard_url
        )
    }
        
  all_failed_jobs
}
```

Added whitespace will also make this easier to follow what's up

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -27,39 +29,51 @@ as_content_list <- function(content_df, client) {
...
-# usage data if it does, then compiles content, job, and usage data together 
-# into a tibble, returning it.
-get_failed_job_data <- function(item, usage) {
-  failed_jobs <- get_jobs(item) |>
-        # filter out successful and running jobs
-        filter(exit_code != 0 & status != 0 & !(is.na(end_time)))
-  
+# filters content jobs down to failures and sets content_recovered depending on 
+# whether or not the latest job ended in error
+filter_jobs <- function(jobs) {
```
jonkeane, line 34: Minor, but could we give this a more descriptive name? `filter` could be filtering on a bunch of different dimensions.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -27,39 +29,51 @@ as_content_list <- function(content_df, client) {
...
+      # filter out successful and running jobs 
+      filter(exit_code != 0 & !(is.na(exit_code)) & status != 0 & !(is.na(end_time))) |>
+      # grab only the columns we use for cleaner dplyr pipeline
+      select(end_time, exit_code, tag, key) 
+    # set content_recovered depending on if latest_job was in failed_jobs 
+    failed_jobs %>%
+      mutate(
+        content_recovered = ifelse(latest_job$key %in% failed_jobs$key, FALSE, TRUE)
+      )
+  }
```
jonkeane, line 52: Same comment as https://github.com/posit-dev/connect-extensions/pull/34/files#r2012130724 if you're returning `NULL` when there are no jobs, return early or save both of these branches to a variable and return that one. 

Alternatively, seeing this in a function: would it be possible to have the no-jobs path return a data.frame with zero rows (and make the rest of the code robust to that)? That might be a more natural approach and have fewer of these if/else branches. It also has the nice property that the functions all return _one_ thing (a data.frame with a specific set of columns) rather than two things (either a data.frame or `NULL`, depending). This isn't _required_ in this PR, but something to think about + might clean this code up and make it easier to follow.

dethmasque:
Noting SE feedback here:

#### Improve discoverability of features and help information

> I appreciate the "What is included here?" section though I didn't know to click on it at first. I have more questions though after reading this. I don't know that I could give a clear description without going to look up the difference between an environment restore, a build, and a render, for example. The description of the types of jobs isn't consumable enough yet to fully understand.

> Similarly, what is a bundle activation? and why are failed deployments not in the report if a failed build or environment restore job is in the table? and wouldn't content that failed a job like this be unpublished and therefore in the dashboard?
Helpful to add to the "not included" section would be a discussion on where or how to find that information. What I also expected to see in "not included" is a specific callout that failures in git-backed deployments (can't find branch or couldn't connect to repo) are not included.

> when i see this list, I wonder what I need to go fix now because it's not running.  You've got that info here, I just didn't realize it.  "Unrecovered content" doesn't scream "broken state" to me.

#### UI/UX improvements

> these filter options weren't intuitively linked to the table. It would be nicer if the filter boxes in the table were drop downs for Reason for Failure and Job Type.

> clicking the content title should open the content in a new window, not same window.

> helpers: add an email button to owner's name.  Add a link to the content logs.  And bonus points - add a button to extract selected rows for emailing or downloading.  Imagine the admin wanting to select the records for a particular piece of content and emailing them to the owner. with a a link to the content item and logs.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -69,37 +99,29 @@ server <- function(input, output, session) {
   client <- connect()
   
   # TODO: use `v1/content/failed` when #30414 merges so we only list content we
-  # know has failed before, filter to deployed within last 60d for now
+  # know has failed before, filter to deployed within last year for now
```
marcosnav, line 102: It landed as an unversioned endpoint now 🙌 , I'm eager to see how much a difference does it make

jonkeane:
Thanks for all the refinements on the app. I poked at the posit.it one a bit and I definitely will underscore this from Katie:

> these filter options weren't intuitively linked to the table. It would be nicer if the filter boxes in the table were drop downs for Reason for Failure and Job Type.

It also was a little frustrating that changing the filters in the left forced the whole table to re-render. So if I had already filtered down to all of my content with search, and thing clicked "Owner not notified" the filtering disappeared. This might be too big to fix now, but we will want to do that in the production version of this.

More minor things from actually using the app:

Could we do filtering at the column level? search overall is nice, but I usually know i want to search for a _title_ with ... or I want to search for someone's username (again, similar to what Katie said). 


The name "broken content" isn't super transparent to me. Maybe something like "currently failing" might be more evocative?  I know there is help, but even that could use some more details (and especially that the opposite of this is something that has failed at some point, but the most recent deploy/run was successful, that might help.)

> Broken content: display items where the latest job ended in failure. 

Something is up with the title of the page, it has the html tags in it: 
image

I know I've asked this before, but: I still wish that even when the rows are collapsed, they showed information from the most recent failure. I actually think most people would want to see that view and then only when something is interesting drill down and see the N failures for that app or dashboard. But currently if you collapse all of them, there is very little information at all.


I'll do an actual code review next, but wanted to get this feedback out

extensions/prototype-content-with-issues-table/app.R
jonkeane: We talked about this in our 1:1 but to document what we talked about: something like `filter_failed_jobs()` would be evocative. The comment about this function signature is great, but the function name is important for communicating later on when we actually call the function. Yes, you can look up the function and find this, but that's one extra step when we are reading the code.

dethmasque:
Noting additional feedback from SE on prototype shared yesterday:

#### Improve sorting behavior 

> wow! awesome iteration!  I think additional feedback from me is around tablecrafting. For example, clicking on the column headers to sort yields unexpected results because the table is grouped by content. 

#### Add filter for content  owner 

> I think I'd like an explicit filter on the LHS for owner. 

#### Better reflect applied filter state and improve filter descriptions in help information

> And I started wondering about being able to filter for multiple combinations of things like - what content is broken and the owner was not notified, and then how does that differ from broken content where the owner WAS notified?  
> Which now also makes me wonder -- if the "owner not notified" box is unchecked, does it include everything or only results that the owner would have received notice of?

#### Update verbiage and display of jobs that don't trigger a failure email

> this nit might go away based on the last comment, but i think the wording of "owner not notified" begs a lot of questions. What's a better way to describe this?

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -27,79 +31,97 @@ as_content_list <- function(content_df, client) {
...
+# email for each failing content item they own
+get_user_email <- function(client, guid) {
+  user_endpoint <- paste0("v1/users/", guid)
+  user <- client$GET(user_endpoint)
+  user$email
+}
+
+# filters content jobs down to failures and sets content_recovered depending on 
+# whether or not the latest job ended in error
+filter_to_failed_jobs <- function(jobs) {
```
dethmasque, line 45: > [Minor, but could we give this a more descriptive name? filter could be filtering on a bunch of different dimensions.](https://github.com/posit-dev/connect-extensions/pull/34/files/a33873f99d5fa8362692742fca20222ad938fbce#r2012148691)

Updated function name to be more explicit.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -27,79 +31,97 @@ as_content_list <- function(content_df, client) {
...
+  failed_jobs <- if (is.null(jobs) || nrow(jobs) == 0) {
```
dethmasque, line 46: Explicitly assigned `if`/`else` output to `failed_jobs` and returned `failed_jobs` following the conditional block instead of relying on automatically returned last computed value [as requested here](https://github.com/posit-dev/connect-extensions/pull/34/files/a33873f99d5fa8362692742fca20222ad938fbce#r2012156486).

extensions/prototype-content-with-issues-table/app.R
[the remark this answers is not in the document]
dethmasque: Updated to assign `if`/`else` output to `all_failed_jobs` and return `all_failed_jobs` following the conditional block instead of relying on automatically returned last computed value.

extensions/prototype-content-with-issues-table/app.R
jonkeane: Cool, thanks! The answer to this is no based on what I'm seeing but just to be explicit, you decided to defer doing something like this suggestion, correct?

> Alternatively, seeing this in a function: would it be possible to have the no-jobs path return a data.frame with zero rows (and make the rest of the code robust to that)? That might be a more natural approach and have fewer of these if/else branches. It also has the nice property that the functions all return one thing (a data.frame with a specific set of columns) rather than two things (either a data.frame or NULL, depending). This isn't required in this PR, but something to think about + might clean this code up and make it easier to follow.

extensions/prototype-content-with-issues-table/app.R
dethmasque: I replaced the `NULL` returns with an empty dataframe, and updated conditionals to only check that `nrow(df) == 0`.

dethmasque:
> I did ask for the (raw) html to be pulled out of the title, which it is, though now the app has no title at all. Would you mind adding one (both in the tab and on the page would be nice so that folks know what they are looking at and) so that this tab is findable when folks have a bunch open on connect.posit.it?

Shiny's UI `titlePanel()` included the header tags by default. With the lack of a favicon and HTML present when viewing the content in open solo it is hard to tell what the tab is with a crowded browser window, as you mentioned. A `title()` was added to the UI's `fluidPage()` instead of having the only title being the table header, which resolves the latter. [This existing Connect issue](https://github.com/posit-dev/connect/issues/20235) tracks populating the favicon in open solo, and the title no longer contains header tags on [the connect.posit deploy](https://connect.posit.it/content-issues-interactive/).

> And looking at the app again, have you decided to pass on this suggestion? It's ok to do that, but I wanted to make sure we explicitly acknowledge that in the PR.
> I know I've asked this before, but: I still wish that even when the rows are collapsed, they showed information from the most recent failure. I actually think most people would want to see that view and then only when something is interesting drill down and see the N failures for that app or dashboard. But currently if you collapse all of them, there is very little information at all.

I attempted to cover this last week using JS to find and select all of the expandable row buttons following table render, but while it would correctly count the number of expandable row elements, the element IDs were identical and while looping through each of the buttons it would only select and collapse the first group of rows. I did not spend additional time trying to get this to work after a bit of debugging and noticing that the IDs were the same across buttons.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -110,45 +132,92 @@ server <- function(input, output, session) {
...
+                                          "&body=Please%20investigate:%0A",
+                                          log_url,
+                                          "'>",
+                                          "✉",
+                                          "</a></span>")) %>%
+              mutate(log_url = paste0('<a href="',
+                                     log_url,
+                                     '" target="_blank">',
+                                     '<span style="font-size: 32px;">🗒',
+                                     '</a></span>')) %>%
```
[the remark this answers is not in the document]
dethmasque: Updated to use `URLencode()` with `reserved = TRUE` so it replaces single quotes.

jonkeane (approved):
Thanks for this

PR #35, opened 2025-03-11
Add workflows to release, version, and list Connect extensions

dotNomad:
This PR adds a new GitHub workflow and associated actions to release, version, and list Connect extensions.

#### High Level

To facilitate this I'll go over some of the high level changes:
- Extensions that we TAR now have a new `extension` section in the `manifest.json` that stores metadata that fills the extension list
  - This includes a `name`, `title`, `description`, `homepage` (the link to the GitHub page for the extension currently), and a [semver](https://semver.org/) `version`
- A new base level file called `extensions.json`
  - This currently just contains `{ extensions: [] }` but is where the gallery's Extension List will live
  - This file is not intended to be edited by hand except for removal (documentation on this to come)
- A new `scripts` folder that contains a `extension.list.ts` script that uses GitHub releases created by the new workflow to add to the Extension List
  - This is a TypeScript file so there are additional files to add TypeScript support and types for easier script writing

#### GitHub Workflow

There are 3 new GitHub actions in this PR:
- `lint-extension`: to check if a given extension name in the repo contains the new `extension` section in its `manifest.json` and that the `version` string is a valid semver
- `package-extension`: to TAR up the extension for releases
  - Takes an extension name and optionally an `artifact-name` for custom workflow extensions (see more below)
- `release-extension`: to release the packaged extension as a GitHub Release and update the Extension List with the data of the release for the Gallery to utilize

The main workflow - `extensions.yml` checks for changes to extensions to lint, package, and if on `main` release the extensions.

It does this in two groups:
- `simple-extensions`: includes all extensions in directories that can just be tarball-ed to release
- `complex-extensions`: all extensions that require custom workflows
  - The only one in this group is `publisher-command-center` at the moment
  
Once all of the extensions have been released as GitHub Releases the extension list script is run to capture that information for the Gallery to consume in something like the below:

<details>
  <summary>Example `extensions.json`</summary>
  
  ```json
  {
  "extensions": [
    {
      "name": "publisher-command-center",
      "title": "Publisher Command Center",
      "description": "A dashboard for publishers to help manage and track their content",
      "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/publisher-command-center",
      "latestVersion": {
        "version": "2.0.2",
        "released": "2025-03-11T22:55:27Z",
        "url": "https://github.com/dotNomad/connect-extensions/releases/download/publisher-command-center%40v2.0.2/publisher-command-center.tar.gz"
      },
      "versions": [
        {
          "version": "2.0.2",
          "released": "2025-03-11T22:55:27Z",
          "url": "https://github.com/dotNomad/connect-extensions/releases/download/publisher-command-center%40v2.0.2/publisher-command-center.tar.gz"
        },
        {
          "version": "2.0.1",
          "released": "2025-03-11T22:39:02Z",
          "url": "https://github.com/dotNomad/connect-extensions/releases/download/publisher-command-center%40v2.0.1/publisher-command-center.tar.gz"
        }
      ]
    },
    {
      "name": "reaper",
      "title": "Reaper",
      "description": "A view to a kill.",
      "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/reaper",
      "latestVersion": {
        "version": "2.0.4",
        "released": "2025-03-11T22:54:56Z",
        "url": "https://github.com/dotNomad/connect-extensions/releases/download/reaper%40v2.0.4/reaper.tar.gz"
      },
      "versions": [
        {
          "version": "2.0.4",
          "released": "2025-03-11T22:54:56Z",
          "url": "https://github.com/dotNomad/connect-extensions/releases/download/reaper%40v2.0.4/reaper.tar.gz"
        },
        {
          "version": "2.0.3",
          "released": "2025-03-11T22:42:29Z",
          "url": "https://github.com/dotNomad/connect-extensions/releases/download/reaper%40v2.0.3/reaper.tar.gz"
        },
        {
          "version": "2.0.1",
          "released": "2025-03-11T22:38:48Z",
          "url": "https://github.com/dotNomad/connect-extensions/releases/download/reaper%40v2.0.1/reaper.tar.gz"
        }
      ]
    }
  ]
}
  ```
</details> 

The actions and general `extensions.yml` workflow files are heavily commented to explain the steps and some of the decisions made.
  
##### Custom Workflow Extensions

The "Complex Extensions" have custom workflows. They cannot simply be packaged and require some additional steps to package, and other steps potentially in the future.

The one example now is `publisher-command-center` which needs to setup `node` to do a `build` of the frontend it ships. This cannot be handled generically so requires a more custom built GitHub workflow to be utilized by the main workflow - `extensions.yml`.

The custom workflows can use the same actions above (and in this PR do) for easy linting, packaging, and releasing.

#### Approach

There were a few goals I kept in mind when writing the workflows and script:
- Simple extensions should have little to no edits in GitHub Workflows to work
  - With this PR only 1 line is required to add a new extension
- Don't release every extension in every directory to avoid releasing an extension earlier than developers of it would like
- Complex extensions like `publisher-command-center` should be possible to build and release
- Actions in the workflow should be generalized to avoid duplicate code
- The flow should be easily extendable
- The flow should run in parallel wherever possible
- It should be easy to consume the latest version of an extension and versions in the Extension List should be sorted to help this as well
- Commit once to update the extension list
- Pull Requests should see if an extension would release, and with what TAR (included as artifact(s) of the workflow)

#### Workflow testing

Each step was done as a POC and then generalized to Actions and then other extensions. The development for this was done on a fork [`dotNomad/connect-extensions`](https://github.com/dotNomad/connect-extensions) to run the workflows.

There are a few workflows I'll point out that were done as tests:
- [Workflow #77](https://github.com/dotNomad/connect-extensions/actions/runs/13799577805/job/38599029749) updates two extensions, one with an error one without. The success still updates the extension list and the failure does not get released
- [Workflow #78](https://github.com/dotNomad/connect-extensions/actions/runs/13799601555) updates two extensions at once successfully
- [Workflow #80](https://github.com/dotNomad/connect-extensions/actions/runs/13799629758) shows no extensions updating and no extension commands run around linting, packaging, etc.
- [Workflow #81](https://github.com/dotNomad/connect-extensions/actions/runs/13799648074) shows an update to a single extension
- [Workflow #82](https://github.com/dotNomad/connect-extensions/actions/runs/13799692825/job/38599376279) shows a correct linting failure
- Workflows [#83](https://github.com/dotNomad/connect-extensions/actions/runs/13799794597/job/38599706942) and [#84](https://github.com/dotNomad/connect-extensions/actions/runs/13799796428/job/38599723775) show two workflows completing near-the-same-time and the extension list still updating.


.github/actions/lint-extension/action.yml
```diff
@@ -0,0 +1,38 @@
...
+    - run: |
+        jq '
+          if (.extension | type) != "object" then error("Missing extension object")
+          elif (.extension.name | type) != "string" then error("Missing extension.name")
+          elif (.extension.title | type) != "string" then error("Missing extension.title")
+          elif (.extension.description | type) != "string" then error("Missing extension.description")
+          elif (.extension.homepage | type) != "string" then error("Missing extension.homepage")
+          elif (.extension.version | type) != "string" then error("Missing extension.version")
+          else . end
+        ' ./extensions/${{ inputs.extension-name }}/manifest.json
```
tdstein, line 25: That's a lot of inline `jq`. Should this, or any of the other `jq` scripts, move into `package.json` so that they can be run locally?

.github/workflows/extensions.yml
```diff
@@ -0,0 +1,170 @@
...
+        working-directory: ./scripts
+
+      # Commits and pushes the updated extension list to the repository
+      # https://github.com/actions/checkout/tree/v4/?tab=readme-ov-file#push-a-commit-using-the-built-in-token
+      - run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add extensions.json
+          git commit -m "Update extension list"
+          git push
```
tdstein, line 170: Is this pushing directly to `main`? Should this go through a pull request to ensure a human is in the loop before changes make it to Connect?

extensions/audit-api/manifest.json
```diff
@@ -1434,5 +1434,12 @@
...
   },
-  "users": null
+  "users": null,
+  "extension": {
+    "name": "audit-api",
+    "title": "Audit Log API",
+    "description": "An API Publisher can call for audit logs.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/audit-api",
+    "version": "0.0.0"
+  }
```
tdstein, line 1444: Are there any drawbacks to using the `mainfest.json` for this purpose?

tdstein (commented):
Very neat! 

Does it make sense to write up the instructions in a `CONTRIBUTING.md` document? That's the first place I would think to look when making changes.

dotNomad:
> Does it make sense to write up the instructions in a `CONTRIBUTING.md` document? That's the first place I would think to look when making changes.

Yes definitely! We are picking up that work right away 👍

.github/actions/lint-extension/action.yml
dotNomad: Oh that is a good idea. I can look into that. It would be helpful to do a pre-push lint.

.github/workflows/extensions.yml
dotNomad: It is pushing directly to `main` yes. I could adjust it to instead open a PR.

My thought around this part of the flow is that a PR containing the updates to extensions has already been reviewed and merged and a release has already gone out in the repo, so the extension list should update right away.

I can see an argument though that any adjustments to `main` should go through a PR flow.

tdstein: That makes sense. I don't see anything wrong with moving forward with this approach. The risk seems minimal.

extensions/audit-api/manifest.json
dotNomad: One drawback that comes to mind is having to manually add this. That is true for the `connect-extension.toml` as well (which we plan to remove). But because this lives in the `manifest.json` we can expand tools that create manifests to support this in the future.

extensions/audit-reports/manifest.json
```diff
@@ -39,5 +39,12 @@
     "unpublished.qmd": {
       "checksum": "dd2d6de0ce972611a454e80776c4ed3d"
     }
+  },
+  "extension": {
+    "name": "audit-reports",
+    "title": "Audit Reports",
+    "description": "A quarto site of audit reports.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/audit-reports",
+    "version": "0.0.0"
```
edavidaja, line 48: Would it make sense for this to use a link that won't change if the extension is updated?

edavidaja:
Just a thought: what if `requires-build` or some more expansive notion of extension complexity was captured by the manifest, rather than as a workflow step?

extensions/audit-reports/manifest.json
dotNomad: I'm not sure I follow this question - can you expand a bit more?

edavidaja: As in, would it make sense to use a permalink:

`https://github.com/posit-dev/connect-extensions/tree/5e5066d9b0ad5f5d3b1454a2cc2437cf97bad5c9/extensions/audit-reports`

rather than the link to `tree/main`, which may refer to something different in the future if the extensions is updated.

dotNomad: Ah that is a great point. Currently the extension list doesn't capture a different homepage for each version instead having a homepage for the entire extension. This is similar in concept to how Homebrew has https://cli.github.com/ as the homepage for the `gh` formulae.

There could be a good amount of value in splitting that up per-version especially since the homepage is linking to a git repo (it doesn't have to).

dotNomad:
> Just a thought: what if `requires-build` or some more expansive notion of extension complexity was captured by the manifest, rather than as a workflow step?

Interesting 🤔 
Thinking through this it would allow for a possible generalization of the `complex-extension-changes` going through each extension that has a extension complexity flag and create a matrix of the directory names. However I believe each extension would still need its own custom workflow to setup the environment, run the write commands, and know how to package itself up.

Is there a big benefit you see in moving this over to the `manifest.json` that I may be missing?

edavidaja:
> Is there a big benefit you see in moving this over to the manifest.json that I may be missing?

I don't think I would describe the benefit as large, necessarily, but I think it offers some interesting flexibility in terms of exposing the information to Connect for use later even if we don't do anything with it now.

.github/workflows/extensions.yml
```diff
@@ -0,0 +1,170 @@
...
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          # Adding a new extension that has a directory that can be TARed?
+          # Add a new line here with the name of the extension and directory path
+          # Be sure the extension name and directory have the same name
+          # e.g. `extension-name: extensions/extension-name/**`
+          filters: |
+            reaper: extensions/reaper/**
+            integration-session-manager: extensions/integration-session-manager/**
```
dotNomad, line 32: After a conversation with @cgraham-rs and the recommendation from @edavidaja I _think_ we can get the simple extension flow to require no change from the extension authors. It will require some additional investigation, but the idea would be to find every `extension/**` directory with a signifier in the manifest that it is a simple extension, and see if it has changes.

.github/actions/release-extension/action.yml
```diff
@@ -0,0 +1,93 @@
...
+      shell: bash
+
+    - name: Get manifest extension version
+      run: |
+        MANIFEST_VERSION=$(semver -c $(jq -c -r '.extension.version' < ./extensions/${{ inputs.extension-name }}/manifest.json))
+        echo "MANIFEST_VERSION=$MANIFEST_VERSION" >> "$GITHUB_ENV"
+      shell: bash
+
+    # Grabs the latest version from the extensions.json file
+    # If an extension hasn't been released yet we default to `0.0.0` so any
```
cgraham-rs, line 26: Do we really want to allow unreleased extensions to be merged to `main`? It could potentially simply things if the act of merging to `main` implied releasing.

.github/actions/release-extension/action.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    # We only want to release if the manifest.json contains a newer semver
+    # version than the latest version in the extensions.json
+    # We compare that here, and echo if a release will occur
+    # This can be helpful when looking at Pull Request action outputs
+    # so it is clear what will happen on a merge to `main`
+    - name: Check if manifest has newer version
+      id: should_release
+      run: |
+        echo "The last released version is '$LATEST_VERSION' and the manifest version is '$MANIFEST_VERSION'"
```
cgraham-rs, line 44: Similar to the great code comments, all of this output seems excellent and should really help humans understand what is happening and why.

.github/workflows/extensions.yml
```diff
@@ -0,0 +1,170 @@
...
+          filters: |
+            publisher-command-center: extensions/publisher-command-center/**
+
+  # Creates and releases the Publisher Command Center extension using a custom
+  # workflow
+  publisher-command-center:
+    needs: [complex-extension-changes]
+    if: ${{ needs.complex-extension-changes.outputs.publisher-command-center == 'true' }}
+    uses: ./.github/workflows/publisher-command-center.yml
+
```
cgraham-rs, line 102: Thoughts on how we might be able to visually indicate "end of complex extension handling"? Since it currently requires additional jobs for each complex one it could make deciphering the flow challenging.

.github/workflows/extensions.yml
```diff
@@ -0,0 +1,170 @@
+name: Extension Workflow
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
```
cgraham-rs, line 8: We should consider adding concurrency settings to prevent the edge case of multiple simultaneous merges to main from colliding, assuming that would cause problems with the extensions list handling.

Similarly, we could use concurrency to prevent multiple simultaneous workflows on the same PR so it will only ever run against the latest commits. It simply saves runner minutes. We have examples of how to handle that in the Connect repo: https://github.com/posit-dev/connect/blob/ec1e47cb8b71715b9793a8e1a8e4c4f07610efe4/.github/workflows/ci.yml#L24

extensions/audit-api/manifest.json
cgraham-rs: One edge case we would want to document around is to remind any user that may re-regnerate a manifest using `rsconnect write-manifest` or other tools that they will lose their `extension` data and have to re-add it.

scripts/extension-list.ts
```diff
@@ -0,0 +1,162 @@
+import fs from "fs";
```
cgraham-rs, line 1: For my education - not having worked on Posit public repos before, are there any requirements or strongly suggested guidance around languages to use? I would expect in our general public facing world that Python would be a more familiar language.

scripts/extension-list.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+class ExtensionList {
+  constructor(public extensions: Extension[]) {
+    this.extensions = extensions;
+  }
+
+  static fromFile(path: string) {
+    const extensions = JSON.parse(fs.readFileSync(path, "utf8")).extensions;
+    return new ExtensionList(extensions);
```
cgraham-rs, line 40: Do we need error handling for potentially non-existent files, bad JSON, etc?

scripts/extension-list.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  public addRelease(manifest: ExtensionManifest, githubRelease) {
```
cgraham-rs, line 43: How rigorous do we need to be with typing? Each and every thing should have typing *always* or is that unrealistic?

scripts/extension-list.ts
```diff
@@ -0,0 +1,162 @@
...
+
+// Runs from ./scripts/dist so go up an additional level
+const extensionListFilePath = path.join(
+  __dirname,
+  "..",
+  "..",
+  "extensions.json"
+);
+
+const releases = JSON.parse(process.env.RELEASES);
```
cgraham-rs, line 153: Similar question about error handling if there's bad JSON.

scripts/extension-list.ts
```diff
@@ -0,0 +1,162 @@
...
+const list = ExtensionList.fromFile(extensionListFilePath);
```
cgraham-rs, line 154: Similar question about file/path error handling.

.github/actions/lint-extension/action.yml
```diff
@@ -0,0 +1,38 @@
...
+runs:
+  using: "composite"
+
+  steps:
+    # Ensures that the manifest.json for the given extension name
+    # contains all the required fields for the rest of the release workflow
+    - run: |
+        jq '
+          if (.extension | type) != "object" then error("Missing extension object")
+          elif (.extension.name | type) != "string" then error("Missing extension.name")
```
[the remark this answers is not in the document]
dotNomad: Definitely a good idea. I'll review the workflow to see if this 1-1 match is required before adding a hard linting rule.

.github/actions/release-extension/action.yml
dotNomad: I didn't want to:
- remove any extensions from the repo that we didn't want initially released
- require that extension authors keep long lived branches until the extension is ready to be released

This may not be the right choice though.

.github/workflows/extensions.yml
dotNomad: As far visually indicate do you mean in the GitHub actions summary?

.github/workflows/extensions.yml
dotNomad: > We should consider adding concurrency settings to prevent the edge case of multiple simultaneous merges to main from colliding, assuming that would cause problems with the extensions list handling.

The extension list updating pulls down the latest from `main` so the timing window to cause an issue is pretty tight. I actually tested this in a workflow linked above where two commits hit `main` near the same time.

We wouldn't want to cancel in progress runs on main to avoid a new PR merge from cancelling another extension from releases / updating the extension list.

We definitely can add the cancellation for when PRs update though (which it looks like what Connect is doing if I'm understanding correctly?)

scripts/extension-list.ts
dotNomad: > are there any requirements or strongly suggested guidance around languages to use

If there are I'm unaware of them as well. I went with TypeScript for two reasons:
- it is my most comfortable scripting language
- custom GitHub actions can be written in JavaScript (or TypeScript with additional tooling) so seemed the most appropriate for GitHub workflow scripting

References:
- https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-javascript-action?learn=create_actions&learnProduct=actions
- https://github.com/actions/javascript-action
- https://github.com/actions/typescript-action

scripts/extension-list.ts
dotNomad: We could, but not sure how much it would add. The most important thing is that the workflow job errors and doesn't continue.

Would potentially improve debugging when/if things go wrong.

scripts/extension-list.ts
dotNomad: I have `strict: false` set for the TypeScript config, and didn't pull in hard types for the GitHub release API JSON object. I can increase the strictness though. Could be a nice iterative step to avoid mistakes when/if we edit this script.

.github/workflows/extensions.yml
cgraham-rs: I was just envisioning a future where there were > 1 complex extension with its own job here and its going to make it harder to at-a-glance follow the logic of a typical simple flow.  Perhaps something as simple as a code comment? 

`# Here we resume common processing of all extensions`

Or something like that.

.github/workflows/extensions.yml
cgraham-rs: > We wouldn't want to cancel in progress runs on main to avoid a new PR merge from cancelling another extension from releases / updating the extension list.

Agreed. No cancellation on `main` ever. I believe if we configure concurrency for `main` but set `cancel-in-progress: false` then it enforces queueing. So nothing is cancelled, but nothing on `main` could run simultaneously.

Hmmmm, it seems that's not necessarily the case per the [concurrency documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency).

> This means that there can be at most one running and one pending job in a concurrency group at any time. When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any existing pending job or workflow in the same concurrency group, if it exists, will be canceled and the new queued job or workflow will take its place.

So if > 2 job hit main simultaneously there would definitely be cancellation. That's an unexpected behavior and not something we'd want.

It may be possible [Github Merge Queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#how-merge-queues-work) may enforce serial queueing on main. Needs more investigation.

I thought we attempted this on Connect for a while but the behavior "...A merge queue creates temporary branches with a special prefix to validate pull request changes..." caused problems with the way our release process works. I could not find any Slack conversations about this, but I did find this closed issue: https://github.com/posit-dev/connect/issues/29351

.github/workflows/extensions.yml
dotNomad: Ahh I see what you mean. I can easily add a comment. 👍

.github/workflows/extensions.yml
dotNomad: It does look like for concurrency we can specify not to cancel on certain branches: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-on-specific-branches

Could be helpful to eliminate the timing issue entirely.

dotNomad: I'm thinking I should do that as a follow-up though to test it in my fork prior to merging it in this repo.

.github/workflows/extensions.yml
dotNomad: Added in https://github.com/posit-dev/connect-extensions/pull/35/commits/c1bbb40af4a95dfdf8758feefb2540efe2878ca1

.github/actions/release-extension/action.yml
jonkeane: Yeah, I very much see where you're coming from @cgraham-rs , that theoretically keeping unrelease out of main makes things nice and clean, but PRs into other branches and having long running branches with development versions is advanced enough git that I would like to avoid requiring it if we can. 

It would be nice to have this noted prominently in a README (if we don't already) that just cause something is on `main` doesn't mean it's released

.github/actions/package-extension/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    description: The name of the artifact
+    required: false
+    type: string
+
+runs:
+  using: "composite"
+
+  steps:
+    # If we are not passed an artifact to use as the source for the tarball
+    # we will create a tarball from the extension's directory
```
jonkeane, line 19: Mostly for my own education: it's not immediately clear to me when we would take this path versus the other path down below for these. Could you educate me?

.github/actions/release-extension/action.yml
```diff
@@ -0,0 +1,93 @@
...
+        HIGHER_VERSION=$(semver "$MANIFEST_VERSION" "$LATEST_VERSION" | tail -n 1)
+          if [ "$MANIFEST_VERSION" = "$HIGHER_VERSION" ] && [ "$MANIFEST_VERSION" != "$LATEST_VERSION" ]; then
+          echo "Should release: the manifest version is higher than the lastest version"
+          echo "should_release=true" >> "$GITHUB_OUTPUT"
+        else
```
jonkeane, line 49: super minor, but it might be nice if these messages had slightly different shapes / cadences. I know I sometimes read to quickly and don't see a "not" that is critical :P Even like a 🚀 or 🛳️ at the front of the first and a 😴 or 🛏️ or something for the second would be helpful

.github/actions/release-extension/action.yml
```diff
@@ -0,0 +1,93 @@
...
+      run: |
+        gh release create $RELEASE_TAG \
+          --title "${{ inputs.extension-name }} v$MANIFEST_VERSION" \
+          ${{ inputs.extension-name }}.tar.gz
+      shell: bash
+
+    # We fetch the full release data to utilize in the extension list update
+    - name: Get release data
+      if: github.ref_name == 'main' && steps.should_release.outputs.should_release == 'true'
+      run: gh api /repos/${{ github.repository }}/releases/tags/$RELEASE_TAG > release-${{ inputs.extension-name }}.json
```
jonkeane, line 82: More for my education / curiosity: what's in the `release-....json`?

.github/actions/package-extension/action.yml
dotNomad: Absolutely. Currently the only time we do this is in the `publisher-command-center` custom worfklow which builds the extension, and then creates an artifact to package: https://github.com/posit-dev/connect-extensions/blob/c1bbb40af4a95dfdf8758feefb2540efe2878ca1/.github/workflows/publisher-command-center.yml#L33

dotNomad: The creation and passing of the artifact to this action allows custom workflows to select which files to package.

.github/actions/release-extension/action.yml
dotNomad: Great suggestion. I'll make that change so this is much easier to scan.

dotNomad: Changed the messaging and added emojis for glance-ability in https://github.com/posit-dev/connect-extensions/pull/35/commits/dab3a99372fe9fc12a26d96029cd9c7be0631a8f

They now read:
> 🚀 Will release! The manifest version is greater than the released version.
> 😴 Holding back from release: The manifest version is not greater than the released version.

.github/actions/release-extension/action.yml
dotNomad: > It would be nice to have this noted prominently in a README (if we don't already) that just cause something is on main doesn't mean it's released

Absolutely. The CONTRIBUTING doc we write for adding extensions will have this callout and tips for reading the GitHub Action logs to determine if a PR will result in a release or not 👍

.github/actions/package-extension/action.yml
jonkeane: Ah ok, so if it's a custom workflow style task, this allows us to build the artifact elsewhere (in the custom workflow) and then pass it to this workflow but/and therefore bypasses the building step in this workflow. Is that a good summary?

jonkeane: Or, even an accurate summary — doesn't need to be good 😂

.github/actions/release-extension/action.yml
jonkeane: Ah interesting. It might be worth stating (more?) explicitly in the comment "We fetch the full release data from the github API" That _wasn't_ at the top of my head for where we were fetching that from, actually.

.github/actions/package-extension/action.yml
dotNomad: > Ah ok, so if it's a custom workflow style task, this allows us to build the artifact elsewhere (in the custom workflow) and then pass it to this workflow but/and therefore bypasses the building step in this workflow. Is that a good summary?

Very close. The only correction I'll make is that this is a composite action. It allows us to build the artifact we want packaged and pass it to this action. It avoids building as part of the packaging step (for those extensions with custom workflows), and simplifies the things that need to be passed to this action.

.github/actions/release-extension/action.yml
dotNomad: Updated the comment in https://github.com/posit-dev/connect-extensions/pull/35/commits/e5340888b0e86d4cb5616466e63b45d21d9181ae to mention that the release data is fetched from the GitHub API 👍

.github/actions/package-extension/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Create tar
+      if: ${{ inputs.artifact-name == '' }}
+      run: tar -czf ${{ inputs.extension-name}}.tar.gz ./extensions/${{ inputs.extension-name}}
+      shell: bash
+
+    # If we are passed an artifact to use as the source for the tarball
+    # we will download the artifact and create a tarball from it
+    # this avoids including extra files in the extension
+    - name: Download optional artifact
+      if: ${{ inputs.artifact-name != '' }}
```
fizzboop, line 29: Do we offer a message to the user if an error occurs during this process? Or is that even possible?

.github/actions/lint-extension/action.yml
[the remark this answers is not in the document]
dotNomad: Going to do this as a follow-up for improved linting.

.github/workflows/extensions.yml
dotNomad: A quick look into this I cannot find a way to do both matrix directories and see which have changed without some inputs. There is probably a way to do this, but it is more complex than I expected. We can tackle this if the feedback from the CONTRIBUTING documentation is that it is too complex.

.github/actions/package-extension/action.yml
dotNomad: There will be a failure if the artifact doesn't exist and the logs will give the name and that no artifact matched it. Thats all built in to the download-artifact action. Does that satisfy that you think?

fizzboop: Yep, I think that makes sense.

extensions/reaper/manifest.json
```diff
@@ -23,5 +23,12 @@
     "connect-extension.toml": {
       "checksum": "c6e6af366dacce9c86f0a06901b931f3"
     }
+  },
+  "extension": {
+    "name": "reaper",
+    "title": "Reaper",
+    "description": "A view to a kill.",
```
cgraham-rs, line 30: Do we want a better description here?

.github/workflows/publisher-command-center.yml
```diff
@@ -1,26 +1,30 @@
...
+
       - uses: actions/setup-node@v4
         with:
-          node-version: 'lts/*'
-          cache: 'npm'
+          node-version: "lts/*"
+          cache: "npm"
           cache-dependency-path: extensions/${{ env.EXTENSION_NAME }}/package-lock.json
 
       - run: npm ci
```
cgraham-rs, line 30: This isn't important and should not prevent merge, but since this is the only complex extension in the gallery and is likely to be used as an example/template to follow it could be useful to have some of your amazing code comments in there that fully explain what's happening each step of the way and why things are in here.

cgraham-rs (approved):
Looking forward to seeing all of this work merged.  Again, great job on the code comments.

.github/actions/package-extension/action.yml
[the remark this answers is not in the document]
dotNomad: I kept the explanations in the actions a bit lighter since "simple content" probably doesn't need to interact with them. We can definitely add some deeper explanations if we get questions often.

extensions/reaper/manifest.json
dotNomad: Added this to #30371

.github/workflows/publisher-command-center.yml
dotNomad: Good point. I'll add comments before I merge this.

PR #36, opened 2025-03-12
most used content (plot edition)

toph-allen:
This iteration adds plots to the "most used content" prototype.

I added two plots:
- Total daily content views for the server.
- Total daily unique visitors for the server.

Per-content-item plot are beyond the scope of this prototype, though I implemented them in an earlier one.

Also added a `formatStyle` directive for the table to truncate very long content names.

As usual, running on:
- Dogfood: https://rsc.radixu.com/most-used-content/
- connect.posit.it: https://connect.posit.it/most-used-content/

I notice the way I'm using date filter params for the new firehose endpoint aren't working, so on Dogfood we're seeing all historical data. I'm going to investigate that shortly, but that is unrelated to the changes here.

### DIFF
--- a/extensions/most-used-content/app.R
+++ b/extensions/most-used-content/app.R
@@ -5,6 +5,9 @@ library(connectapi)
 library(dplyr)
 library(purrr)
 library(lubridate)
+library(ggplot2)
+library(plotly)
+library(tidyr)
 
 shinyOptions(
   cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
@@ -24,6 +27,18 @@ ui <- page_fillable(
 
         actionButton("clear_cache", "Clear Cache", icon = icon("refresh"))
       ),
+
+      tabsetPanel(
+        tabPanel(
+          "Plot of Daily Views",
+          plotlyOutput("daily_views_plot")
+        ),
+        tabPanel(
+          "Plot of Unique Daily Visitors",
+          plotlyOutput("unique_daily_visitor_plot")
+        )
+      ),
+
       card(
         DTOutput(
           "content_usage_table"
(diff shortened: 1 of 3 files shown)


jonkeane:
> Per-content-item plot are beyond the scope of this prototype, though I implemented them in an earlier one.

Looking at [the issue](https://github.com/posit-dev/connect/issues/30320), it wasn't written as clearly as it probably should have been and should have probably included per-item plot explicitly. Regardless, the per-item plot definitely should be in scope and is definitely the right plot for this. We want this to let someone know which items are most popular (+ how popular they are) and not _just_ how popular is the server overall. Would you mind swapping that plot to be per-item? 

In case it's helpful (these aren't _requirements_, but want to be (more) explicit): I'm imagining something like along the y axis is content item titles, and then bars or dots that show the number of people who have visit this item for all time.

On a smaller note: I might have just been unlucky with the cache but when I went to both of these this morning they took a _while_ to load.

toph-allen:
@jonkeane Per-content-item plots — I'm gonna spitball a bit to get my thoughts out on how best to display that information.

> Looking at [the issue](https://github.com/posit-dev/connect/issues/30320), it wasn't written as clearly as it probably should have been and should have probably included per-item plot explicitly. Regardless, the per-item plot definitely should be in scope and is definitely the right plot for this. We want this to let someone know which items are most popular (+ how popular they are) and not _just_ how popular is the server overall. Would you mind swapping that plot to be per-item?
> 
> In case it's helpful (these aren't _requirements_, but want to be (more) explicit): I'm imagining something like along the y axis is content item titles, and then bars or dots that show the number of people who have visit this item for all time.

Thanks for clarification! I think when I initially thought about "per-content-item-plots" I was thinking about separate timelines for each content item (like, when you click on them, showing it in a popover, or something) but actually a leaderboard style plot like we have in "who deploys most often" is the obvious choice lol. 🤦🏻

I think it'd be nice to have stacked bars for logged-in and anonymous viewers, with the logged-in viewers being the base of the bar and the anonymous views the portion on top. I think that'd be really clear in its meaning and communicative of the fullness of the data.

And possibly could do a timeline showing… perhaps the top 25 content items' popularity over time, as a frequency polygon.

> On a smaller note: I might have just been unlucky with the cache but when I went to both of these this morning they took a _while_ to load.

And yeah if you went to them in the morning before anyone else had visited them, you were probably unlucky with the cache. There might also be a bug or an API problem causing the longer load time:

A few thoughts:

- If you were looking on connect.posit.it, it was using the old endpoint.
- If you were looking on Dogfood, it uses the firehose endpoint. There's currently a bug in this content where we don't use the date filters to limit the data for the week, so it's loading ALL of dogfood's usage history. I think the majority of the slowness here is from R parsing that large data frame. So hopefully your problem will be fixed soon!

---

If not a bug, and if we're thinking about putting these into production, it might be worthwhile thinking about ways to fix that problem. Thinking of your example of "my CEO checks the dashboard every morning", you don't want _them_ to incur the long load time!

Options: 
 - My first thought is: show old data from the cache while new data is being loaded asynchronously (possibly with a message saying that new data is loading). I don't know if that's easy to do out of the box though.
- With the audit logs data, I had a second piece of content cache the data on a schedule

marcosnav (approved):

PR #37, opened 2025-03-13
who deploys most often (plot edition)

toph-allen:
Add plots to "Who Deploys Most Often".

- Dogfood: https://rsc.radixu.com/who-deploys-most-often/
- connect.posit.it: https://connect.posit.it/who-deploys-most-often/

I added four plots, visible in panels above the table.

There are three bar plots:

- Plot by Total Active Content
- Plot by Number of New Content
- Plot by Number of Deploys

These all display bar plots of total deploy activity by username; the top 25 users (due to space constraints), for readability.

- Timeline of Deploy Activity

Deployment activity over time, a line plot. These are more useful when zoomed in.

Points to note:

- Table-based filtering does not apply to the plot. It may make sense to remove data filters from the table and put them in the sidebar instead. This makes sense to do as part of the [date range](https://github.com/posit-dev/connect/issues/30315) story to me, since that is already going to be adding filter controls to the sidebar, but I could also add it as part of this one. (The same is true for the other "plot edition" story.)

### DIFF
--- a/extensions/who-deploys-most-often/app.R
+++ b/extensions/who-deploys-most-often/app.R
@@ -7,6 +7,8 @@ library(purrr)
 library(connectapi)
 library(pins)
 library(tidyr)
+library(ggplot2)
+library(plotly)
 
 shinyOptions(
   cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 24)
@@ -24,6 +26,24 @@ ui <- page_fillable(
         title = "Filter Data",
         open = FALSE,
         actionButton("clear_cache", "Clear Cache", icon = icon("refresh"))
+      ),
+      tabsetPanel(
+        tabPanel(
+          "Plot by Total Active Content",
+          plotlyOutput("total_active_content_plot")
+        ),
+        tabPanel(
+          "Plot by Number of New Content",
+          plotlyOutput("new_content_plot")
+        ),
+        tabPanel(
+          "Plot by Number of Deploys",
+          plotlyOutput("n_deploys_plot")
+        ),
+        tabPanel(
+          "Timeline of Deploy Activity",
+          plotlyOutput("deploys_timeline")
+        )
       ),
       card_body(textOutput("data_note"), fill = FALSE),
       card(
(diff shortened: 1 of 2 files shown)


extensions/who-deploys-most-often/app.R
```diff
@@ -130,6 +170,49 @@ server <- function(input, output, session) {
...
       formatDate(columns = "Time of Latest Deploy", method = "toLocaleString")
   })
+
+  output$total_active_content_plot <- renderPlotly(
+    ggplotly(
+      summarized_data() |>
+        arrange(desc(active_content)) |>
+        slice_head(n = 25) |>
+        ggplot() +
+        geom_col(aes(x = reorder(full_name, active_content), y = active_content)) +
```
dethmasque, line 180: Super nit, but tooltips would be more readable if you use something like `text = paste("User: ", full_name, " Active Content: ", active_content)` across plots where `reorder()` is used. Currently the tooltip displays:
```
reorder(full_name, active_content): <full_name_value>
active_content: <active_content_value>
```

jonkeane: :100: agree with @dethmasque on this

jonkeane:
From poking around at this: I noticed a perceptible delay when switching to the timeline, do you know what's up with that? It might not be worth fixing, but it is noticeable (the first time I navigate to there, though it's happened twice today maybe I'm just unlucky on the cache?)

The other feedback I have on the plots is for the timeline view: could we remove folks who are at 0 for the whole period? The long list of names is a little rough and if I'm seeing this correctly, most are at 0 in the time period.

toph-allen (commented):
Reviewing my own PR:

- The table-based filters have stopped working for some reason. Also, I think that with the presence of the plots, people will expect the filters to apply to the plots _and_ the table. **I think this means I should remove the filters from the table and move them to the sidebar, and add filters to the data processing steps.** This might require some changes to caching down the road, as the deploy timeline seems to take a non-insignificant time to render right now, but I'm not gonna worry about that now.
-

toph-allen:
> From poking around at this: I noticed a perceptible delay when switching to the timeline, do you know what's up with that? It might not be worth fixing, but it is noticeable (the first time I navigate to there, though it's happened twice today maybe I'm just unlucky on the cache?)

Yep, I noticed the delay (see my previous comment, which I wrote before I saw your comment). I think it _might_ be to do with the fact that I'm not caching the computation that creates the data for the plot, but perhaps it's more that Shiny is doing the actual data transmission lazily, when the tab is clicked, rather than ahead of time. I'm going to investigate that.

In my previous comment, I said that adding filters to the sidebar might call for a change in caching behavior, and that thought was related to this performance issue. If the plot takes a second to build, and it has to be rebuilt every time the filters change, _and_ the building happens lazily, then each time the filters changed, that tab would take a second to display.

So — I'll investigate performance.

> The other feedback I have on the plots is for the timeline view: could we remove folks who are at 0 for the whole period? The long list of names is a little rough and if I'm seeing this correctly, most are at 0 in the time period.

Good point! I think that also having filters that affect both the plot and the table will help with this — essentially the "Number of Deploys" filter would have its lower bound set to `1` by default.

toph-allen:
Closing because this is no longer planned.

PR #38, opened 2025-03-13
who is visiting this content (table view)

toph-allen:
The actual issue describing the "table view" goes into more detail than just a table view — I'm guessing that's actually referring to later issues in the series (https://github.com/posit-dev/connect/issues/30286).

Dogfood: https://rsc.radixu.com/who-is-visiting-this-content/
connect.posit.it: https://connect.posit.it/who-is-visiting-this-content/

Most basic possible approach: paste in a content GUID and get a table of individual visits and a table of total activity, sorted by descending date and number of visits respectively. Shows a summary sentence of total visits.

Screen Shot 2025-03-13 at 06 15 55 PM@2x

Notes:

- Uses the firehose on dev builds and the existing APIs if it isn't available, so the intersection of the available data is displayed (i.e. just username and timestamp info).
- Data are for the last week.
- Shows a message asking you to paste in a GUID if you haven't entered anything.
- The summary sentence is going to look bad for content that doesn't have a title (which might mainly be a problem on Dogfood)


jonkeane:
> The actual issue describing the "table view" goes into more detail than just a table view — I'm guessing that's actually referring to later issues in the series (https://github.com/posit-dev/connect/issues/30286).

💯 right about this, drafting error when we (I, I think in this case) was breaking things down.

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,151 @@
...
+
+  observe({
+    if (nchar(input$content_guid) == 0) {
+      show("guid_input_msg")
+      hide("content_visit_tables")
+    } else {
+      hide("guid_input_msg")
+      show("content_visit_tables")
+    }
+  })
```
jonkeane, line 74: I _think_ this would go here, but what do you think about adding in a parser that lets someone paste in a full URL that includes the GUID? This is a bit of an extension / scope creep so if we want to make a separate issue for it and do that in a follow on that's totally fine. But the UX of needing to select _just_ the GUID form the URL (or get to it in the settings panel) could be better. We had good success with the publisher where the field that accepts GUID also accepts the URL (and, in fact, in the publisher it's the GUID alone that is the hidden easter egg that it _just works_, we though that folks would be more confident / able to / comfortable with grabbing the URL)

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,151 @@
...
+
+  # Default dates. "This week" is best "common sense" best represented by six
+  # days ago thru the end of today. Without these, content takes too long to
+  # display on some servers.
+  date_range <- reactive({
+    list(
+      from_date = today() - ddays(6),
+      to_date = today()
+    )
+  })
```
jonkeane, line 87: This is a great default + kick the can of "which is the best period? Should I do forever? etc" for when we get to the filtering edition later 💯

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,151 @@
...
+
+  content <- reactive({
+    # Somewhat hacky to get all content and filter for the guid, but makes error
+    # handling somewhat easier when printing content title.
+    get_content(client)
+  }) |> bindCache("static_key")
```
jonkeane, line 93: I might be confused on this myself, but the comment mentions "all content ... for the guid" but this looks like it's getting all content items period (without any reference to the guid? Or am I misunderstanding?

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,151 @@
...
+  aggregated_visits_data <- reactive({
+    usage_data() |>
+      filter(content_guid == input$content_guid) |>
+      group_by(user_guid) |>
+      summarize(n_visits = n()) |>
+      left_join(user_names(), by = "user_guid") |>
+      replace_na(list(full_name = "[Anonymous]")) |>
+      arrange(desc(n_visits)) |>
+      select(n_visits, full_name, username)
+  }) |> bindCache(date_range()$from_date, date_range()$to_date, input$content_guid)
```
jonkeane, line 128: This might very well be another issue/PR later, since it wasn't in the original and is conceptually a bit of an addition, but:

If I'm remembering / understanding correctly items that have multiple URL routes and that folks navigate through will have a bunch of lines in the firehosue endpoint that are really one session of visits (to use the correct word to describe it, but we have overloaded that slightly with our old API endpoint! I'm meaning the abstract / general use not specifically what comes from our shiny sessions endpoint). What do you think about adding in a collapse here too? Something like "entries within N minutes of each other are considered one visit"? We definitely wouldn't want only one measure or the other — both are valuable.

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,151 @@
...
+      "Content '{content_title}' had {nrow(hits)} between ",
+      "{min(hits$timestamp)} and {max(hits$timestamp)}."
+    )
+  })
+
+
+  output$summary_message <- renderText(summary_message())
+  output$all_visits <- renderTable(
+    all_visits_data()
+  )
```
jonkeane, line 145: How hard would it be to add nice formatting to the timestamp column here?

extensions/who-is-visiting-this-content/app.R
toph-allen: I was trying to keep this as minimal as possible, figuring it wouldn't be the final UI for selecting a piece of content.

Perhaps it would be fine to add a bit of extra logic to extract the GUID if a URL is pasted in. OTOH, perhaps even having this hide / show logic and a "paste in a content guid" message is too much code to pour into something that won't be a final UI. I can really see both arguments. I'd rather not put more effort in here till a later edition, at the very least.

extensions/who-is-visiting-this-content/app.R
toph-allen: Yes, this is definitely a priority!

Following our discussion What do you think about separating this investigation out as its own issue? Kind of a spike on computing the right metric? It'll involve some research and experimentation, and possibly discussion with @marcosnav.

extensions/who-is-visiting-this-content/app.R
toph-allen: Very easy; its lack of presence is just a factor of "try to do a quick first draft". However, since we'll be switching the implementation to `gt` rather than `DT`, perhaps I won't implement it in `DT`?

extensions/who-is-visiting-this-content/app.R
toph-allen: Sorry, yeah, this comment is not super clear.

We grab the entire content data frame here and cache it, and then filter for the pasted-in GUID, to obtain things such as the content title.

I called it "hacky" because within the frame of single-content-item dashboard, I feel like a more "proper" way would be to request that data whenever a GUID is provided at the `v1/content/{GUID}` endpoint.

I did it this way here… mainly to avoid having to try/catch the request in the code on line 130 that generates a summary sentence. Filtering this data frame for a GUID that isn't present would just return a blank string for the title, whereas making a `connectapi::content_item()` call with a GUID that isn't present returns an error message, and I wanted to keep things as lightweight as possible for something (the summary sentence) that I didn't even know if we'd want to keep in the dashboard.

extensions/who-is-visiting-this-content/get_usage.R
```diff
@@ -0,0 +1,60 @@
...
+    connectapi:::unversioned_url("instrumentation", "content", "hits"),
+    query = list(
+      from = from,
+      to = to
+    )
+  )
+
+  # FIXME for connectapi: This is slow, it's where most of the slowness is with
+  # the new endpoint.
+  usage_parsed <- connectapi:::parse_connectapi_typed(usage_raw, usage_dtype)
```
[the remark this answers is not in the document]
toph-allen: There is now an issue to track it: https://github.com/posit-dev/connectapi/issues/383 :)

It's not _horrendously_ slow, but it is definitely at least worth investigating as a `connectapi` issue.

And as for typing enforcement — I guess that depends on what you mean by "need". It's _nice_: `parse_connectapi_typed()` "automatically" handles deserializing the JSON from Connect into the correct R types and structure, abstracting away a lot of stuff that's fiddly in R, and there's value in that.

Another way I'm thinking about it is… if I just coded from scratch all the things `parse_connectapi_typed` does to shape the data into the shape I want, maybe be just as slow. And if that were the case, then I'd want to keep it, because it makes life a lot simpler in a few ways.

However, if, say, the base `parse_connectapi()` function, which doesn't do any type enforcement, was much faster, than it'd probably be better to use that and handle type conversion in a different way.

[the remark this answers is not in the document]
jonkeane: I can't quite tell from this part of your comment;

> However, if, say, the base `parse_connectapi()` function, which doesn't do any type enforcement, was much faster, than it'd probably be better to use that and handle type conversion in a different way.

Is `parse_conenctapi()` much faster in this case?

extensions/who-is-visiting-this-content/app.R
jonkeane: > We grab the entire content data frame here and cache it, and then filter for the pasted-in GUID, to obtain things such as the content title.

👍 would you mind updating the comment (if you haven't already) to reflect that?

extensions/who-is-visiting-this-content/app.R
jonkeane: Sure, that's fine

extensions/who-is-visiting-this-content/get_usage.R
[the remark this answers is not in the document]
toph-allen: I wasn't sure when I wrote the comment (I wanted to avoid opening cans of worms while I was responding to your comments).

It isn't much faster:

```r
> length(usage_raw)
[1] 485794
> system.time(connectapi:::parse_connectapi_typed(usage_raw, usage_dtype))
   user  system elapsed 
 28.245   1.488  29.798 
> system.time(connectapi:::parse_connectapi(usage_raw))
   user  system elapsed 
 17.602   0.330  18.128
```

I'm also going to post this info to the `connectapi` issue so that it doesn't get lost.

extensions/who-is-visiting-this-content/app.R
toph-allen: Done!

extensions/who-is-visiting-this-content/app.R
toph-allen: I have a note to try switching to `gt` in https://github.com/posit-dev/connect/issues/30322

extensions/who-is-visiting-this-content/app.R
toph-allen: https://github.com/posit-dev/connect/issues/30515

toph-allen:
> So far looks good to me, some minor feedback from running it.
> 
> 
> 
> - Table headers, `timestamp` -> `Time`, `full_name` -> `Full Name`, `n_visits` -> `Total Visits`
> 
> - Values on the timestamp column to be human friendly

Thanks for taking a look!

As part of the feedback from the review session, we're gonna investigate changing the table display framework we're using. That will change how we perform date formatting and header renaming, so in [a discussion with @jonkeane](https://github.com/posit-dev/connect-extensions/pull/38#discussion_r2000999236) I decided to skip adding those for this version.

toph-allen:
As I got two comments about it, I added column names and date formatting to this version, rather than waiting till a later iteration.

extensions/who-is-visiting-this-content/app.R
toph-allen: I added formatting. :)

extensions/who-is-visiting-this-content/app.R
```diff
@@ -0,0 +1,153 @@
...
+library(shinyjs)
+library(connectapi)
+library(dplyr)
+library(glue)
+library(lubridate)
+library(tidyr)
+
+shinyOptions(
+  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
+)
```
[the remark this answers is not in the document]
toph-allen: It takes 30 seconds to process the firehose endpoint in `connectapi`. There is an issue I opened to see if that performance cost can be reduced, but for now, caching after that processing step makes the app responsive faster after loading.

marcosnav (approved):
Thank you for addressing previous comments

PR #39, opened 2025-03-13
content with issues prototype: table view (follow-up)

dethmasque:
Additional feedback was received in #33 following merge. This PR addresses that feedback prior to continuing with #34.

[Deployed to Dogfood](https://rsc.radixu.com/connect/#/apps/1cf75066-007f-4f54-af28-f21ea6e5e262) 

<details>
<summary> Loading from cache GIF </summary>

content with issues dogfood load from cache

</details>

[Deployed to connect.posit.it](https://connect.posit.it/connect/#/apps/bc3dc19c-6c49-4d8b-9db0-c2e629a050b9/) 

<details>
<summary> Loading from cache GIF </summary>

content with issues connect posit it load from cache

</details>

Cache expires every 8h and takes ~7-8m to refresh.

This R Shiny prototype extension requests and caches usage data from the last 30 days along with all available content, filtered to items deployed within the last 60 days. Jobs are requested for each content item, filtered down to failed jobs, and then compiled with last visit and content information into a static table.

### DIFF
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -31,17 +31,9 @@ as_content_list <- function(content_df, client) {
 # usage data if it does, then compiles content, job, and usage data together 
 # into a tibble, returning it.
 get_failed_job_data <- function(item, usage) {
-  failed_jobs <- tryCatch(
-    {
-      get_jobs(item) |>
+  failed_jobs <- get_jobs(item) |>
         # filter out successful and running jobs
         filter(exit_code != 0 & status != 0 & !(is.na(end_time)))
-    },
-    error = function(e) {
-      # content item's content URL 404s (incomplete deployment)
-      NULL
-    }
-  )
   
   if (is.null(failed_jobs) || nrow(failed_jobs) == 0) {
     # content item does not have failed jobs
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -8,7 +8,7 @@ library(tidyr)
 
 # cache data to disk with a refresh every 8h
 shinyOptions(
-  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
+  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 12)
 )
 
 source("get_usage.R")
@@ -49,18 +49,18 @@ get_failed_job_data <- function(item, usage) {
       slice_max(timestamp) %>%
       select(timestamp)
     # return required information from https://github.com/posit-dev/connect/issues/30288
-    all_failed_jobs <- map_dfr(seq_len(nrow(failed_jobs)), function(i) {
+    all_failed_jobs <- map_dfr(seq_len(nrow(failed_jobs)), ~
       tibble(
         "content_title" = item$content$title,
         "content_guid" = item$content$guid,
         "content_owner" = item$content$owner[[1]]$username,
-        "job_failed_at" = failed_jobs$end_time[i],
-        "failed_job_type" = failed_jobs$tag[i],
-        "failure_reason" = failed_jobs$exit_code[i],
+        "job_failed_at" = failed_jobs$end_time[.x],
+        "failed_job_type" = failed_jobs$tag[.x],
+        "failure_reason" = failed_jobs$exit_code[.x],
         "last_deployed_time" = item$content$last_deployed_time,
         "last_visited" = as.POSIXct(last_visit$timestamp)
       )
-    })
+    )
     all_failed_jobs
   }
 }
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -4,6 +4,7 @@ library(DT)
 library(connectapi)
 library(dplyr)
 library(purrr)
+library(tidyr)
 
 # cache data to disk with a refresh every 8h
 shinyOptions(
@@ -31,34 +32,7 @@ get_failed_job_data <- function(item, usage) {
     {
       get_jobs(item) |> 
         # filter successful jobs
-        filter(exit_code != 0) |> 
-        # map content job types to something more readable 
-        mutate(tag = case_when(
-               tag %in% c("build_report", "build_site", "build_jupyter") ~ "Building",
-               tag %in% c("packrat_restore", "python_restore") ~ "Restoring environment",
-               tag == "configure_report" ~ "Configuring report",
-               tag %in% c("run_app", 
-                          "run_api", 
-                          "run_tensorflow", 
-                          "run_python_api",
-                          "run_dash_app",
-                          "run_gradio_app",
-                          "run_streamlit",
-                          "run_bokeh_app",
-                          "run_fastapi_app",
-                          "run_voila_app",
-                          "run_pyshiny_app") ~ "Running",
-               tag == "render_shiny" ~ "Rendering",
-               tag == "ctrl_extraction" ~ "Extracting parameters",
-               TRUE ~ tag)) |>
-        # map exit codes to something more readable 
-        mutate(exit_code = as.character(exit_code)) |>
-        mutate(exit_code = case_when(
-               exit_code %in% c("1", "2", "134") ~ "failed to run / error during running",
-               exit_code == "137" ~ "out of memory",
-               exit_code %in% c("255", "15", "130") ~ "process terminated by server",
-               exit_code %in% c("13", "127") ~ "configuration / permissions error",
-               TRUE ~ exit_code))
+        filter(exit_code != 0) 
     },
     error = function(e) {
       # content item does not have any jobs 
@@ -74,11 +48,8 @@ get_failed_job_data <- function(item, usage) {
       filter(content_guid == item$content$guid) %>%
       slice_max(timestamp) %>%
       select(timestamp)
-    if (is.na(item$content$title)) {
-      item$content$title <- "" # use empty strings when content is missing title
-    }
-    # return required information from https://github.com/posit-dev/connect/issues/30288 
-    all_failed_jobs <- bind_rows(lapply(seq_len(nrow(failed_jobs)), function(i) {
+    # return required information from https://github.com/posit-dev/connect/issues/30288
+    all_failed_jobs <- map_dfr(seq_len(nrow(failed_jobs)), function(i) {
       tibble(
         "content_title" = item$content$title,
         "content_guid" = item$content$guid,
@@ -89,7 +60,7 @@ get_failed_job_data <- function(item, usage) {
         "last_deployed_time" = item$content$last_deployed_time,
         "last_visited" = as.POSIXct(last_visit$timestamp)
       )
-    }))
+    })
     all_failed_jobs
   }
 }
@@ -119,7 +90,34 @@ server <- function(input, output, session) {
   
   # output the datatable of failed jobs
   output$jobs <- renderDT({
-    datatable(bad_content_df(), 
+    datatable(bad_content_df() |>
+                # map job type to something more readable
+                mutate(failed_job_type = case_when(
+                  failed_job_type %in% c("build_report", "build_site", "build_jupyter") ~ "Building",
+                  failed_job_type %in% c("packrat_restore", "python_restore") ~ "Restoring environment",
+                  failed_job_type == "configure_report" ~ "Configuring report",
+                  failed_job_type %in% c("run_app", 
+                             "run_api", 
+                             "run_tensorflow", 
+                             "run_python_api",
+                             "run_dash_app",
+                             "run_gradio_app",
+                             "run_streamlit",
+                             "run_bokeh_app",
+                             "run_fastapi_app",
+                             "run_voila_app",
+                             "run_pyshiny_app") ~ "Running",
+                  failed_job_type == "render_shiny" ~ "Rendering",
+                  failed_job_type == "ctrl_extraction" ~ "Extracting parameters",
+                  TRUE ~ failed_job_type)) |>
+                # map exit codes to something more readable 
+                mutate(failure_reason = case_when(
+                  failure_reason %in% c(1, 2, 134) ~ "failed to run / error during running",
+                  failure_reason == 137 ~ "out of memory",
+                  failure_reason %in% c(255, 15, 130) ~ "process terminated by server",
+                  failure_reason %in% c(13, 127) ~ "configuration / permissions error",
+                  TRUE ~ as.character(failure_reason))) |>
+                mutate(content_title = replace_na(content_title, "")),
               rownames = FALSE, 
               escape = FALSE,
               options = list( # non-interactive table for this prototype
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -77,7 +77,7 @@ server <- function(input, output, session) {
     as_content_list(content, client)
   }) |> bindCache("static_key")
   
-  # cache last 30d of usage (Jobs.MaxCompleted is 30d), takes ~4m to build
+  # cache last 30d of usage (Jobs.MaxCompleted is 30d), takes ~5m to build
   usage <- reactive({
     from = (Sys.time() - days(30))
     to = Sys.time()
--- a/extensions/prototype-content-with-issues-table/app.R
+++ b/extensions/prototype-content-with-issues-table/app.R
@@ -4,11 +4,13 @@ library(DT)
 library(connectapi)
 library(dplyr)
 library(purrr)
+library(lubridate)
 library(tidyr)
 
-# cache data to disk with a refresh every 8h
+# cache data to disk with a refresh every 8h, table renders in ~7m when cache
+# is expired, deleted, or on initial deploy
 shinyOptions(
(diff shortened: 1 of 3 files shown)


extensions/prototype-content-with-issues-table/app.R
```diff
@@ -31,34 +32,7 @@ get_failed_job_data <- function(item, usage) {
...
-               tag == "ctrl_extraction" ~ "Extracting parameters",
-               TRUE ~ tag)) |>
-        # map exit codes to something more readable 
-        mutate(exit_code = as.character(exit_code)) |>
-        mutate(exit_code = case_when(
-               exit_code %in% c("1", "2", "134") ~ "failed to run / error during running",
-               exit_code == "137" ~ "out of memory",
-               exit_code %in% c("255", "15", "130") ~ "process terminated by server",
-               exit_code %in% c("13", "127") ~ "configuration / permissions error",
-               TRUE ~ exit_code))
```
dethmasque, line 61: The above was moved to the table render code in response to:

> [What do you think about moving this munging into the display code rather than the querying code?](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989910956)

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -119,7 +90,34 @@ server <- function(input, output, session) {
...
+                  failed_job_type == "render_shiny" ~ "Rendering",
+                  failed_job_type == "ctrl_extraction" ~ "Extracting parameters",
+                  TRUE ~ failed_job_type)) |>
+                # map exit codes to something more readable 
+                mutate(failure_reason = case_when(
+                  failure_reason %in% c(1, 2, 134) ~ "failed to run / error during running",
+                  failure_reason == 137 ~ "out of memory",
+                  failure_reason %in% c(255, 15, 130) ~ "process terminated by server",
+                  failure_reason %in% c(13, 127) ~ "configuration / permissions error",
+                  TRUE ~ as.character(failure_reason))) |>
```
dethmasque, line 119: In response to:

> [Do we need to turn this into characters at all?](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989912855)

We did not need to convert prior to `mutate()` but still need to treat all unmapped `exit_code` values as characters, or there will be a type mismatch.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -31,34 +32,7 @@ get_failed_job_data <- function(item, usage) {
     {
```
dethmasque, line 32: > [What are we catching here?](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989916166)

We are reacting to the error returned by `connectapi` when jobs are requested for a content item that has none, like unpublished content.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -74,22 +48,19 @@ get_failed_job_data <- function(item, usage) {
       filter(content_guid == item$content$guid) %>%
       slice_max(timestamp) %>%
       select(timestamp)
-    if (is.na(item$content$title)) {
-      item$content$title <- "" # use empty strings when content is missing title
-    }
-    # return required information from https://github.com/posit-dev/connect/issues/30288 
-    all_failed_jobs <- bind_rows(lapply(seq_len(nrow(failed_jobs)), function(i) {
+    # return required information from https://github.com/posit-dev/connect/issues/30288
+    all_failed_jobs <- map_dfr(seq_len(nrow(failed_jobs)), ~
```
dethmasque, line 52: This was simplified from `all_failed_jobs <- bind_rows(lapply(seq_len(nrow(failed_jobs)), function(i)` in response to:

> [I'm struggling to follow what you're doing here. You're going down the vector of failed jobs, and for each one pulling out details from that list and then enriching it with information about the content item itself, yeah? If yes (but I'm honestly not sure if this is true or not)... Would it be possible to start from failed_jobs and then use dplyr / vector recycling to do this instead?](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989922153)

Here we are generating the dataframe containing all failed jobs for a given content item. If there is more than one failed job then we need to bind all failed job rows with the content and visit information included.

extensions/prototype-content-with-issues-table/get_usage.R
```diff
@@ -16,13 +16,9 @@ usage_dtype <- tibble::tibble(
 
 # A rough implementation of how a new firehose usage function would work in
 # `connectapi`.
-get_usage_firehose <- function(client, from = NULL, to = NULL) {
+get_usage_firehose <- function(client) {
```
dethmasque, line 19: `from` and `to` parameters removed in response to:

> [Here, but also elsewhere in these functions: you've got from/to, but are you using those anywhere here?](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989924592)

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -4,10 +4,11 @@ library(DT)
 library(connectapi)
 library(dplyr)
 library(purrr)
+library(tidyr)
 
 # cache data to disk with a refresh every 8h
 shinyOptions(
-  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 8)
+  cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60 * 12)
```
jonkeane, line 11: Thought while I'm waiting for the cache to rebuild so I can poke at this: if this takes upwards of an hour to run on some servers, maybe we should have the cache be longer than twelve hours so we reduce the likelihood of someone hitting it? Or, we could consider a different caching mechanism (or maybe shiny caching can do this and I / we just don't know about it?) that lets the cache be refreshed in the background so that there's always _something_ available to look at even if it's not the freshest.

extensions/prototype-content-with-issues-table/app.R
jonkeane: Thanks!

extensions/prototype-content-with-issues-table/app.R
jonkeane: Can we detect that there are no jobs before we hit this endpoint? If yes, we could do that here in an `if` / `else` (or we could do it down in [line 117](https://github.com/posit-dev/connect-extensions/blob/43cfec36926851d6a883187fcea3cd14f0a4761d/extensions/prototype-content-with-issues-table/app.R#L117) inside of the `bad_content_df` generation). This way, we know that it's being skipped because there are no jobs and not for some other reason that would get eaten by the try/catch.

extensions/prototype-content-with-issues-table/app.R
jonkeane: Thanks for the simplification. I still think that a `dplyr` pipeline would be simpler and much easier to read here, but I won't hold up merging this just for that.

extensions/prototype-content-with-issues-table/app.R
jonkeane: _nods_ great, this might be worth noting "treat all unmapped exit_code values as characters for type consistency" in a comment above line 119 here

extensions/prototype-content-with-issues-table/get_usage.R
jonkeane: Thanks!

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -4,10 +4,11 @@ library(DT)
 library(connectapi)
 library(dplyr)
 library(purrr)
+library(tidyr)
```
jonkeane, line 7: This is add for `replace_na()` yeah? Are there other functions we use `tidyr` for?

jonkeane (commented):
Thanks for this follow up. I haven't been able to get the demo to load yet, I assume that it is churning through making the cache. I will try and come back to it later and poke around so I can give feedback at an experience-level.

extensions/prototype-content-with-issues-table/app.R
dethmasque: Not until [Marco's PR to add a `failed` query parameter to `v1/content` and failed job data to content items](https://github.com/posit-dev/connect/pull/30414) merges.

extensions/prototype-content-with-issues-table/app.R
dethmasque: I am not sure caching is working as we expect on Connect. Perhaps with min procs set to 1 the cache would rebuild every 12 hours, but as it is now it seems to rebuild when the app is visited and the cache is older than the provided `max_age`.

extensions/prototype-content-with-issues-table/app.R
dethmasque: Yes, added in response to [this suggestion](https://github.com/posit-dev/connect-extensions/pull/33#discussion_r1989918296):

> Would [replace_na()](https://tidyr.tidyverse.org/reference/replace_na.html) do what you're looking for here? 

Only used within datatable render on L121 `mutate(content_title = replace_na(content_title, ""))`.

jonkeane: Got it, that's fine but something for us to remember if that ends up getting factored differently, we can clean this up.

extensions/prototype-content-with-issues-table/app.R
jonkeane: Hmm, ok maybe I'm misunderstanding something then. When I tried this just now for an item on dogfood that has no jobs, I didn't get an error I got a dataframe with 0 rows (or an empty list if I used `get_job_list()`)

```
> client <- connect()
Defining Connect with server: https://rsc.radixu.com/
> item <- content_item(client, "b33cc440-f674-45a0-b5d3-7c0bb162023f")
> get_jobs(item)
# A tibble: 0 × 23
# ℹ 23 variables: id <chr>, ppid <chr>, pid <chr>, key <chr>, remote_id <chr>,
#   content_id <chr>, content_guid <chr>, app_id <chr>, app_guid <chr>,
#   variant_id <chr>, bundle_id <chr>, start_time <dttm>, end_time <dttm>,
#   last_heartbeat_time <dttm>, queued_time <dttm>, queue_name <chr>,
#   tag <chr>, exit_code <int>, status <int>, hostname <chr>, cluster <chr>,
#   image <chr>, run_as <chr>
> get_job_list(item)
list()
```

We might need to treat that differently in the code that follows here, but from what I'm seeing that shouldn't return _an error_. Could you describe / show the exact error(s) that you're catching here? 

<details>
I _also_ tried with an item that is unpublished, and that also didn't error for me, though it did actually have a job. That unpublished-but-with-a-job might be unique to the kind of failure that particular item had:

```
> unpublished_item <- content_item(client, "f8bf721e-06bf-4a69-8adf-15d6fee4bd37")
> get_jobs(unpublished_item)
# A tibble: 1 × 23
  id       ppid    pid   key   remote_id content_id content_guid app_id app_guid
  <chr>    <chr>   <chr> <chr> <chr>     <chr>      <chr>        <chr>  <chr>   
1 47369771 3261649 3261… Z93V… NA        74085      f8bf721e-06… 74085  f8bf721…
# ℹ 14 more variables: variant_id <chr>, bundle_id <chr>, start_time <dttm>,
#   end_time <dttm>, last_heartbeat_time <dttm>, queued_time <dttm>,
#   queue_name <chr>, tag <chr>, exit_code <int>, status <int>, hostname <chr>,
#   cluster <chr>, image <chr>, run_as <chr>
```
</details>

extensions/prototype-content-with-issues-table/app.R
jonkeane: Another idea that might work if the shiny caching isn't doing what we're hoping for: could we store this data as a pin and then update it out of band? I'm not saying that is the only (or even the best way), just a thought that might be helpful here

extensions/prototype-content-with-issues-table/app.R
dethmasque: Perhaps our assumption that the error was due to the content item having no jobs is incorrect, and this is instead malformed content on Dogfood that is missing data we return as part of the `Content` object?

extensions/prototype-content-with-issues-table/app.R
jonkeane: I just came back to poke at the deployed version at https://connect.posit.it/connect/#/apps/bc3dc19c-6c49-4d8b-9db0-c2e629a050b9/ and I'm still getting the error. I assume you're still working on this, but if not just FYI it's still happening. I'm signing off in a bit, but let me know once this is back to ready and I'll take a look (though it might be next week when I get back, sorry!)

extensions/prototype-content-with-issues-table/app.R
jonkeane: Did you get to the bottom of this? Or are you still investigating?

extensions/prototype-content-with-issues-table/app.R
jonkeane: Is this up + testable now?

dethmasque: This was redeployed following your comment on Friday, but we are still subject to the cache expiring and rebuilding when the app is run again (this happened just now, data should populate in ~1h). 

I set minimum processes to `1` on the connect.posit.it deploy of this content to allow for rebuilds without the content being visited. Dogfood refreshes the cache significantly faster and I have yet to see the `libcurl` error when pulling in usage data on that server, so min procs on the Dogfood deploy were left as-is.

dethmasque: I think we can continue to have `bindCache()` generate the RDS caches of the content list, usage, and failed jobs, and start using `pins` to create a board from the failed jobs RDS file prior to checking if the cache expired. We can then use the pinned board to generate the failed jobs table.

extensions/prototype-content-with-issues-table/app.R
jonkeane: Is the list of malformed items on dogfood smallish? If yes, IMHO making a blocklist of content guids that we don't even try from dogfood and then not swallowing errors would be the way to go here rather than wrapping fully in a try/catch like this. It's fine if this is the way forward for now, but it is very much a code funk that we don't want to pull over into our full version of this as we transition into that part of our process.

extensions/prototype-content-with-issues-table/app.R
```diff
@@ -98,28 +68,60 @@ server <- function(input, output, session) {
   # initialize Connect API client
   client <- connect()
   
-  # get content once up front and pass it around for additional filtering
-  content <- get_content(client, limit = inf)
-  
-  # cache content list
+  # TODO: use `v1/content/failed` when #30414 merges so we only list content we
+  # know has failed before, filter to deployed within last 60d for now
```
dethmasque, line 72: As discussed in the Metrics sync today we needed to introduce temporary date range restrictions to improve load times on connect.posit.it. Until we surface failed content job data via `v1/content/failed` we are filtering to content deployed within the last 60 days. Additionally, since completed jobs are reaped every 30 days by default we only collect usage data for visits within that last 30 days.

extensions/prototype-content-with-issues-table/app.R
dethmasque: I removed the `tryCatch` as 404ing content URLs were not encountered while requesting failed jobs against connect.posit.it.

extensions/prototype-content-with-issues-table/app.R
dethmasque: [Leaving my in-code comment here as well](https://github.com/posit-dev/connect-extensions/pull/39#pullrequestreview-2700110285) since it helps address the cache refresh time issues noted above:

> As discussed in the Metrics sync today we needed to introduce temporary date range restrictions to improve load times on connect.posit.it. Until we surface failed content job data via v1/content/failed we are filtering to content deployed within the last 60 days. Additionally, since completed jobs are reaped every 30 days by default we only collect usage data for visits within that last 30 days.

extensions/prototype-content-with-issues-table/get_usage.R
```diff
@@ -45,14 +45,9 @@ get_usage_legacy <- function(client, from = NULL, to = NULL) {
...
-    error = function(e) {
-      print("Could not use firehose endpoint; trying legacy usage endpoints.")
-      get_usage_legacy(client, from, to)
-    }
-  )
+  from <- format(from, "%Y-%m-%dT%H:%M:%SZ")
+  to <- format(to, "%Y-%m-%dT%H:%M:%SZ")
+  # disable firehose for deploying example to connect.posit.it 
+  #get_usage_firehose(client, from, to)
+  get_usage_legacy(client, from, to)
```
jonkeane, line 52: Can you explain the reasoning for why you're removing the `tryCatch` here?

jonkeane (approved):
This is looking much better. There is one substantive question, but once that's answered we can merge this.

extensions/prototype-content-with-issues-table/get_usage.R
dethmasque: Since we removed the `tryCatch` around `get_jobs()` because connect.posit.it does not have malformed content the `tryCatch` here was also deleted, as it was checking for an endpoint we know is not available on connect.posit.it.

extensions/prototype-content-with-issues-table/app.R
dethmasque: We can remove `purrr::map_dfr()` and instead call `dplyr::mutate()` to compile the failed jobs data with content and visit information in the next iteration of this work (#34). Only selecting the failed jobs columns we care about (`exit_code`, `tag`, and `end_time`) will further simplify the new `dplyr` pipeline.

PR #41, opened 2025-03-18
who is visiting this content (hits vs visits)

toph-allen:
Add a configurable "visits" statistic that aggregates hits occurring within a configurable amount of time.

- Dogfood: https://rsc.radixu.com/who-is-visiting-this-content/
- connect.posit.it: https://connect.posit.it/who-is-visiting-this-content/

The statistic is computed like so:

1. Group by user.
2. Compute the time lag between each hit (replace `NA` with `Inf` because that's just the first hit in our data).
3. Filter out hits with a time lag smaller than the value set in the slider in the sidebar.
4. Profit:

Screen Shot 2025-03-18 at 06 14 44 PM

(I renamed the column after recording this gif.)

I suspect there's a natural cutoff around 3–5 minutes, but it'd be interesting and possible with some work to plot the number of visits on the y axis against a range of cutoffs on the x axis for a variety of content items and types (of course, you'd wanna do so on a real server, not Dogfood).

This story was simply about computing this metric, so please excuse the still very minimal UI etc.

I'm not sure if "visits" is the best term, but "sessions" is both a little more confusing to the lay user and already has a technical meaning in Shiny sessions.


extensions/who-is-visiting-this-content/app.R
```diff
@@ -112,30 +121,55 @@ server <- function(input, output, session) {
...
       select(timestamp, full_name, username)
-  }) |> bindCache(date_range()$from_date, date_range()$to_date, input$content_guid)
+  })
 
   aggregated_visits_data <- reactive({
-    usage_data() |>
+    unfiltered_hits <- usage_data() |>
+      filter(content_guid == input$content_guid) |>
+      group_by(user_guid) |>
+      summarize(n_hits = n())
```
toph-allen, line 143: I think in a final version of this, we might not even wanna show the unfiltered "hits" variable, as that might be more confusing. I'm leaving it in as we develop this so that we can see the difference the filtering makes.

extensions/who-is-visiting-this-content/app.R
```diff
@@ -149,7 +202,12 @@ server <- function(input, output, session) {
   )
   output$aggregated_visits <- renderTable(
     aggregated_visits_data() |>
-      rename("Total Visits" = n_visits, "Full Name" = full_name, "Username" = username)
+      rename(
+        "Total Visits" = n_visits,
+        "Number of Hits" = n_hits,
```
dethmasque, line 207: Is this latest code deployed? I am seeing `n_hits` as a column name in the Aggregated Visits tab.

toph-allen: I'm not sure -- will check shortly -- but this code is all gonna change anyway in the next issue where we switch from `DT` to `gt` 😅

toph-allen: I just re-deployed the content, so the correct column names should be showing now.

dethmasque (approved):
Small question on column header title seen on deployed prototype, otherwise LGTM!

toph-allen:
> Small question on column header title seen on deployed prototype, otherwise LGTM!

I re-deployed to make sure, and things looked good to me now -- thanks for checking!

PR #42, opened 2025-03-19
Lint match extension name and directory

dotNomad:
This PR adds another step to our extension linting action that checks that the extension directory name and the `extension.name` part of the `manifest.json` match.

If they do not the linting step fails avoiding issue further on in the workflow.

### DIFF
--- a/.github/actions/lint-extension/action.yml
+++ b/.github/actions/lint-extension/action.yml
@@ -11,6 +11,16 @@ runs:
   using: "composite"
 
   steps:
+    - name: Check extension name matches directory
+      run: |
+        MANIFEST_NAME=$(jq -r '.extension.name' < ./extensions/${{ inputs.extension-name }}/manifest.json)
+        if [ ${{ inputs.extension-name }} != $MANIFEST_NAME ]; then
+          echo "Error: Extension name, directory, and manifest.json mismatch"
+          echo "Extension '${{ inputs.extension-name }}' must be in the folder '/extensions/${{ inputs.extension-name }}'" and have the name '${{ inputs.extension-name }}' in the manifest.json.
+          exit 1
+        fi
+      shell: bash
+
     # Ensures that the manifest.json for the given extension name
     # contains all the required fields for the rest of the release workflow
     - run: |
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -24,8 +24,9 @@ jobs:
         id: changes
         with:
           # Adding a new extension that has a directory that can be TARed?
-          # Add a new line here with the name of the extension and directory path
-          # Be sure the extension name and directory have the same name
+          # Add a new line here with the name of the extension as the name of
+          # the filter and the value as the extension's directory path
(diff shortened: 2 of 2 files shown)


.github/actions/lint-extension/action.yml
```diff
@@ -11,6 +11,16 @@ runs:
   using: "composite"
 
   steps:
+    - name: Check extension name matches directory
+      run: |
+        MANIFEST_NAME=$(jq -r '.extension.name' < ./extensions/${{ inputs.extension-name }}/manifest.json)
+        if [ ${{ inputs.extension-name }} != $MANIFEST_NAME ]; then
+          echo "Error: Extension name, directory, and manifest.json mismatch"
+          echo "Extension '${{ inputs.extension-name }}' must be in the folder '/extensions/${{ inputs.extension-name }}'" and have the name '${{ inputs.extension-name }}' in the manifest.json.
```
cgraham-rs, line 19: I had a little trouble retracing where exactly `inputs.extension-name` came from. It looks like this is the filter name a human manually typed into [the dorny/paths-filter step](https://github.com/posit-dev/connect-extensions/blob/1ed413e8a37357bb1a76395a9984215d8c9723d7/.github/workflows/extensions.yml#L30). 

Some potential improvements:
- Be explicit in informing the user the filter as the source of the name in case _that_ is actually where a typo exists. [We already have code comments](https://github.com/posit-dev/connect-extensions/blob/1ed413e8a37357bb1a76395a9984215d8c9723d7/.github/workflows/extensions.yml#L28) that I think are adequate in educating the user the filter has to match the extension.
- This is extremely nit-picky, is there any areas of improvement for how we refer to these? Referring to extension name in the workflow, but that's really a filter name, and there's also a literal `name:` field in the manifest. Minor opportunity for confusion.
- Go out of our way to further help the user by providing a verbose diff so it's entirely clear which of the things aren't matching.

cgraham-rs (approved):
LGTM. I left one comment to with potential improvements we could discuss/consider.

.github/actions/lint-extension/action.yml
dotNomad: > It looks like this is the filter name a human manually typed into the dorny/paths-filter step

This is correct. If the extension name is typo-ed the first line of this will fail saying that `./extensions/${{ inputs.extension-name}}/manifest.json` doesn't exist:
```
MANIFEST_NAME=$(jq -r '.extension.name' < ./extensions/${{ inputs.extension-name }}/manifest.json)
```
That will be the first clue.

Once that passes then the check is if the manifest has the right `name`. Those in tandem check the filter, directory, and manifest name all together.

I added improved comments for the filter name in: https://github.com/posit-dev/connect-extensions/pull/42/commits/4c76cefdf6482eaece55e5835a0743d4842c6f89

I want to avoid commenting in the lint action itself that the name comes from the filter name since that isn't necessarily true for the complex tasks.

dotNomad: I'm going to get this merged, but we can continue the discussion and follow-up with any other improvements!

PR #43, opened 2025-03-19
Add concurrency to update-extension-list

dotNomad:
This PR adds a concurrency setting to the `update-extension-list` workflow job:

```yaml
  update-extension-list:
    concurrency:
      group: update-extension-list
      cancel-in-progress: false
```

The group is set to `update-extension-list` rather than something like `${{ github.ref }}` to make every run of `update-extension-list` a part of the same group.

The `update-extension-list` job needs the `fetch-releases` job to complete, and for it to contain actual release data which only occurs when running on a `push` (read merge) to `main`. Meaning that a distinction of workflow, branch, or any other concurrency group isn't necessary.

`cancel-in-progress: false` explicitly prevents the concurrency setting from cancelling the currently running job if another `update-extension-list` job is pending. (I believe this is the default, but it is better to be specific here).

The `concurrency` settings further prevent any race conditions when attempting to checkout `main` and make a commit based on the output of the extension list updating script.

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -100,9 +100,8 @@ jobs:
     if: ${{ needs.complex-extension-changes.outputs.publisher-command-center == 'true' }}
     uses: ./.github/workflows/publisher-command-center.yml
 
-
   # All extensions have been linted, packaged, and released, if necessary
-  # Continuing to update the extension list with the latest release data 
+  # Continuing to update the extension list with the latest release data
 
   # Gathers all release data from GitHub releases triggered by this workflow
   # For use in the `update-extension-list` job
@@ -135,6 +134,9 @@ jobs:
   # job output
   update-extension-list:
     runs-on: ubuntu-latest
+    concurrency:
+      group: update-extension-list
+      cancel-in-progress: false
     needs: [fetch-releases]
     # Only runs if there are releases to update the extension list with
     # https://github.com/actions/runner/issues/2205


cgraham-rs (commented):
I'm not entirely certain this is going to work like we want. There's some confusing language in [the concurrency documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios) that implies no matter what, workflows may be automatically cancelled.

> ...there can be at most one running and one pending job in a concurrency group at any time. When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any existing pending job or workflow in the same concurrency group, if it exists, will be canceled and the new queued job or workflow will take its place.

I think for the specific scenario we're attempting to prevent the better choice is a [GitHub merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) may be more appropriate. But I have no personal hands-on experience configuring and testing those. I believe we did attempt to use merge queue in the Connect repo and encountered some edge case due to the way the merge queue uses a new temporary branch to handle the rebasing. But I don't recall what those specific issues were.  Research needed.

dotNomad:
> I'm not entirely certain this is going to work like we want.

Thank you for calling this out. I re-read your comment and the docs and I think you are 100% right.

I don't think merge queues give us what we want. The temporary branch created by the queue will only run one at a time, but it doesn't prevent multiple runs on `main`. I'm going to close this PR up.

PR #44, opened 2025-03-19
Setup first release of Reaper and Publisher Command Center

dotNomad:
This PR sets up the initial releases of two extensions:
- Reaper
- Publisher Command Center

It also utilizes the new version constraint section of the `manifest.json` to broaden the Python versions the two extensions can be used with.

### DIFF
--- a/extensions/publisher-command-center/manifest.json
+++ b/extensions/publisher-command-center/manifest.json
@@ -11,6 +11,11 @@
 			"package_file": "requirements.txt"
 		}
 	},
+	"environment": {
+		"python": {
+			"requires": ">=3.8, <4"
+		}
+	},
 	"packages": {},
 	"files": {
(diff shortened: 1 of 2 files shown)


Lytol (approved):
Ez pz. Also, so exciting to see the version constraints already used!

extensions/publisher-command-center/manifest.json
```diff
@@ -11,6 +11,11 @@
 			"package_file": "requirements.txt"
 		}
 	},
+	"environment": {
+		"python": {
+			"requires": ">=3.8, <4"
```
jonkeane, line 16: If I understand [pep440](https://peps.python.org/pep-0440/#examples) correctly, `"requires": "~=3.8"` would do this same thing, right?

PR #45, opened 2025-03-19
Use ssh-key to bypass PR requirement for extension list updates

dotNomad:
This PR undoes the version bump to the two extensions released in #44 and adds an `ssh-key` to be utilized by the `update-extension-list` action to bypass Pull Request requirements to `main`.

The associated git tags and GitHub releases for the extensions have already been removed to avoid duplication for when we update the extension list with a re-release. 

References:
- https://github.com/orgs/community/discussions/25305
> Anyone with collaborator access to this repository can use these secrets and variables for actions.

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -154,6 +154,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: main
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
 
       - uses: actions/setup-node@v4
         with:
--- a/extensions/publisher-command-center/manifest.json
+++ b/extensions/publisher-command-center/manifest.json
@@ -63,6 +63,6 @@
 		"title": "Publisher Command Center",
 		"description": "A dashboard for publishers to help manage and track their content",
 		"homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/publisher-command-center",
-		"version": "0.0.1"
+		"version": "0.0.0"
 	}
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/extensions.yml
```diff
@@ -154,6 +154,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: main
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
```
cgraham-rs, line 157: I'm trying to think if there is any chance someone could writing a new workflow in a PR that leverages this bypass to do bad things.

dotNomad: It would require they be made a contributor prior, similar to other secrets in our repo.

cgraham-rs (approved):
@dotNomad correctly confirmed that GitHub Secrets are well protected, with the additional layer of protection the contributor list provides.

dotNomad:
> I wonder if it would make sense to send the extensions list to some branch that is not main? Similar to how `gh-pages` is used as something of a special branch. So long as we know how to get to the list reliably, it doesn't _have to be_ on main right

I dug into this a bit by experimenting with what this would look like.

We can create an orphaned branch containing the content list (like this: https://github.com/posit-dev/connect-extensions/tree/content-list) easily enough. The action that runs the script to update, and then commit the changes to the extension list requires that the scripts be present.

Having both the content list and scripts belong to that orphaned branch then makes me wonder if that branch should have some protection on it to avoid force pushes, and require reviewed pull requests. In addition to the potential confusion around an orphaned branch off `main` containing important files.

Then that raises the question again how we can get around branch protection for the branch. If that train of logic makes sense then I think keeping things in `main` is the simpler of the two options.

dotNomad:
I'm going to merge this, and based on further conversation we can move to the orphaned branch method easily enough.

PR #46, opened 2025-03-19
Release reaper and publisher-command-center

dotNomad:
Increases the `version` for both the Reaper and Publisher Command Center extensions to trigger a release.


fizzboop (approved):
🐱 🚀

PR #47, opened 2025-03-20
Add Gallery documentation to the contributing guide

dotNomad:
This PR introduces documentation into the Contributing guide that covers:
- What is needed for content being added to the gallery
- How to add content to the gallery
- How to update content in the gallery
- How to revert an update to content, or remove the content entirely
- Tips for manually testing Pull Requests, and seeing if a PR will trigger a content release

This PR also removes the `connect-extension.toml` template file since it is no longer necessary, and includes a link to the Contributing doc in the README.

### DIFF
--- a/contributing.qmd
+++ b/contributing.qmd
@@ -24,11 +24,11 @@ A file that has the name of your extension, the categories it falls into, and a 
 
 # Connect Gallery
 
-The Connect Gallery displays content that can be easily added to Posit Connect
-servers.
+The Connect Gallery displays apps, dashboards, and reports that can be easily
+added to Posit Connect servers.
 
-This section covers how to ready your content for the gallery, adding it, and
-updating it.
+This section covers how to ready your content for the gallery, add it, and
+release new updates.
 
 ## Additional Requirements
 
@@ -56,7 +56,7 @@ to the `manifest.json`:
 
 It is recommended to begin with `"version": "0.0.0"` to avoid triggering a
 release during development of your content. When you are ready to release to
-the Gallery check the [Adding content to the Connect Gallery](#adding-content-to-the-connect-gallery) section.
+the gallery check the [Adding content to the Connect Gallery](#adding-content-to-the-connect-gallery) section.
 
 ### Language version constraints
 
@@ -250,3 +250,12 @@ And after the pull request is merged:
 
 - Remove all associated releases for the content from the [GitHub Releases page](https://github.com/posit-dev/connect-extensions/releases).
 - Remove all associated tags for the content from the [GitHub Tags page](https://github.com/posit-dev/connect-extensions/tags).
+
+# Removing Content
+
+Removing content from the `extensions/` subdirectory that has already been
+added to the Connect Gallery requires that you follow instructions in the 
+above [Removing an entire piece of content](#removing-an-entire-piece-of-content)
+section.
+
+In addition references to the content in GitHub workflows will need to removed.
--- a/contributing.qmd
+++ b/contributing.qmd
@@ -35,8 +35,8 @@ release new updates.
 ### The `extension` section in `manifest.json`
 
 Content in the gallery requires some additional details in the `manifest.json`
-file to display in the gallery, install on Connect, and automatically
-release using automation in this repository.
+file to display, install on Connect, and automatically release using automation
+in this repository.
 
 Below is an example of the additional details that need to be manually added
 to the `manifest.json`:
@@ -116,6 +116,8 @@ filters: |
   ...
   my-content-name: extensions/my-content-name/**
 ```
+
+A good example is how `reaper` is setup.
 
 This will make any code changes to your content trigger a few things:
 
--- a/contributing.qmd
+++ b/contributing.qmd
@@ -142,24 +142,8 @@ setting up the environment needed to build the content.
 #### Creating a custom workflow
 
 To faciliate control over how the content is being built we utilize custom
-GitHub Workflows, for example the [Publisher Command Center workflow](https://github.com/posit-dev/connect-extensions/blob/main/.github/workflows/publisher-command-center.yml).
-
-A few steps will be common across custom workflows using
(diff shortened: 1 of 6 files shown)


contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
...
+  requests and to be released
+- release: When the `version` in the `manifest.json` is incrased and merged to
+  `main`, releases the content to users of Posit Connect
+
+Lastly set the `version` in the `manifest.json` file to your initial release.
+
+When the changes above are merged the first release will be kicked off, and
+your content will be in the gallery. 🚀
+
+### Adding custom built content
```
dotNomad, line 134: The custom built content contains a lot of links to the github actions / workflows. I wonder if it would be easier to just link to the relevant examples, and rely on the comments rather than keeping these links updated + duplicating some of the information here.

contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
...
+
+Once your content has the requirements above it is ready to be added to the
+Connect Gallery.
+
+The Connect Gallery uses GitHub Workflows to automate releases of new content
+when pull requests in this repo merge into `main`.
+
+To add content to the Connect Gallery, follow the steps below:
+
+### Adding simple content
```
dotNomad, line 97: The creation of simple content feels pretty straight forward, but I feel like the explanation for _what_ simple content is may not be sufficient. Looking for feedback here.

fizzboop: Maybe list examples of what "simple content" means in this context. Something like: Has no API keys, third party dependencies, requiring Node, and etc.

Lytol (approved):
Seems very reasonable!  The true test will be when we all start adding extensions, and we'll update if needed then. :ship: :it:

contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
...
+      the first element in the `versions` array
+2. Revert the content's `version` in the `manifest.json` file to the previous
+   release
+    
+Once the pull request is merged continue with the steps below:
+
+1. Remove the associated release from the [GitHub Releases page](https://github.com/posit-dev/connect-extensions/releases).
+2. Remove the associated tag from the [GitHub Tags page](https://github.com/posit-dev/connect-extensions/tags).
+
+### Removing an entire piece of content
```
cgraham-rs, line 259: I originally expected "removing" to be deleting the content and code from the repo entirely. However, "removing" seems to be describing a scenario where the related code still exists in the repo, but the content is simply deleted from the resulting `extension.json` in order to cause it to not be made visible in the Connect Gallery.

Should we also document the scenario where the author wants to remove all releases and the content code? It will a little more involved as they will need to modify `extensions.yml` and possibly other workflows if it is of the complex content type.

contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
 {{< include _template/connect-extension.qmd >}}
 ```
 
-### `connect-extension.toml`
+# Connect Gallery
 
-A file with the name, title, a brief description of the extension, and the access type it should have on being created
+The Connect Gallery displays content that can be easily added to Posit Connect
```
jonkeane, line 27:
```suggestion
The Connect Gallery displays apps, dashboards, and reports that can be easily added to Posit Connect
```

Trying to see if using slightly more specific words than "content" here is helpful

contributing.qmd
jonkeane: I like this idea of using our examples (well commented) as the "this is how to do this thing"

contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
...
+
+```json {filename="manifest.json"}
+{
+  ...
+  "environment": {
+    "python": {
+      "requires": ">=3.8, <4"
+    },
+    "r": {
+      "requires": ">=4.2, <5"
```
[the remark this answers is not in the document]
dotNomad: @Lytol and I discussed this when pairing to add this. We both agreed that the `>= , <` syntax was more understandable if the reader wasn't familiar with the `~=` syntax, but both would work and resolve the same.

I'm happy to change it. It would avoid thoughts like "Wait Python 4?"

contributing.qmd
dotNomad: Yeah I wrote the removal sections specifically for how to address removal in terms of the Connect Gallery.

> Should we also document the scenario where the author wants to remove all releases and the content code?

I can add a short note about how removal of content's code will need to follow the instructions for removing content from the gallery and references in the workflows.

dotNomad: Added note in https://github.com/posit-dev/connect-extensions/pull/47/commits/32d94e8a6fa183a91ae6ae01431ba3c09b2176bd

contributing.qmd
```diff
@@ -22,10 +22,251 @@ A file that has the name of your extension, the categories it falls into, and a
...
+
+Below is an example of the additional details that need to be manually added
+to the `manifest.json`:
+
+```json {filename="manifest.json"} 
+{
+  ...
+  "extension": {
+    "name": "my-content-name",
+    "title": "My Content Name",
```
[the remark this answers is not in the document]
dotNomad: I believe the Publisher project always set `name: ""`. The [API docs say "Default: ∅"](https://docs.posit.co/connect/api/#post-/v1/content) which isn't clear if it is not defaulted and required, or defaults to empty. The others have `type|null` with `Default: null` so that makes me think it is required.

contributing.qmd
dotNomad: > Has no API keys, third party dependencies, requiring Node, and etc

Simple content could include API keys, 3rd party dependencies, but content that require building Node are definitely an example of custom built content.

I left the explanation for now, but can improve it as we get questions.

PR #48, opened 2025-03-20
who is visiting this content (plot and updated display edition)

toph-allen:
Links to deployed versions can be found on the attached issue on the Connect repo.

- Updated input workflow, Shiny bookmarking
- Switch to `reactable`
- Click on a row to filter data by that visitor
- Layout updates
- Email viewers, individually and in an email blast
- Improved date filter controls.
- An additional data visualization
- More conventional user filter controls

---

### Initial changes

Incorporate updates post feedback meeting. Changes include:

- Switch to `gt` for table rendering
- Add a plot showing hits per day
- Update layout to better serve as an "inner dashboard"
- Because it's going to become a child view of an outer dashboard, I moved all the controls (e.g. GUID entry) to the sidebar and named them "Dev Controls"

In the story as written, I had talked about removing the "list of hits" table, but it may be useful for discussing the hit aggregation — it's still in there right now.

I had wanted to implement something where you click on a table row to filter the plot by that user, but `gt` doesn't support that; only `DT` supports table selection. This means that switching to `gt` in the outer dashboard is impractical for our use case.

For demo purposes, so that it isn't blank when it loads, it loads with its own guid pre-populated in the Content GUID field.


toph-allen:
I pushed some tiny updates:

- Merged the "Full Name" and "Username" columns into a "Visitor" column displaying "Full Name (username)".
- Renamed some stuff, e.g. "Visits" (the aggregated leaderboard) to "Top Visitors".

marcosnav (commented):
LGTM, a couple questions before approving:
- Do you plan to address the feedback about visits/sessions in this one?
- IIRC you mentioned something about a lag or disconnection between the input and slider, is that still present? I didn't find any issue with that.

jonkeane:
Poking around at this a bit more, one note for future-us 59a1f153-dcd8-44ac-849b-3371829b7002 (leading to https://connect.posit.it/connect/#/apps/59a1f153-dcd8-44ac-849b-3371829b7002) is an item everyone has access to and has a decent number of visits. (the default guid in the app has 0 views, and finding an item from the big list that one has access to + finding the GUID took me some digging).

* It would be very nice if I could select/deselect the visitor list and have the plot reflect that. I might be interested in how frequently _my boss_ looks at the report and currently that's hard to see.
* It would be nice to have a visualization of when the actual visits happen. I see them broken down by day now (and for big things with lots of visits, that makes sense), but if I have a relatively few number of visits (I'm looking at just my boss's visits) it would be good to see those in a more granular way (is it at 7am?, 4pm?, every hour on the hour?) A different visualization might help here too — possibly dots over time instead of bars with counts. These aren't mutually exclusive, we might need both (behind tabs maybe? so they aren't overwhelming each other?)
* Visually, I wonder if it would make sense to have the plot be full width and the table below that? That gives you more space in the plot to be able to have more details.
* Could we add email buttons to each of the visitors too? If we take the perspective of the author of the item being interested in it, they might want to reach out and ask Joe from accounting why they are so interested in the daily blip reports.

toph-allen:
> Poking around at this a bit more, one note for future-us 59a1f153-dcd8-44ac-849b-3371829b7002 (leading to [connect.posit.it/connect#/apps/59a1f153-dcd8-44ac-849b-3371829b7002](https://connect.posit.it/connect/#/apps/59a1f153-dcd8-44ac-849b-3371829b7002)) is an item everyone has access to and has a decent number of visits. (the default guid in the app has 0 views, and finding an item from the big list that one has access to + finding the GUID took me some digging).

I think I'm going to make most of these changes in a version of this that's hosted in the outer dashboard — this will also obviate the problem of having it display a zero-information GUID initially.

> * It would be very nice if I could select/deselect the visitor list and have the plot reflect that. I might be interested in how frequently _my boss_ looks at the report and currently that's hard to see.

Hard agree! This was one of the features that I realized wasn't possible to implement when using `gt` for tables, as it does not yet have a mechanism for row selection. Switching to `reactable`, I'll definitely implement this.

> * It would be nice to have a visualization of when the actual visits happen. I see them broken down by day now (and for big things with lots of visits, that makes sense), but if I have a relatively few number of visits (I'm looking at just my boss's visits) it would be good to see those in a more granular way (is it at 7am?, 4pm?, every hour on the hour?) A different visualization might help here too — possibly dots over time instead of bars with counts. These aren't mutually exclusive, we might need both (behind tabs maybe? so they aren't overwhelming each other?)

Agreed on this. I'm not sure the best way to visualize this info, but I agree that probably using dots on a timeline — or maybe even just having a version with a much shorter bin width — might be good. Whatever it winds up being, yes, I'll put it in a different tab; once I have that set up we can make changes pretty easily.

> * Visually, I wonder if it would make sense to have the plot be full width and the table below that? That gives you more space in the plot to be able to have more details.

Resize the window to make it narrower and watch what happens! I can change the cutoff width for that layout change easily.

I prioritized the side-by-side display thinking through the following: (1) this will be presented in a modal without a sidebar, so will be able to take up more space; (2) when selecting table rows to filter the plot, it's nicer to have the table beside the plot; (3) the table will be pretty narrow, so it looks kinda weird hanging out by itself at the bottom.

Really, the plot should probably take up 2/3 of the space, or expand to fit the space not taken up by the table.

> * Could we add email buttons to each of the visitors too? If we take the perspective of the author of the item being interested in it, they might want to reach out and ask Joe from accounting why they are so interested in the daily blip reports.

Great idea!

extensions/who-is-visiting-this-content/app.R
```diff
@@ -6,70 +6,197 @@ library(dplyr)
...
+              step = 0.5
+            ),
+
+            textInput(
+              "visit_lag_cutoff_text",
+              label = NULL,
+              value = 1
+            ),
+
+            actionButton("clear_cache", "Clear Cache", icon = icon("refresh"))
```
jonkeane, line 90: We can leave it in as we develop if it's useful, but is this something we would keep in the final version?

extensions/who-is-visiting-this-content/app.R
```diff
@@ -125,24 +246,31 @@ server <- function(input, output, session) {
...
   }) |> bindCache("static_key")
 
-  usage_data <- reactive({
+  firehose_usage_data <- reactive({
     get_usage(
       client,
-      from = date_range()$from_date,
-      to = date_range()$to_date + hours(23) + minutes(59) + seconds(59)
+      from = as.POSIXct(date_range()[1]),
+      to = as.POSIXct(date_range()[2]) + hours(23) + minutes(59) + seconds(59)
```
jonkeane, line 262: Any reason to _not_ use the names here? It's mostly obvious, but `from_date` and `to_date` make it super clear as opposed to `1` and `2`

extensions/who-is-visiting-this-content/app.R
```diff
@@ -152,20 +280,29 @@ server <- function(input, output, session) {
...
-      select(timestamp, full_name, username)
+      select(user_guid, display_name, timestamp)
+
+    # Conditionally filter by selection from other table
+    filter_selection <- getReactableState("aggregated_visits", "selected")
+    if (isTruthy(filter_selection)) {
+      filter_guid <- aggregated_visits_data()[filter_selection, ] |>
+        pull(user_guid)
+      all_visits |>
+        filter(if (is.na(filter_guid)) is.na(user_guid) else user_guid == filter_guid)
```
jonkeane, line 294: Do we need this `if` clause here? When is `filter_guid` `NA`? Is it for Anonymous visits only?

Separately: Are you sure that `user_guid == filter_guid` is doing what you want here when there is more than one GUID? I would have expected `user_guid %in% filter_guid`

extensions/who-is-visiting-this-content/app.R
```diff
@@ -177,38 +314,145 @@ server <- function(input, output, session) {
...
+  # Create day by day hit data for plot
+  daily_hit_data <- reactive({
+    all_dates <- seq.Date(date_range()[1], date_range()[2], by = "day")
+
+    all_visits_data() |>
+      mutate(date = date(timestamp)) |>
+      group_by(date) |>
+      summarize(daily_visits = n(), .groups = "drop") |>
+      tidyr::complete(date = all_dates, fill = list(daily_visits = 0))
+  })
```
jonkeane, line 336: If we aggregate at the day level when we plot (have you tried using `scale_x_date(date_breaks = "days")` https://ggplot2.tidyverse.org/reference/scale_date.html#arg-date-breaks ?), that might help us in making the more detailed plot where we see over time since we're not collapsing the data until we plot it.

extensions/who-is-visiting-this-content/app.R
```diff
@@ -177,38 +314,145 @@ server <- function(input, output, session) {
...
+  # Render content metadata and other text ----
+
+  output$filter_message <- renderUI({
+    # req(getReactableState("aggregated_visits", "selected"))
     hits <- all_visits_data()
     glue(
-      "Content '{content_title}' had {nrow(hits)} visits between ",
-      "{min(hits$timestamp)} and {max(hits$timestamp)}."
+      "{nrow(hits)} visits between ",
+      "{date_range()[1]} and {date_range()[2]}."
```
jonkeane, line 397: Same comment here about `from_date` and `to_date` it's _pretty obvious_ but if we have the names, that would be lovely to use

jonkeane (commented):
Looking good. One substantive code question, I'm also going to add my functionality feedback from slack here so it's in a more concrete place:

* the Top Visitors and the List of Visits tabs are interlinked, yeah? So when I select someone from the Top Visitors tab, it changes what is in the List of Vists. I like that I can filter the List of Visits tab, but it's a bit weird UX that that filter control is on the other tab (that is then hidden)

* I would love to be able to have a finer-grain plot of the visits. For example, when I have someone that has visited a bunch, I would love to see some visualization that represents each visit clearly (rather than always binning them into days)

extensions/who-is-visiting-this-content/app.R
toph-allen: I think it's mostly been useful during development, and could probably be removed in the final version.

(Part of me thinks, it'd be good to leave it in as a failsafe if we have a 24-hour window on the cache. That honestly speaks to a need to rethink the caching behavior more than keeping the button in, though. I honestly would love more experience with how performant the firehose is on servers with large usage databases.)

extensions/who-is-visiting-this-content/app.R
toph-allen: The switch from names to indices happened because I switched from my toy fixed date range to wrapping the `dateRangeInput()`, which just provides indices.

But I think that since we're wrapping it anyway, we could make the values named to make it clearer elsewhere in the code. :)

extensions/who-is-visiting-this-content/app.R
toph-allen: The code that was up when you reviewed was before I had implemented the multiple filter behavior. Now this code reads:

```r
    if (length(input$selected_users) > 0) {
      all_visits <- filter(all_visits, user_guid %in% input$selected_users)
    }
    all_visits
```

jonkeane: Yes, that is definitely clearer that way for sure

extensions/who-is-visiting-this-content/app.R
toph-allen: I had already been using `scale_x_datetime()` in the new plot, and this prompted me to also add `scale_x_date()` to the old plot, which looks nicer, but doesn't work for longer date ranges.

I tried to just specify minor breaks to at least have the nice visual rhythm provided by daily lines, but ggplotly does not respect minor breaks

https://github.com/plotly/plotly.R/issues/952

toph-allen: Going to try calculating the ticks by like, looking at the total number of days.

jonkeane: _nods_ yeah I always get tripped up by the ways ggplotly doesn't totally work. I think it's ok to leave this as is now that we have the timeline view especially. It would be nice to be able to be pure(r) in the separation between our data visualization code and our data munging, but it's ok if we can't for other reasons like this

extensions/who-is-visiting-this-content/app.R
jonkeane: > (Part of me thinks, it'd be good to leave it in as a failsafe if we have a 24-hour window on the cache. That honestly speaks to a need to rethink the caching behavior more than keeping the button in, though. I honestly would love more experience with how performant the firehose is on servers with large usage databases.)

💯 and also: the cache behavior when this is wrapped up in the larger list view / outerdashboard will need to be unified there too (e.g. blowing and regening the cache probably should be at that higher level and true for everything)

extensions/who-is-visiting-this-content/app.R
toph-allen: I'll leave this convo unresolved for now, since we _would_ like to be able to have better ticks.

At least mousing over a column tells you what date it is.

toph-allen:
@jonkeane I've gone as far as I can this week — fixed all the bugs I could find!

marcosnav (commented):
Minor comments. One finding.

While trying it out, noticed there is a blinking error, nothing big since it corrects itself. However it'd be good to fix it for a good UX.

It could be that this gets fixed when merging the single-content into multi-content? I'm assuming this happens while the GUID value propagates and observing methods react to it.

**video**
https://github.com/user-attachments/assets/128c1a42-3337-4761-85de-2349d16a2adc

extensions/who-is-visiting-this-content/README.md
```diff
@@ -0,0 +1,3 @@
+# Content Usage Metrics Dashboard
+
+See what content is popular on your Connect instance. Dive into details for a content item to learn who's viewing it, and when.
```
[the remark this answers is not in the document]
toph-allen: Yeah — I didn't add it to the list of extensions in `.github/workflows/extensions.yml` because the final name will be something more like this, but currently this doesn't match the parent directory name. The next PR in this series will be to integrate the multi-content view into this one, but merging now so that those changes happen in a different PR

marcosnav (approved):

extensions/who-is-visiting-this-content/manifest.json
```diff
@@ -0,0 +1,3107 @@
...
+      "checksum": "5b2476731ce8e473c09abd1c593ccfa7"
+    }
+  },
+  "users": null,
+  "extension": {
+    "name": "content-usage-metrics-dashboard",
+    "title": "Content Usage Metrics Dashboard",
+    "description": "See what content is popular on your Connect instance. Dive into details for a content item to learn who's viewing it, and when.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/who-is-visiting-this-content",
+    "tags": ["R", "shiny", "usage-metrics"],
```
dotNomad, line 3098: We don't have any tags yet, so this is going to fail linting if / when this is added to the workflow (just a heads up)

extensions/who-is-visiting-this-content/manifest.json
```diff
@@ -0,0 +1,3107 @@
...
+    "get_usage.R": {
+      "checksum": "5b2476731ce8e473c09abd1c593ccfa7"
+    }
+  },
+  "users": null,
+  "extension": {
+    "name": "content-usage-metrics-dashboard",
+    "title": "Content Usage Metrics Dashboard",
+    "description": "See what content is popular on your Connect instance. Dive into details for a content item to learn who's viewing it, and when.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/who-is-visiting-this-content",
```
dotNomad, line 3097: Looks like this may be out of date based on the name above? We can address all of the extension manifest attributes in a follow-up since this has a version of `0.0.0` and isn't going to be released yet.

extensions/who-is-visiting-this-content/manifest.json
```diff
@@ -0,0 +1,3107 @@
...
+    "tags": ["R", "shiny", "usage-metrics"],
+    "minimumConnectVersion": "2025.04.0",
+    "version": "0.0.0"
+  },
+  "environment": {
+    "r": {
+      "requires": ">=4.3, <5"
```
dotNomad, line 3104: The highest depends R I'm seeing in the manifest looks like `4.0`, not sure if we can go down that low, just noting that here.

extensions/who-is-visiting-this-content/app.R
```diff
@@ -6,143 +6,355 @@ library(dplyr)
...
+  )
+}
 
-        textInput(
-          "visit_lag_cutoff_text",
-          label = "Visit Merge Window (sec)",
-          value = 1
-        ),
+server <- function(input, output, session) {
+  # Bookmarking ----
```
dotNomad, line 185: Am I understanding the bookmarking correctly that it only works in "Open Solo" mode?

dotNomad (approved):
Pulled this down, zipped it up into a TAR, and installed it. It worked great.

I did notice a few things noted below, and in individual comments. Nothing that would prevent the current state of this being merged though. Since this isn't going out with this PR everything I noted we can decide _if_ we should tackle it, and create issues for them.

---

I did notice that we don't have fallback behavior if an oauth integration is not associated. 

We have the logs, but other than that I just got a "Disconnected from server" note in the UI.

It would be helpful to have a note in the README for how to setup this dashboard, and potentially fallback behavior to detect if we cannot call the oauth integration endpoints.

This can all be done as a follow-up. If we don't do it here let's make an issue to accomplish that.

---

> While trying it out, noticed there is a blinking error, nothing big since it corrects itself. However it'd be good to fix it for a good UX.

I noticed this as well. Very minor, and this PR is already pretty large so we can possibly do something as a follow-up in the future.

---

Another important follow-up. There is no error behavior if you provide a GUID or URL that is not valid.

extensions/who-is-visiting-this-content/manifest.json
toph-allen: Yeah — I haven't tested it on different R versions yet so was erring on the side of caution.

extensions/who-is-visiting-this-content/app.R
toph-allen: I think so. There are also issues with some aspects of the bookmarking that I need to raise with the Shiny team, but haven't had time to do so yet.

Specifically, when updating a URL's query string via the default bookmarking pathway on Connect, it works fine, but when calling the `updateQueryString()` function manually, Connect adds an extra path segment to the URL. (https://positpbc.slack.com/archives/C0615R4HN9L/p1744902995229319)

extensions/who-is-visiting-this-content/manifest.json
toph-allen: Yes, I think the name of this single-dashboard version (`who-is-viewing-this-content`) isn't going to make any sense when combined into a single dashboard. I'm going to move the files in the following PR when the functionality changes though.

PR #50, opened 2025-03-25
Add Connect integration deployment tests

cgraham-rs:
Adds automated execution of content deployment tests to an actual Connect server, verifying the deployment is successful and the content URL can be accessed and returns a `200`. Across all supported Connect versions.

## High level

- Automatically obtains the list of simple extensions that passed lint and packaging
- Tests each of those extensions across all supported Connect versions
- Collects the pass/fail results from the matrix of test runners and publishes them to the GitHub status screen
- Leverages the pass/fail results to only release extensions that passed ALL of their integration tests

## Details

The initial work was leveraged from https://github.com/posit-dev/posit-sdk-py
- New integration folder to contain all of the integration-specific Makefile, docker compose, intermediary reports and artifacts
- Some additional files and Makefile targets added to the project root to facilitate the integration tests

This additional work was to get things working within the existing workflows
- New action that executes the integration test for a single extension against a single Connect version
- New integration test workflow to orchestrate the full matrix of tests required and collect their results so we know which extensions passed ALL of their integration tests and can continue being processed for release
- Inserted these new processes into the existing extension.yml workflow, updating downstream jobs to only run for extensions that passed ALL of their integration tests

## Example workflow runs

- Single simple extension was changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14068722489/job/39397724251
- Multiple simple and one complex extension was changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14061910126
- With no extensions changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14068809950

### DIFF
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -8,9 +8,9 @@ on:
         required: true
         type: string
     outputs:
-      success_matrix:
+      successful_extensions:
         description: "Extensions that passed all tests"
-        value: ${{ jobs.collect-results.outputs.matrix }}
+        value: ${{ jobs.collect-results.outputs.successful_extensions }}
     secrets:
       CONNECT_LICENSE:
         required: true
@@ -28,11 +28,7 @@ jobs:
         # The `jq` command is "output compact, raw input, slurp, split on new lines, and remove the last element". This results in a JSON array of Connect versions (e.g., ["2025.01.0", "2024.12.0"]).
         run: |
           versions=$(make print-versions | jq -c -Rs 'split("\n") | .[:-1]')
-
-          # Show processed versions
           echo "Versions: $versions"
-
-          # Set output for next jobs
           echo "versions=$versions" >> "$GITHUB_OUTPUT"
 
   connect-integration-test:
@@ -98,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 1
     outputs:
-      matrix: ${{ steps.collect.outputs.success_matrix }}
+      successful_extensions: ${{ steps.collect.outputs.successful_extensions }}
     if: |
       !contains(needs.connect-integration-test.result, 'skipped') &&
       !contains(needs.connect-integration-test.result, 'cancelled')
@@ -119,26 +115,12 @@ jobs:
 
           if [[ -z "$matrix_results" || "$matrix_results" == "{}" ]]; then
             echo "❌ No matrix results found"
-            echo "success_matrix=[]" >> $GITHUB_OUTPUT
+            echo "successful_extensions=[]" >> $GITHUB_OUTPUT
             exit 0
           fi
 
           echo "Debug: Full matrix context:"
           echo "$matrix_results" | jq '.'
-          
-          # Get job results for each matrix combination
-          job_results=$(echo "$matrix_results" | jq -r '.jobs[] | {
-            extension: .matrix.extension,
-            version: .matrix.connect_version,
-            result: .result
-          }')
-          
-          echo "Debug: Processed job results:"
-          echo "$job_results" | jq '.'
-          
-          echo "🔍 Processing matrix results"
-          echo "Versions to check: $all_versions"
-          echo "Extensions to check: $extensions"
           
           # Track extensions that passed ALL version tests
           success_list=()
@@ -180,13 +162,12 @@ jobs:
           done
           
           # Format output
-          success_matrix=$(jq -n --arg arr "$(IFS=,; echo "${success_list[*]}")" \
+          successful_extensions=$(jq -n --arg arr "$(IFS=,; echo "${success_list[*]}")" \
             '$arr | split(",")' -c)
           
-          if [[ -z "$success_matrix" || "$success_matrix" == "[]" ]]; then
+          if [[ -z "$successful_extensions" || "$successful_extensions" == "[]" ]]; then
             echo "⚠️ No extensions passed all tests"
-            echo "success_matrix=[]" >> $GITHUB_OUTPUT
+            echo "successful_extensions=[]" >> $GITHUB_OUTPUT
           else
-            echo "📊 Final success matrix: $success_matrix"
-            echo "success_matrix=$success_matrix" >> $GITHUB_OUTPUT
-          fi
+            echo "📊 Final successful extensions: $successful_extensions"
+            echo "successful_extensions=$successful_extensions" >> $GITHUB_OUTPUT
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -1,0 +1,77 @@
+name: Connect Integration Tests
+
+on:
+  workflow_call:
+    inputs:
+      extension-name:
+        description: 'The name of the extension'
+        required: true
+        type: string
+
+jobs:
+  setup-integration-test:
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.versions.outputs.versions }}
+    steps:
+        - uses: actions/checkout@v4
+        - id: versions
+          working-directory: ./integration
+          # The `jq` command is "output compact, raw input, slurp, split on new lines, and remove the last element". This results in a JSON array of Connect versions (e.g., ["2025.01.0", "2024.12.0"]).
+          run: |
+            versions=$(make print-versions | jq -c -Rs 'split("\n") | .[:-1]')
+            echo "versions=$versions" >> "$GITHUB_OUTPUT"
+
+  connect-integration-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10 # Max time to run the integration tests
+    needs: setup-integration-test
+    strategy:
+      fail-fast: false
+      matrix:
+        CONNECT_VERSION: ${{ fromJson(needs.setup-integration-test.outputs.versions) }}
+    steps:
+      - uses: actions/checkout@v4
+
+      # Here we download the packaged extension artifact created from the calling workflow
+      - uses: actions/download-artifact@v4
+        with:
+          name: ${{ inputs.extension-name }}.tar.gz
+          path: integration/bundles
+
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Write Posit Connect license to disk
+        run: echo "$CONNECT_LICENSE" > ./integration/license.lic
+        env:
+          CONNECT_LICENSE: ${{ secrets.CONNECT_LICENSE }}
+
+      - name: Run integration tests
+        run: |
+          make -C ./integration ${{ matrix.CONNECT_VERSION }} \
+            EXTENSION_NAME=${{ inputs.extension-name }}
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: ${{ inputs.extension-name || matrix.CONNECT_VERSION }} - Integration Test Report
+          path: integration/reports/*.xml
+          retention-days: 7
+
+  integration-test-report:
+    needs: connect-integration-test
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      pull-requests: write
+    if: always()
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          path: artifacts
+      - uses: EnricoMi/publish-unit-test-result-action@v2
+        with:
+          check_name: integration-test-results
+          comment_mode: off
+          files: "artifacts/**/*.xml"
+          report_individual_runs: true
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -78,7 +78,7 @@ jobs:
         id: download
         with:
           path: artifacts
-          pattern: "*Integration Test Report"
+          pattern: "*-test-report"
 
       - uses: EnricoMi/publish-unit-test-result-action@v2
         if: ${{ steps.download.outputs.download-path != '' }}
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -121,6 +121,9 @@ jobs:
             echo "success_matrix=[]" >> $GITHUB_OUTPUT
             exit 0
           fi
+
+          echo "Debug: Matrix results structure:"
+          echo "$matrix_results" | jq '.'
           
           echo "🔍 Processing matrix results"
           echo "Versions to check: $all_versions"
@@ -134,22 +137,23 @@ jobs:
             all_passed=true
             echo "📦 Checking extension: $ext"
             
-          # Process each extension
-          for ext in $(echo "$extensions" | jq -r '.[]'); do
-            all_passed=true
-            echo "📦 Checking extension: $ext"
-            
             # Must pass ALL versions
             for version in $(echo "$all_versions" | jq -r '.[]'); do
-              # Get matrix result directly from needs context
-              matrix_id="${ext}_${version//./_}"  # Create safe ID from ext and version
-              result='${{ needs.connect-integration-test.result }}'
+              echo "🔎 Checking $ext @ $version"
               
-              echo "🔎 $ext @ $version: $result"
+              # Get specific matrix combination result
+              result=$(echo "$matrix_results" | jq -r \
+                --arg ext "$ext" \
+                --arg ver "$version" \
+                '.jobs[] | select(.matrix.extension == $ext and .matrix.connect_version == $ver) | .result')
               
-              if [[ "$result" == "failure" ]]; then
+              if [[ -z "$result" ]]; then
+                echo "❌ No result found for $ext @ $version"
                 all_passed=false
-                echo "❌ Failed: $ext @ $version"
+                break
+              elif [[ "$result" != "success" ]]; then
+                echo "❌ Failed: $ext @ $version ($result)"
+                all_passed=false
                 break
               else 
                 echo "✅ Passed: $ext @ $version"
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -7,6 +7,9 @@ on:
         description: 'The name of the extension'
         required: true
         type: string
+    secrets:
+      CONNECT_LICENSE:
+        required: true
 
 jobs:
   setup-integration-test:
--- a/.github/workflows/connect-integration-tests.yml
+++ b/.github/workflows/connect-integration-tests.yml
@@ -26,14 +26,11 @@ jobs:
       - id: versions
         working-directory: ./integration
         # The `jq` command is "output compact, raw input, slurp, split on new lines, and remove the last element". This results in a JSON array of Connect versions (e.g., ["2025.01.0", "2024.12.0"]).
-        run: |
-          # Show raw make output
-          echo "Debug: Raw versions from make:"
-          make print-versions
-          
-          # Show processed versions
+        run: |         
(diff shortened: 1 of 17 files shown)


integration/tests/posit/connect/test_deployments.py
```diff
@@ -0,0 +1,76 @@
...
+    def teardown_method(self):
+        """Clean up any content created during the test."""
+        self.content.delete()
+        assert self.client.content.count() == 0
+
+    def test_extension_deploys(self):
+        """Test that an Extension can be deployed to Posit Connect."""
+        # Get required environment variables
+        extension_name = os.getenv("EXTENSION_NAME")
+        api_key = os.getenv('CONNECT_API_KEY')
```
dotNomad, line 23: It looks like the `Makefile` handles setting up the env vars for this which is a really nice way to avoid the double checking necessary if they were set in the GitHub workflow/actions individually. 👍

.github/workflows/publisher-command-center.yml
```diff
@@ -59,9 +59,25 @@ jobs:
...
+    uses: ./.github/workflows/connect-integration-tests.yml
+    secrets: inherit
+    with:
+      extensions: '["publisher-command-center"]'  # JSON array format to match the workflow input schema
+
+  release:
+    runs-on: ubuntu-latest
+    needs: [extension, connect-integration-tests]
+    # Only release if the extension passed ALL integration tests
+    if: needs.connect-integration-tests.outputs.successful_extensions != '[]'
```
dotNomad, line 73: Not suggesting we don't add this, but is it necessary here? If any of the jobs in the Connect Integration Tests fail will the `needs` capture that partial failure and not continue without this `if`?

I know its necessary in the `extensions.yml` since we are using `always()` in a few spots so we need to check additional details for success. 🤔

.github/workflows/extensions.yml
```diff
@@ -49,6 +49,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    outputs:
+      successful_extensions: ${{ steps.collect-successes.outputs.successful_extensions }}
```
dotNomad, line 53: It looks like this might have been left in accidentally? The `collect-simple-extensions` does this collection, and I'm not seeing any step that is doing this collection in this job? Maybe I'm missing a detail though - these aren't the easiest files to parse.

.github/workflows/extensions.yml
```diff
@@ -61,9 +63,74 @@ jobs:
...
+  simple-extension-connect-integration-tests:
+    needs: [collect-simple-extensions]
+    uses: ./.github/workflows/connect-integration-tests.yml
+    with:
+      extensions: ${{ needs.collect-simple-extensions.outputs.successful_extensions }}
+    if: fromJSON(needs.collect-simple-extensions.outputs.successful_extensions) != '[]'
+    secrets: inherit
+
+  # Runs the release process for each extension that passed all tests and if the semver is updated.
+  simple-extension-release:
```
dotNomad, line 119: I think we need to setup the `env` with a `GH_TOKEN` for this job similar to `simple-extensions`

.github/workflows/extensions.yml
```diff
@@ -61,9 +63,74 @@ jobs:
...
+  # Runs Connect integration tests for each extension that were successfully linted and packaged
+  simple-extension-connect-integration-tests:
+    needs: [collect-simple-extensions]
+    uses: ./.github/workflows/connect-integration-tests.yml
+    with:
+      extensions: ${{ needs.collect-simple-extensions.outputs.successful_extensions }}
+    if: fromJSON(needs.collect-simple-extensions.outputs.successful_extensions) != '[]'
+    secrets: inherit
+
+  # Runs the release process for each extension that passed all tests and if the semver is updated.
```
dotNomad, line 118: Slight correction here doesn't this run for each extension that was packaged and passed all tests, and only if the semver is updated when merged to `main` release?

Just a clarity thing to avoid making people think that this job always releases.

.github/workflows/extensions.yml
```diff
@@ -61,9 +63,74 @@ jobs:
...
+      - name: Mark successful jobs
+        if: success()
+        run: echo "${{ matrix.extension }}" >> success-${{ matrix.extension }}.txt
+
+      - uses: actions/upload-artifact@v4
+        if: success()
+        with:
+          name: extension-success-${{ matrix.extension }}
+          path: success-${{ matrix.extension }}.txt
+          retention-days: 1
```
dotNomad, line 76: It looks like we can simplify this and `collect-simple-extensions` a bit.

Here we have the `success` files, but each successful run of `simple-extensions` with the matrix results in a `${{ extension-name }}.tar.gz`. Instead of using success files we can re-use the packaged extensions to create the `successful_extensions` output which would reduce our artifacts (clean that up for easier use) and simplify this already complex workflow.

What do you think?

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+on:
+  workflow_call:
+    inputs:
+      extensions:
+        description: "JSON array of extension names to test"
+        required: true
+        type: string
+    outputs:
+      successful_extensions:
+        description: "Extensions that passed all tests"
```
dotNomad, line 12: May be helpful to mention the structure in the description or a comment.

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+jobs:
+  # Determine the Connect versions to test against
+  setup-integration-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      versions: ${{ steps.versions.outputs.versions }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: versions
```
dotNomad, line 27: I have found it helpful to always include a empty line in between steps. Nitpicky, but it helps debugging a lot too.

integration/Makefile
```diff
@@ -0,0 +1,180 @@
...
+	EXTENSION_NAME=$(EXTENSION_NAME) \
+	DOCKER_CONNECT_IMAGE_TAG=$(DOCKER_CONNECT_IMAGE_TAG) \
+	DOCKER_CONNECT_IMAGE=$(DOCKER_CONNECT_IMAGE) \
+	DOCKER_PROJECT_IMAGE_TAG=$(DOCKER_PROJECT_IMAGE_TAG) \
+	PYTEST_ARGS="$(PYTEST_ARGS)" \
+	$(DOCKER_COMPOSE) -p $(PROJECT_NAME)-$(subst .,-,$(CONNECT_VERSION)) up -V --abort-on-container-exit --no-build
+
+
+# Show available versions
+print-versions:
```
dotNomad, line 131: When attempting to use this locally I'm seeing:

```
❯ make print-versions
Makefile:79: warning: overriding commands for target `preview'
Makefile:50: warning: ignoring old commands for target `preview'
preview
```

Do I have to setup something? I was trying to see how this worked for understanding the new action

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+        working-directory: ./integration
+        # The `jq` command is "output compact, raw input, slurp, split on new lines, and remove the last element". This results in a JSON array of Connect versions (e.g., ["2025.01.0", "2024.12.0"]).
```
dotNomad, line 29: Why do we remove the last element?

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+      matrix:
+        extension: ${{ fromJson(inputs.extensions) }}
+    permissions:
+      checks: write
+      pull-requests: write
+    # Only run if tests weren't skipped or cancelled
+    if: |
+      always() && 
+      !contains(needs.connect-integration-test.result, 'skipped') &&
+      !contains(needs.connect-integration-test.result, 'cancelled')
```
dotNomad, line 83: Neat little shortcut you can use an array with `contains`: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+        if: ${{ steps.download.outputs.download-path != '' }}
+        with:
+          check_name: "Integration test results - ${{ matrix.extension }}"
+          comment_mode: off
+          files: "artifacts/**/*.xml"
+          report_individual_runs: true
+
+  # Provides a matrix of extensions that passed all of the Connect integration tests
+  collect-results:
+    needs: [connect-integration-test, setup-integration-test]
```
dotNomad, line 101: Since `connect-integration-test` needs `setup-integration-test` already I don't think this does.

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,198 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      successful_extensions: ${{ steps.collect.outputs.successful_extensions }}
+    if: |
+      !contains(needs.connect-integration-test.result, 'skipped') &&
+      !contains(needs.connect-integration-test.result, 'cancelled')
+    steps:
+      - id: collect
+        run: |
```
dotNomad, line 111: Is there any value to splitting this up into multiple run commands for readability? I'll leave it up to you. It wasn't as bad as I thought once I started reading.

dotNomad (commented):
I'm not the most adept with dockerfiles and makefiles. I did look at them closely, but if you want more of an expert opinion on any of them it may be beneficial to ask for another set of eyes on those 👀

jonkeane:
> Leverages the pass/fail results to only release extensions that passed ALL of their integration tests

I know we don't (yet) have support for specifying a minimum connect version for items in the gallery (so this might be yagni!), but we do anticipate that some items might require a version of connect or newer. We don't need to build this here (it might be YAGNI!), but is what we're building here compatible with that? Are we making something that would be hard to extend in that way? It might be ok to do that, but given that we were already thinking about it it would be good to know how much possible future debt we're taking on there.

cgraham-rs:
> I know we don't (yet) have support for specifying a minimum connect version for items in the gallery (so this might be yagni!), but we do anticipate that some items might require a version of connect or newer. We don't need to build this here (it might be YAGNI!), but is what we're building here compatible with that? Are we making something that would be hard to extend in that way? It might be ok to do that, but given that we were already thinking about it it would be good to know how much possible future debt we're taking on there.

We did discuss this but opted to punt all of those considerations into the future if/when we implement minimum versions. I think we can simply add that logic it into `setup-integration-test` by further filtering the array of Connect versions to be tested.

.vscode/settings.json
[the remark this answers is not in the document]
cgraham-rs: Correct, I did not intend to commit this.

pyproject.toml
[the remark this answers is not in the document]
cgraham-rs: I briefly tried that, things stopped working, and I put it back to where it is currently. However, I did not have the hours of Makefile experience I now have so I'll give it another attempt.

.github/workflows/publisher-command-center.yml
cgraham-rs: You're correct, for complex extensions we don't require this. If there are any integration test failures then the release job would be skipped automatically.

.github/workflows/extensions.yml
cgraham-rs: Good catch. This is absolutely leftover from when I was trying to do the success collection directly within the `simple-extensions` job.

.github/workflows/extensions.yml
cgraham-rs: It looks like you're corret. During my testing I never actually fully released anything so the release action worked without error for the steps it actually executed. But I see the final steps (that I did not exercise) make github API calls.

.github/workflows/connect-integration-tests.yml
cgraham-rs: This code and the first part of the comment was copied directly from `posit-sdk`. The command splits on newline which results in an empty string as the final element. I've made a minor edit to the code comment hoping it is adequate. We can certainly refactor the command to some different parsing command if we think it would be more understandable.

.github/workflows/connect-integration-tests.yml
cgraham-rs: Honestly, I banged my head on this one for hours and finally had to rely heavily on CoPilot to help me get an array of only the extensions that passed ALL of their tests. Functionally it's overly complex for my taste and I don't know that breaking it up would help that aspect.  I did try to ensure the code comments and outputs were verbose enough to help make it as understandable as possible. Happy to revisit.

.vscode/settings.json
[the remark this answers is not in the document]
cgraham-rs: Done.

.github/workflows/publisher-command-center.yml
cgraham-rs: Done.

.github/workflows/connect-integration-tests.yml
```diff
@@ -0,0 +1,200 @@
...
+        # This results in a JSON array of Connect versions (e.g., ["2025.01.0", "2024.12.0"])
+        run: |
+          versions=$(make print-versions | jq -c -Rs 'split("\n") | .[:-1]')
+          echo "Versions: $versions"
+          echo "versions=$versions" >> "$GITHUB_OUTPUT"
+
+  # Run the Connect integration tests for each extension against each Connect version
+  connect-integration-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15 # Max time to run the integration tests
```
jonkeane, line 40: Do we have reason to believe that these will get stuck frequently? If not, I would leave this be the default, it's longer than any of them _should_ take, but it's shorter than rerunning (which we almost certainly would do if any timed out at least once just to check if it's that thing).

integration/rstudio-connect.gcfg
```diff
@@ -0,0 +1,59 @@
...
+
+[TensorFlow]
+Enabled = true
+Executable = /usr/bin/tensorflow_model_server
+
+[RPackageRepository "CRAN"]
+URL = https://packagemanager.posit.co/cran/__linux__/jammy/latest
+
+[RPackageRepository "RSPM"]
+URL = https://packagemanager.posit.co/cran/__linux__/jammy/latest
```
jonkeane, line 52:
```suggestion
[RPackageRepository "CRAN"]
URL = https://packagemanager.posit.co/cran/__linux__/jammy/latest

[RPackageRepository "RSPM"]
URL = https://packagemanager.posit.co/cran/__linux__/jammy/latest
```

Do you need both of these fields given they have the same URL? I know we have them in a few other places in configs like this, but I don't think you do. Mind dropping the RSPM one and seeing if it works?

.github/workflows/connect-integration-tests.yml
cgraham-rs: > Do we have reason to believe that these will get stuck frequently? If not, I would leave this be the default, it's longer than any of them should take, but it's shorter than rerunning (which we almost certainly would do if any timed out at least once just to check if it's that thing).

After experiencing some occasional runner hangs in the Connect repo we started including these timeouts. So I jammed them in here along the way. Otherwise if the runners or the code hang GitHub let's them run for 6 hours (the default) before terminating them.

We can certainly get rid of all of these and address it in the future if we ever see any kind of 6 hour GitHub cancelation.

integration/rstudio-connect.gcfg
cgraham-rs: > Do you need both of these fields given they have the same URL? I know we have them in a few other places in configs like this, but I don't think you do. Mind dropping the RSPM one and seeing if it works?

I simply copy-pasted that from posit-sdk. As you mentioned, it's also in a bunch of our Connect configs including Dogfood. We can try dropping the RSPM and see what happens.

vars.mk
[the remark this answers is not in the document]
cgraham-rs: Note that I removed them after folding everything into just the integration folder. Vars are only in the Makefile now.

pyproject.toml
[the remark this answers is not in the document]
cgraham-rs: Done.

This was fairly complicated as there's a lot of potential circular references with the various Make targets calling each other. But it finally works locally on my Mac and in CI.

.github/workflows/extensions.yml
cgraham-rs: Done.

.github/workflows/extensions.yml
cgraham-rs: Done.

We should pair to test a physical release going all the way through the system.

.github/workflows/extensions.yml
cgraham-rs: Done.

.github/workflows/extensions.yml
cgraham-rs: This is a very smart optimization, nice catch! 👍 

Done, and appears to work well as seen in this run.

.github/workflows/connect-integration-tests.yml
cgraham-rs: Done.

integration/Makefile
cgraham-rs: Fixed.

The issue was with including `preview` in the list of `CONNECT_VERSIONS` and also having a `preview` make target. I rectified that by calling it `connect-preview` in `CONNECT_VERSIONS` so there's no name collisions. This allows us to still manually run `make preview` commands locally, if we want. Which I did while refactoring things. Always happy to discuss alternative solutions.

integration/rstudio-connect.gcfg
cgraham-rs: @jonkeane it might be more useful to [remove it from the Dogfood config](https://github.com/posit-dev/connect/blob/283c2f3b132062d5328f0a7b48e1c4fa7dfcd714/config/dogfood.gcfg#L79-L80). We will have a larger variety of content being deployed there so we can more quickly discover if it breaks anything. Thoughts?

pyproject.toml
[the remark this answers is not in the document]
dotNomad: Thank you for doing that extra work. I think it makes the repo a bit easier to understand 👍

.github/workflows/extensions.yml
dotNomad: We can absolutely do that 👍

.github/workflows/connect-integration-tests.yml
dotNomad: No refactoring needed, its understandable as is - just noting the functionality.

dotNomad:
Taking another look at this, but it looks like `CONNECT_LICENSE` is needed for the `setup-integration-test` jobs? I'm reading it and not quite understanding why it is needed. Any thoughts on that CI error @cgraham-rs?

dotNomad:
> Taking another look at this, but it looks like `CONNECT_LICENSE` is needed for the `setup-integration-test` jobs? I'm reading it and not quite understanding why it is needed. Any thoughts on that CI error @cgraham-rs?

Oh I'm seeing it now, `connect-integration-tests` as a workflow requires it, but it is not being passed. We are using `secrets: inherit`, but both the `publisher-command-center` workflow and the `extensions` workflow aren't setting up that secret.

I'm going to continue with another review from there, but we need to get that fixed up.

dotNomad (commented):
Poured over the current state of things and everything is looking great. Lets get the CI passing, look at the logs, and then I think we should be good to get this in. After that we can pair to do a release making sure that flow works.

dotNomad:
Closing up this PR in favor of #53. This was using a branch on a fork, and we need access to a repo secret to test the workflows

PR #51, opened 2025-03-25
most used content (updated table)

toph-allen:
This PR contains the updated content usage dashboard.

- Dogfood: https://rsc.radixu.com/most-used-content/
- connect.posit.it: https://connect.posit.it/most-used-content/
  - Note: Because this version uses the small-page paginated "legacy" metrics endpoints, it takes a long time to initially load data — maybe about ten minutes to load one week of data. Don't expect to be able to load 30 or 90 days of data on that version. The final version of this content should never use those endpoints, as it's not a good experience, even with caching.
  - It would be better if we could adjust the config for connect.posit.it to enable the dev version of the firehose API. In fact, that might be a good idea, because then we can get a more realistic impression of performance with the firehose API and larger data sizes.

This PR has been open too long. The core `app.R` file has changed significantly, but my hope is that it's pretty self-explanatory to review. The majority of the changed lines are in files like `manifest.json`, the bulk of which is auto-generated. (Aside: I think this was a by-product of the approach of iterating based on feedback in Slack rather than in commits. I'll try to do a better job going forward of explicitly scoping commits, and directly prompting for review to close out the current round of changes.) Happy to pair on a review.

Changes in this version:

- Uses `reactable` instead of `gt`.
- Incorporated the "Visit Merge Window" slider from the content-specific dashboard.
- Rejiggered display of items across a few iterations. Things I'm unsure about:
  - Is "Unique Visitors" worth including? It's really only "Unique Logged-in Visitors", but that's too much text to display on one row in the title. It just be extra clutter.
  - Same for "Last Viewed".
- Added filter by app mode.
- Added two export buttons
  - One downloads a CSV of the raw hits data.
  - The other downloads the currently-displayed table, including a content guid column that is hidden from display. @kmasiello requested that specifically — the kind of thing you could feed into a "content rebuild" script.
- Added a text line showing the total number of visits and content items displayed.
- Added a placeholder modal that displays when you click on a row, which I'm next going to add the content-specific view to.
- Removed the "Last Viewed At" column.
- Added a link to open the Connect content dashboard.
- Added a checkbox to show the GUID.
- Made the date range controller nicer.

Changes to come in a future pull request:

- Do not use the legacy content endpoints?
- Use the visitor's API key and show some kind of "content scope" control (e.g. "All content I can see", "My content", "Content I can contribute to", etc.).
- Possibly add an "Email" icon / button next to the Owner column? I'm not sure if this will feel silly when viewing "My content", but perhaps that's overthinking.
- Possibly integrate the single content dashboard (mega dashboard).

### DIFF
--- a/extensions/most-used-content/www/styles.css
+++ b/extensions/most-used-content/www/styles.css
@@ -23,25 +23,18 @@
   text-transform: uppercase;
 }
 
-.header:hover {
-  background-color: #eee;
-}
-
-
 /* Bar chart styles */
 
 .bar-cell {
   display: flex;
   align-items: center;
   width: 100%;
-  /* gap: 5%; */
 }
 
 .number {
   font-family: "Fira Mono", Consolas, Monaco, monospace;
   white-space: pre;
   font-size: 0.84375rem;
-  /* text-align: right; */
 }
 
 .bar-chart {


toph-allen:
Note to self: Check sparkline math.

Screen Shot 2025-04-03 at 03 23 26 PM@2x

extensions/most-used-content/www/styles.css
```diff
@@ -0,0 +1,62 @@
...
+  font-size: 0.8125rem;
+  font-weight: 400;
+  text-transform: uppercase;
+}
+
+.header:hover {
+  background-color: #eee;
+}
+
+
```
dotNomad, line 30: Looks like this may be unused.

```suggestion
```

extensions/most-used-content/www/styles.css
```diff
@@ -0,0 +1,62 @@
...
+/* Bar chart styles */
+
+.bar-cell {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  /* gap: 5%; */
```
dotNomad, line 37:
```suggestion
```

extensions/most-used-content/www/styles.css
```diff
@@ -0,0 +1,62 @@
...
+}
+
+.number {
+  font-family: "Fira Mono", Consolas, Monaco, monospace;
+  white-space: pre;
+  font-size: 0.84375rem;
+  /* text-align: right; */
```
dotNomad, line 44:
```suggestion
```

extensions/most-used-content/www/styles.css
```diff
@@ -0,0 +1,62 @@
...
+  background-color: #eee;
+}
+
+
+/* Bar chart styles */
+
+.bar-cell {
+  display: flex;
+  align-items: center;
+  width: 100%;
```
dotNomad, line 36: I believe this is unnecessary since your `bar-chart` flex items have `flex-grow: 1` and will automatically take up all of the horizontal space they can.

dotNomad (commented):
I pulled down this down from the `tar.gz` created in CI and it worked great.

I had some nitpicky comments about the CSS, but all of that was really minor.

I'm not the best to review R code, so perhaps it would be best to get another reviewer to take a look at that, but I can dive in if this PR lingers a bit.

extensions/most-used-content/www/styles.css
toph-allen: Thanks!

extensions/most-used-content/manifest.json
```diff
@@ -0,0 +1,2784 @@
...
+      "checksum": "ee5657ccebbcd0e9c0226f784a6a2497"
+    }
+  },
+  "users": null,
+  "extension": {
+    "name": "most-used-content",
+    "title": "Most Used Content",
+    "description": "A dashboard listing the most used content on your Connect instance.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/most-used-content",
+    "version": "0.0.0"
```
fizzboop, line 2782: Should the `version` be set to `0.0.1` to ensure it gets picked up by the workflow?

dotNomad: If we want to release it, I'm not sure if @toph-allen wanted to add with this PR

dotNomad:

toph-allen:
Closing as this content has been turned into the `metrics-usage-dashboard`

dotNomad:
Thank you for cleaning this up 🧹

PR #52, opened 2025-04-08
Add `portfolio-dashboard` to Connect Gallery

fizzboop:
## Intent
Addresses [#30961](https://github.com/posit-dev/connect/issues/30961)

This moves the JumpStart item `portfolio-dashboard` to the Connect Gallery.


extensions/portfolio-dashboard/manifest.json
```diff
@@ -0,0 +1,2377 @@
...
+  "extension": {
+    "name": "portfolio-dashboard",
+    "title": "Portfolio Dashboard",
+    "description": "A shiny application makes it easy to transform your analysis into an interactive dashboard using R so users can ask and answer questions in real-time, without having to touch any code.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/portfolio-dashboard",
+    "version": "0.0.1"
+  },
+  "version": 1,
+  "locale": "en_US",
+  "platform": "4.3.1",
```
jonkeane, line 11: Could we bump this to use `environment.r.requires` as well? I bet it is compatible with at least `>=4.2` but even `>=4.3` would be more flexible

jonkeane: https://github.com/posit-dev/connect/pull/30451 is where we implemented that setting, if it's reference is helpful.

fizzboop: Updated in commit [669da4a](https://github.com/posit-dev/connect-extensions/pull/52/commits/669da4a89e6e946dc162880da04a27b923dbc0f8).

dotNomad:
Let's try out the new integration tests from #53 on this branch when that merges in.

extensions/portfolio-dashboard/manifest.json
```diff
@@ -0,0 +1,2352 @@
...
+    "title": "Portfolio Dashboard",
+    "description": "A shiny application makes it easy to transform your analysis into an interactive dashboard using R so users can ask and answer questions in real-time, without having to touch any code.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/portfolio-dashboard",
+    "version": "0.0.1"
+  },
+  "environment": {
+    "r": {
+      "requires": ">=4.4"
+    }
+  },
```
dotNomad, line 16: I had to re-create the `manfest.json` to get this to pass the integration tests and deploy successfully - alongside the inclusion of the `returns.rds` file.

However some of the packages listed below have:

```json
"Depends": "R (>= 4.4.0), grDevices, graphics, stats, utils",
```

which restricts our R version range here to `>=4.4`. I'm unsure if there is a good way for us to extend this range given the package versions outlined below.

@jonkeane perhaps you have some insight so we can keep this as general as possible for gallery users? I will investigate as well

jonkeane: > alongside the inclusion of the returns.rds file.

Hah, good catch: Thanks, tests!

> However some of the packages listed below have: `>=4.4.0`

Ah, right, you're right. Given the way that our manifest files work for R, we will have to use `>=4.4` here.

*TL;DR, but if anyone is curious!*
 
This dependency on `R >= 4.4` is a side effect of having all of our package versions explicitly listed (and pinned) in the manifest. In this case, recent release of the Matrix R package is more strictly tied to the R version than usual: Matrix 1.7.0 and later requires `R >= 4.4.0` ; and Matrix less than 1.7.0 requires only `R >= 3.5.0`. You can't install the newer version on older Rs, and the older version is hard(er, and depending on where and how effectively not possible) to install on newer versions of R. Though the packages that depend on Matrix are totally fine if they use 1.7.0 or earlier (they don't have _any_ version constrain listed, actually). So this would technically work if we could send a manifest that didn't request a specific version of Matrix, so long as when the environment was constructed on Connect that one picked the right one. But that's not how (dependency) version constraints work today.

dotNomad: Thank you for the very detailed reply 👍 

I'll mark this one as ready to merge with the new knowledge.

jonkeane: And FTR / for history: another thought would be to see if we can get rid of the dependency on Matrix, but it ends up being a (transitive) dependency of ggplot2, which would be hard(er) to give up given how popular and familiar it is with our audience. But if it were less important, that might be a different approach to getting around this one particular issue.

toph-allen: Hmm, yeah. Since manifests must list packages with specific versions, it sounds like R content may in effect be tied to a relatively narrow range of versions?

I guess using Matrix < 1.7.0 would be better given your earlier response @jonkeane but — maybe not worth worrying about in that much detail yet?

PR #53, opened 2025-04-09
Add Connect integration deployment tests

cgraham-rs:
Adds automated execution of content deployment tests to an actual Connect server, verifying the deployment is successful and the content URL can be accessed and returns a `200`. Across all supported Connect versions. 

See original PR #50 for more detailed discussion around these changes.

## High level

- Automatically obtains the list of simple extensions that passed lint and packaging
- Tests each of those extensions across all supported Connect versions, including our preview (dev) image
- Collects the pass/fail results from the matrix of test runners and publishes them to the GitHub status screen
- Leverages the pass/fail results to only release extensions that passed ALL of their integration tests

## Details

The initial work was leveraged from https://github.com/posit-dev/posit-sdk-py
- New integration folder to contain all of the integration-specific Makefile, docker compose, intermediary reports and artifacts
- There were originally some additional files and Makefile targets from the posit-sdk project root that were folded into the integration folder to keep everything concise

This additional work was to get things working within the existing workflows
- New action that executes the integration test for a single extension against a single Connect version
- New integration test workflow to orchestrate the full matrix of tests required and collect their results so we know which extensions passed ALL of their integration tests and can continue being processed for release
- Inserted these new processes into the existing extension.yml workflow, updating downstream jobs to only run for extensions that passed ALL of their integration tests
- A variety of tweaks to the various posit-sdk files to get this all working for us, and to allow the commands to run locally on Mac and in CI. I had particular problems with the Docker venv. But there still should be general parity and familiarity when comparing the two repos in case anyone needs to jump between those particular projects.

## Example workflow runs

- Multiple simple extensions, one complex extension, everything runs integration tests and pass and are ready for release:
https://github.com/posit-dev/connect-extensions/actions/runs/14368086202?pr=53

## Original example workflow runs (before refactoring everything to exist within just the integration folder)

- Single simple extension was changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14068722489/job/39397724251
- Multiple simple and one complex extension was changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14061910126
- With no extensions changed: https://github.com/cgraham-rs/connect-extensions/actions/runs/14068809950


extensions/reaper/app.py
```diff
@@ -1,3 +1,4 @@
+# CHANGE TO FORCE CI
```
dotNomad, line 1: Before merging lets get rid of these changes

dotNomad (approved):
Fantastic work. Thanks for diving into that integration tests `collect-results` bug we found. The new fix looks great.

Prior to merging we should remove the `app.py` changes that triggered workflows.

I'll note some of the additional improvements we discussed:
- adjusting `retention-days` to `1` for in-workflow used artifacts
- improving the `name` of some jobs for easier reading
- improving the summaries for the jobs / workflow for easier reading of the output

All of those can be individual follow-ups we can do. The core work you've done here to get the gallery content tested is fantastic. Marking as approved so you can merge once the `app.py` changes are removed. If you would like a re-review for any reason feel free to re-request it.

PR #56, opened 2025-04-10
Set retention days for data-passing artifacts to 1

dotNomad:
This PR cleans up our `retention-days` on two types of artifacts we create during our workflows. Both are only used for data-passing between matrix jobs and only need to be around until the workflows finish.

This PR sets the `retention-days` to `1` (the minimum) 

References:
- https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period


cgraham-rs (approved):
👍

PR #57, opened 2025-04-10
Improve workflow readability with better names, and add release summaries

cgraham-rs:
Fixes: https://github.com/posit-dev/connect/issues/30990

- Rename integration test job names so they are shorter and more readable in the GitHub action sidebar
- Remove integration test GitHub summary as it was not useful
- Improve release job logging when handling 0.0.0
- Add release GitHub summary so we don't have to dive into logs

### DIFF
--- a/.github/actions/release-extension/action.yml
+++ b/.github/actions/release-extension/action.yml
@@ -42,13 +42,21 @@ runs:
       id: should_release
       run: |
         echo "The manifest version is '$MANIFEST_VERSION' and the released version is '$LATEST_VERSION'"
-        HIGHER_VERSION=$(semver "$MANIFEST_VERSION" "$LATEST_VERSION" | tail -n 1)
-          if [ "$MANIFEST_VERSION" = "$HIGHER_VERSION" ] && [ "$MANIFEST_VERSION" != "$LATEST_VERSION" ]; then
-          echo "🚀 Will release! The manifest version is greater than the released version."
-          echo "should_release=true" >> "$GITHUB_OUTPUT"
-        else
-          echo "😴 Holding back from release: The manifest version is not greater than the released version."
+        
+        # Special case: Never release version 0.0.0
+        if [ "$MANIFEST_VERSION" = "0.0.0" ]; then
+          echo "⚠️ Version 0.0.0 is reserved and will never be released."
           echo "should_release=false" >> "$GITHUB_OUTPUT"
+        else
+          # Normal version comparison logic
+          HIGHER_VERSION=$(semver "$MANIFEST_VERSION" "$LATEST_VERSION" | tail -n 1)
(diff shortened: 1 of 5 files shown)


extensions/integration-session-manager/app.py
```diff
@@ -1,3 +1,4 @@
+# CHANGED TO FORCE CI
```
dotNomad, line 1: As always we can remove this prior to merging

.github/actions/release-extension/action.yml
```diff
@@ -41,14 +41,34 @@ runs:
...
-        else
-          echo "😴 Holding back from release: The manifest version is not greater than the released version."
+        # Add the extension name to the summary header for clarity
+        echo "# Extension: ${{ inputs.extension-name }}" >> $GITHUB_STEP_SUMMARY
+        echo "" >> $GITHUB_STEP_SUMMARY
+        
+        VERSION_INFO="The manifest version is '$MANIFEST_VERSION' and the released version is '$LATEST_VERSION'"
+        echo "$VERSION_INFO"
+        echo "$VERSION_INFO" >> $GITHUB_STEP_SUMMARY
+        echo "" >> $GITHUB_STEP_SUMMARY
```
dotNomad, line 51: We could move this into the if block to avoid the "released version is `0.0.0`" wording that could be confusing.

Then each could have its own version info string.

dotNomad (commented):
Fantastic improvement. The new names and the summary are way helpful.

To get this ready to merge let's take a pass at improving the `### Checking if your extension will release` section of the contributing guide to reference the now, much easier, summary. We can also avoid the "released version is `0.0.0`" wording.

.github/actions/release-extension/action.yml
cgraham-rs: image

dotNomad (approved):
Thank you for updating the docs as well 🚀

cgraham-rs:
@dotNomad One snafu. If there are no simple extensions changed, `simple-extensions-release` runs `always()` and will error as there are no integration-test outputs available since those are skipped. 

See this run: https://github.com/posit-dev/connect-extensions/actions/runs/14409320218
```
[Extension Workflow](https://github.com/posit-dev/connect-extensions/actions/runs/14409320218/workflow)
Error when evaluating 'strategy' for job 'simple-extension-release'. .github/workflows/extensions.yml (Line: 132, Col: 20): Error parsing fromJson,.github/workflows/extensions.yml (Line: 132, Col: 20): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/extensions.yml (Line: 132, Col: 20): Unexpected value ''
```

We removed the logic to handle that scenario in a different PR assuming `needs` would handle it for us but apparently `always()` overrides that.  

I have added that logic back in which generates the behavior in this run: https://github.com/posit-dev/connect-extensions/actions/runs/14410516174

From [the docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds)
> ...If you would like a job to run even if a job it is dependent on did not succeed, use the always() conditional expression...

PR #59, opened 2025-04-11
Add `quarto-stock-report-python` content

cgraham-rs:
Add quarto-stock-report-python from Posit Connect examples.


extensions/quarto-stock-report-python/thumbnail.jpg
dotNomad, line 1: We currently aren't using the thumbnails so I would argue we shouldn't include them in the repo to prevent confusion around whether or not they should be included.

extensions/quarto-stock-report-python/manifest.json
```diff
@@ -0,0 +1,56 @@
+{
+  "version": 1,
+  "locale": "en_US.UTF-8",
+  "metadata": {
+    "appmode": "quarto-static"
+  },
+  "extension": {
+    "name": "quarto-stock-report-python",
+    "title": "Quarto Stock Report using Python",
+    "description": "An example of how you might automate regular updates using a data source",
```
dotNomad, line 10: I'm wondering if the description should be the entire description from the jump start

> This stock report is generated using Quarto and Python. It is an example
of how you might automate regular updates using a data source. The report
also generates a custom email, sharing the results directly to
stakeholders.

I guess it would be cut-off in our current card view 🤔

extensions/quarto-stock-report-python/manifest.json
```diff
@@ -0,0 +1,56 @@
+{
+  "version": 1,
+  "locale": "en_US.UTF-8",
+  "metadata": {
+    "appmode": "quarto-static"
+  },
+  "extension": {
+    "name": "quarto-stock-report-python",
+    "title": "Quarto Stock Report using Python",
```
dotNomad, line 9: Thinking about the titles, not sure if any change is necessary:

The IDE jumps to using `quarto-stock-report-python` which isn't particularly helpful. Publisher uses "Stock Report" to match the Quarto project title. The title you have may be the most helpful.

extensions/quarto-stock-report-python/README.md
```diff
@@ -0,0 +1,20 @@
...
+## Learn more
+
+* [Quarto](https://quarto.org)
+
+## Requirements
+
+* Quarto version 1.4 or higher
+* Python version 3.9 or higher
+
+<!-- NOTE: this file is generated -->
```
dotNomad, line 20:
```suggestion
```

dotNomad: Since it not generated here

extensions/quarto-stock-report-python/manifest.json
```diff
@@ -0,0 +1,56 @@
...
+    "description": "An example of how you might automate regular updates using a data source",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/quarto-stock-report-python",
+    "version": "0.0.0"
```
dotNomad, line 12: Lets release it and get it added to the gallery 🚀

dotNomad (approved):
This is great. Thank you for adding this.

I had some thoughts during review, but the only actionables are:
-  removing the thumbnail
- upping the release so this gets added

dotNomad:
Oh also I will note that I installed the content via the `tar.gz` generated here and it worked perfectly.

extensions/quarto-stock-report-python/manifest.json
cgraham-rs: What is the guidance here? I'm happy to update it to whatever we want. 

Whatever we decide we should note it in the contributing guide, include a screenshot of how it physically appears in the in-app Connect Gallery for reference, and consider adding a max len to our linting to prevent accidental truncation.

extensions/quarto-stock-report-python/README.md
cgraham-rs: Done.

extensions/quarto-stock-report-python/manifest.json
dotNomad: Makes perfect sense. I think the name of the content changes based on how you deploy though which is what I was trying to note above.

extensions/quarto-stock-report-python/thumbnail.jpg
cgraham-rs: Done.

extensions/quarto-stock-report-python/manifest.json
dotNomad: > What is the guidance here? I'm happy to update it to whatever we want.

We have a design discussion on this page a bit later today. Since we can update the description with further releases what you have is totally fine to be merged. My hope is to have some guidance around this from the design discussion.

PR #63, opened 2025-04-14
Remove Quarto website

dotNomad:
This PR removes the Quarto website on this repo in favor of using Markdown for the documents, and relying on the Connect Gallery feature in Connect to showcase the content.

This removes the need for contributing `connect-extension.qmd` alongside content. I did not remove any of those files however since they can be helpful in future work. They contain `categories` and descriptions for the content that can be turned into content tags for the gallery and `README` files.

The `CONTRIBUTING` guide was updated to not require it as was the `Makefile` target to add new content to the repo.

Features of Quarto were ported as best as possible to markdown
- Callouts > Alerts
- filenames as comments in the code block
- borders > quotes

After this PR is merged I will:
- [ ] Update the repo description to remove the link to `posit-dev.github.io/connect-extension/`
- [ ] Unpublish the GitHub Page


CONTRIBUTING.md
```diff
@@ -201,25 +195,16 @@ you can do that by looking in the GitHub Actions summary. Click on the "Checks"
...
 
-It will either say:
-
-::: {.border}
-The manifest version is '1.1.0' and the released version is '1.0.0'  
-🚀 Will release! The manifest version is greater than the released version.
-:::
-
-or:
+It will either say one of the follow based on the `version` in the
```
cgraham-rs, line 198:
```suggestion
It will say one of the following based on the `version` in the
```

CONTRIBUTING.md
```diff
@@ -110,14 +103,13 @@ To add content to the Connect Gallery, follow the steps below:
...
 
-::: {.callout-note}
-If you can run
-```bash
-tar -czf my-extension-name.tar.gz ./extensions/my-extension-name
-```
-and the resulting TAR file can be published by uploading the bundle to Posit
-Connect, then it is piece of simple content.
-:::
+> [!NOTE]
```
cgraham-rs, line 106: This doesn't preview properly in VS Code, at least on my end. Though the markdown seems correct.

image

CONTRIBUTING.md
```diff
@@ -41,7 +30,9 @@ in this repository.
 Below is an example of the additional details that need to be manually added
 to the `manifest.json`:
 
-```json {filename="manifest.json"} 
+```json
+// manifest.json
+
 {
   ...
```
cgraham-rs, line 37: nit: observed we don't seem to be consistent with ellipsis use in the code blocks. There are only ellipsis before the code here, but both before and after in the next manifest code block, etc.

README.md
```diff
@@ -1,8 +1,8 @@
-# connect-extensions
+# Connect Extension
```
cgraham-rs, line 1:
```suggestion
# Connect Extensions
```

We seem to use the plural version everywhere.

cgraham-rs (approved):
Standard markdown definitely makes simpler and it can be viewed in the GitHub UI. 👍 

One small suggestion is to move the `Connect Gallery` section to the top. Currently it is in the middle of the more procedural instructions which seems to break the flow a little.

CONTRIBUTING.md
dotNomad: Yeah this is part of GitHub's specific flavor of Markdown. I believe the VS Code preview is a narrower markdown syntax.

CONTRIBUTING.md
dotNomad: Jeez I had a few flubs in here. Thank you for catching them.

CONTRIBUTING.md
dotNomad: Good callout. I made it more consistent in https://github.com/posit-dev/connect-extensions/pull/63/commits/13c72478a8530e773c534d5e079ed9ccc6353768

dotNomad:
> One small suggestion is to move the `Connect Gallery` section to the top. Currently it is in the middle of the more procedural instructions which seems to break the flow a little.

I think I will address this in another PR. It is a good callout 👍

PR #64, opened 2025-04-14
add quarto-document jump start example to gallery

toph-allen:
Add the `quarto-document` jump start example to the Gallery.

I'm unsure if the version constraint syntax applies to the `quarto` field of `manifest.json`. I don't _think_ it does — I think it's just the `environment` field — but the baseline Quarto version matching is probably fine, as it defaults to "nearest" and this particular document is extremely simple.

### DIFF
--- a/extensions/quarto-document/manifest.json
+++ b/extensions/quarto-document/manifest.json
@@ -20,9 +20,6 @@
     },
     "index.qmd": {
       "checksum": "52f82c15783f37b2de073e22e1b5a648"
-    },
-    "thumbnail.jpg": {
-      "checksum": "3691931bc6c8ff0b84317140edbd4b8b"
     }
   },
   "users": null,
--- a/extensions/quarto-document/manifest.json
+++ b/extensions/quarto-document/manifest.json
@@ -1,0 +1,36 @@
+{
+  "version": 1,
+  "locale": "en_US",
+  "platform": "4.4.1",
+  "metadata": {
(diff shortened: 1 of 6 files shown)


dotNomad (commented):
> I'm unsure if the version constraint syntax applies to the quarto field of manifest.json. I don't think it does — I think it's just the environment field

You are correct, the version constraint syntax only applies to R and Python.

extensions/quarto-document/manifest.json
```diff
@@ -0,0 +1,36 @@
...
+  "files": {
+    "_quarto.yml": {
+      "checksum": "e498dcc5956dba39a6df10ac9568fcb9"
+    },
+    "index.qmd": {
+      "checksum": "52f82c15783f37b2de073e22e1b5a648"
+    },
+    "thumbnail.jpg": {
+      "checksum": "3691931bc6c8ff0b84317140edbd4b8b"
+    }
```
dotNomad, line 26: The `thumbnail.jpg` isn't present here so we probably want to remove that from the `manifest.json`

dotNomad (approved):
Looks great. I deployed it using the `tar.gz` from CI with no issues.

I did notice the `thumbnail.jpg` was still in the `files` part of the `manifest.json`. That doesn't seem to matter, but would be a good cleanup prior to merging.

toph-allen:
@dotNomad Oh yeah I deleted it but forgot to adjust the manifest. Thanks!

PR #65, opened 2025-04-15
Add Jump Start example stock-api-fastapi

cgraham-rs:


extensions/stock-api-fastapi/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "extension": {
+    "name": "stock-api-fastapi",
+    "title": "FastAPI Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-fastapi",
+    "version": "0.0.0"
+  },
+  "environment": {
+    "python": {
+      "requires": ">=3.11, <4"
```
dotNomad, line 17: Looking at the `pyproject.toml` files for the dependencies it looks like the requirement is `requires-python = '>=3.9'`

```suggestion
      "requires": ">=3.9, <4"
```

extensions/stock-api-fastapi/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "metadata": {
+    "appmode": "python-fastapi",
+    "entrypoint": "main"
+  },
+  "extension": {
+    "name": "stock-api-fastapi",
+    "title": "FastAPI Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-fastapi",
+    "version": "0.0.0"
```
dotNomad, line 13: Let's release it to the gallery 🚀

extensions/stock-api-fastapi/manifest.json
dotNomad: After reviewing I noticed that the `README` has "Python version 3.11 or higher" I'm unsure why

cgraham-rs: The README came from the original Jump Start source. Are we sure we want to go against its guidance?

dotNomad: Yeah lets stick with that. I'm sure there was a reason why `3.11` was put in the README, maybe one of the imports or usages of the standard Python library requires it?

I did a quick look and didn't see anything, but still.

extensions/stock-api-fastapi/README.md
```diff
@@ -0,0 +1,16 @@
+# FastAPI Stock Pricing Service
+
+## About this example
+
+An API allows you to turn your models into production services that other tools and teams can use. APIs are a great way for software engineering teams to use your models without translating them into different languages.
```
toph-allen, line 5: I'm gonna have the exact same comments about these sentences :)

```suggestion
APIs allow you to turn your models into production services that other tools and teams can use without translating them into different languages.
```

extensions/stock-api-fastapi/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": 1,
+  "locale": "en_US.UTF-8",
+  "metadata": {
+    "appmode": "python-fastapi",
+    "entrypoint": "main"
+  },
+  "extension": {
+    "name": "stock-api-fastapi",
+    "title": "FastAPI Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
```
toph-allen, line 11: Same as the other review.

```suggestion
    "description": "A simple example API, showing how you can turn your Python models into production services that other tools and teams can use without translating them into different languages.",
```

toph-allen (approved):
Generally looks good! Same comments about descriptions as the other items. I'm not sure if the required version of Python can be lower, not sure what the implication of those packages is.

extensions/stock-api-fastapi/README.md
dotNomad: Yeah the original descriptions are leaning a lot on the names to differentiate them the API jump starts from each other. 

I do think @cgraham-rs has a point about the descriptions already being provided for us. We can always make an additional pass once we get these in.

dotNomad:

dotNomad (approved):
Thanks for adding that `minimumConnectVersion` this is looking good 😎

extensions/stock-api-fastapi/manifest.json
cgraham-rs: Opting to skip the renaming for now, we will revisit them all prior to Gallery release.

PR #66, opened 2025-04-15
Add Jump Start example stock-api-plumber

cgraham-rs:


extensions/stock-api-plumber/manifest.json
```diff
@@ -0,0 +1,984 @@
...
+  "extension": {
+    "name": "stock-api-plumber",
+    "title": "Stock Pricing Service with Plumber API",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-plumber",
+    "version": "1.0.0"
+  },
+  "environment": {
+    "r": {
+      "requires": ">=3.2, <5"
```
dotNomad, line 21: The packages in the `manifest.json` specify a R version requirement of `>=3.6` so I think we have to update this and the README since we are using a `manifest.json` here which is a bit different than how our jump starts work now.

cgraham-rs: I did not change any of the source code and chose the `requires` version based on the Readme. I believe `rsconnect` will generate manifest by inspecting all of the R files for dependencies and using data from the local R environment. I definitely did **_not_** go out of my way to create a totally clean 3.2 renv when creating the manifest. Should we be doing that?  I think this may be where we ask for some R and manifest generation guidance. And then update the contributing guide to match best practices. @jonkeane @toph-allen

dotNomad: > I believe rsconnect will generate manifest by inspecting all of the R files for dependencies and using data from the local R environment.

This is correct if deploying using `rsconnect`, but the Gallery items get deployed by uploading the bundle meaning that the `manifest.json` we add here is the `manifest` that is used for publishing.

jonkeane: We don't need to go all the way back to R 3.2 for these even if they technically could support that. The tidyverse support policy is a good barometer for "how far back is reasonable" which is R 4.1 these days (I suspect that's also where rig is drawing the line).

jonkeane: > we need to be creating a fresh dev environment that specifically matches the environment described in the Jump Start README and then running rsconnect to build the manifest.

Yeah, given the way that items in the gallery are setup (that they need to have a manifest set) we will need to create a local R environment and create manifest files for these. Ideally we would also save the renv.lock files with these too, since that's what the publisher looks to for deploying. 

But we are definitely feeling the tension / friction with how our manifest files specify strict pins of versions like we did with R and Python versions. This tension might be enough that we need to think about resolving that (e.g. think about updating the way we communicate dependencies to connect) sooner than later. But for now we can create manifests with modern(ish, maybe we target 4.4 for now since we have the matrix issue there...) versions of R and that should be good enough for most current (and future for the next bit) connect instances.

jonkeane: That works, but given the matrix issue, I would say we should do `rig add 4.4` for now (and then swap out 4.4 in all the places below) that should also make installing dependencies quicker since they will be hitting binaries

toph-allen: @cgraham-rs I think the process you outline will work to create a manifest for 4.2.x. I don't think that the resulting content will necessarily work in an R 4.4 environment, however. I ran into this problem too ([here](https://positpbc.slack.com/archives/C08BR1ET8TX/p1744742045548469?thread_ts=1744740197.806919&cid=C08BR1ET8TX))

This is because (1) the packages specified in the manifest are exact versions, and the environment restore process tries to install those exact versions, and (2) some R packages, such as `Matrix`, are compatible with pretty narrow versions of R.

toph-allen: @cgraham-rs I think the process you outline will work to create a manifest for 4.2.x. I don't think that the resulting content will necessarily work in an R 4.4 environment, however. I ran into this problem too ([here](https://positpbc.slack.com/archives/C08BR1ET8TX/p1744742045548469?thread_ts=1744740197.806919&cid=C08BR1ET8TX))

This is because of the problem @jonkeane outlined in their most recent comment: (1) the packages specified in the manifest are exact versions, and the environment restore process tries to install those exact versions, and (2) some R packages, such as `Matrix`, are compatible with pretty narrow versions of R.

I second @jonkeane’s suggestion that we target R 4.4 for now makes sense for now.

toph-allen: @cgraham-rs I think your process is mostly good. I don't think you need to worry about when you install `rsconnect` — when you snapshot an `renv` project, the packages that it records in the lockfile are not {everything in your environment} — they are {every dependency identified in the project} (@jonkeane correct me if I'm wrong there). If you don't explicitly load `rsconnect` or call an `rsconnect` function in the content's code, it shouldn't be recorded in the lockfile or the manifest.

I wouldn't do this step:

> Update the extension README and manifest with the minimum matching R version that works

Manually updating the manifest with the minimum matching R version that works is risk for now due to the aforementioned package version constraint issue.

cgraham-rs: >  I don't think you need to worry about when you install rsconnect — when you snapshot an renv project, the packages that it records in the lockfile are not {everything in your environment} — they are {every dependency identified in the project} 

That makes sense to me. I did not see `rsconnect` update the lockfile, nor would I expect the lockfile to update unless we `snapshot()` again. So it sounds like we're all good there. It's been a couple of years since I've done this type of R work, hence all of the questions.

> Manually updating the manifest with the minimum matching R version that works is risk for now due to the aforementioned package compatibility constraint issue.

I think @jonkeane is suggesting `4.4.3` will be the min supported R version for Jump Starts. So the manifest would look something like.
```
{
  "version": 1,
  "locale": "en_US",
  "platform": "4.2.3",
  "metadata": {
    "appmode": "api",
    "primary_rmd": null,
    "primary_html": null,
    "content_category": null,
    "has_parameters": false
  },
  "extension": {
    "name": "stock-api-plumber",
    "title": "Stock Pricing Service with Plumber API",
    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-plumber",
    "version": "1.0.0"
  },
  "environment": {
    "r": {
      "requires": ">=4.2.3, <5"
    }
  },
  "packages": {
    "R6": {
      "Source": "CRAN",
      "Repository": "https://cloud.r-project.org",
      "description": {
        "Package": "R6",
        "Title": "Encapsulated Classes with Reference Semantics",
        "Version": "2.6.1",
        ...
```

jonkeane: > I think @jonkeane is suggesting 4.4.3 will be the min supported R version for Jump Starts. So the manifest would look something like.

Anything above 4.4 is ok and what we should use for now. 4.4.0, 4.4.1, 4.4.2, and 4.4.3 are all ok with the current version of Matrix. There's a slight disconnect in your comment, @cgraham-rs , between the text (you mention 4.4.3) and the manifest below where you list `>=4.2.3`.

This is what the manifest should look like:

```
{
  "version": 1,
  "locale": "en_US",
  "platform": "4.4.3",
  "metadata": {
    "appmode": "api",
    "primary_rmd": null,
    "primary_html": null,
    "content_category": null,
    "has_parameters": false
  },
  "extension": {
    "name": "stock-api-plumber",
    "title": "Stock Pricing Service with Plumber API",
    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-plumber",
    "version": "1.0.0"
  },
  "environment": {
    "r": {
      "requires": ">=4.4.0, <5"
    }
  },
  "packages": {
```

Note that `requires` there isn't exactly the same as `platform`, since we know that R 4.4.0 is totally fine with the versions of Matrix.

dotNomad:

extensions/stock-api-plumber/manifest.json
cgraham-rs: I've refactored this to use R 4.4.0, all tests pass, and I believe this is ready for approval and release.

.gitignore
```diff
@@ -1,6 +1,7 @@
 # RStudio
 .Rproj.user
 **/renv/
+.Rprofile
```
dotNomad, line 4: We have a couple `.Rprofiles` in the repo already, wonder if we should clean them up and remove them from `manifest.json`s.

To be clear I don't mean in this PR

extensions/stock-api-plumber/manifest.json
```diff
@@ -0,0 +1,914 @@
...
+    "appmode": "api",
+    "primary_rmd": null,
+    "primary_html": null,
+    "content_category": null,
+    "has_parameters": false
+  },
+  "extension": {
+    "name": "stock-api-plumber",
+    "title": "Stock Pricing Service with Plumber API",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
```
dotNomad, line 15: We could include the whole description from the original `README` since the cards no longer truncate.

dotNomad (approved):
🎉  thank you for adding the `minimumConnectVersion` this is look great.

extensions/stock-api-plumber/manifest.json
cgraham-rs: Done. But others were already merged without longer descriptions. As we previously discussed many of these will need re-wording.

PR #67, opened 2025-04-15
Add Jump Start example stock-api-flask

cgraham-rs:


extensions/stock-api-flask/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "extension": {
+    "name": "stock-api-flask",
+    "title": "Flask Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-flask",
+    "version": "0.0.0"
+  },
+  "environment": {
+    "python": {
+      "requires": ">=3.11.7, <4"
```
dotNomad, line 17: Looking at the `requires-python` attributes in the `pyproject.toml`s for the dependencies the lowest I saw was:
`requires-python = ">=3.9"` so I believe we can change this to
```suggestion
      "requires": ">=3.9, <4"
```

extensions/stock-api-flask/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "metadata": {
+    "appmode": "python-api",
+    "entrypoint": "app"
+  },
+  "extension": {
+    "name": "stock-api-flask",
+    "title": "Flask Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-api-flask",
+    "version": "0.0.0"
```
dotNomad, line 13: Let's release it to add it to the gallery 🎉 

```suggestion
    "version": "1.0.0"
```

dotNomad (commented):
Pulled it down and tried it out - worked great. I think we want to get this released, and we can lower the Python version a bit though. 🚀

extensions/stock-api-flask/manifest.json
dotNomad: Made a similar note on another PR not noticing that the Python version in the README mentions `3.10`, but I'm not sure why.

cgraham-rs: Ahhhhh, I was simply copying the same version present in the manifest. But that is generated based on the user's local Python version for the environment they happen to use when running `rsconnect` to generate the manifest.

dotNomad: > that is generated based on the user's local Python version for the environment they happen to use when running rsconnect to generate the manifest.

Bingo

> In case this has any bearing on our user guidance, per the public documentation [the current min supported version of Python for Connect is 3.8](https://docs.posit.co/connect/admin/python/#python-install)

Yeah definitely, our minimum should always be Python 3.8 even with no dependencies restricting it.

cgraham-rs: Good learnings happened here.

I've updated `environment.python.requires` to correctly match the readme.

extensions/stock-api-flask/README.md
```diff
@@ -0,0 +1,16 @@
+# Flask Stock Pricing Service
+
+## About this example
+
+An API allows you to turn your models into production services that other tools and teams can use. APIs are a great way for software engineering teams to use your models without translating them into different languages.
```
toph-allen, line 5: These (preexisting) sentences in the README are… basically saying different flavors of the same thing.

```suggestion
APIs allows you to turn your models into production services that other tools and teams can use without translating them into different languages.
```

toph-allen (approved):
Look fine to me! Some suggestions for descriptive text.

extensions/stock-api-flask/manifest.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": 1,
+  "locale": "en_US.UTF-8",
+  "metadata": {
+    "appmode": "python-api",
+    "entrypoint": "app"
+  },
+  "extension": {
+    "name": "stock-api-flask",
+    "title": "Flask Stock Pricing Service",
+    "description": "APIs are a great way for software engineering teams to use your models without translating them into different languages",
```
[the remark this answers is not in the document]
cgraham-rs: I don't have a strong opinion, but I did not rewrite any of the text from the original Jump Starts under the assumption someone had already carefully chose the existing text since it is visibly presented to every new Connect user in the Jump Start modal.

dotNomad:

PR #73, opened 2025-04-18
Add support for `minimumConnectVersion` specification

dotNomad:
This PR adds a new item to the `extension` section of the `manifest.json` for content in the Gallery - `minimumConnectVersion`. As the name implies it is to set the minimum Connect version that is necessary for the content to work.

The field is required. The `lint` CI task has been updated to check if it is included. The contributing guide has also been updated with instructions for how to include it, recommending `2025.04.0` as a good place to start since _all_ content requires the Gallery feature.

As part of this PR I have updated all of the `manifest.extension` sections, the script that creates the extension list, and the extension list itself to add `minimumConnectVersion` to previous releases of content.


fizzboop (approved):
🐱 This makes sense to me.

PR #74, opened 2025-04-23
Ensure the same uv version is being used everywhere for integration tests

cgraham-rs:
This fixes our integration test makefile to ensure all environments are installing the exact same version of `uv` using the version specified in our `pyproject.toml` 

This fixes an issue experienced during integration tests where the `uv` versions differed and caused the setup to fail.

```
3.700 echo "Installing UV in virtualenv"; \
3.700 uv pip install "uv >= 0.4.27" --quiet; \
3.700 
3.701 Installing UV in virtualenv
3.706 error: Required uv version `==0.6.14` does not match the running version `0.6.16`
3.707 make: *** [Makefile:120: ensure-uv] Error 2
```

### DIFF
--- a/integration/Makefile
+++ b/integration/Makefile
@@ -10,8 +10,8 @@ PROJECT_NAME := connect-extensions
 # Python settings
 PYTHON ?= $(shell command -v python || command -v python3)
 UV ?= uv
-# Read UV version from pyproject.toml
-UV_VERSION ?= $(shell grep -Po '(?<=required-version = ")[^"]+' pyproject.toml 2>/dev/null || echo "0.6.14")
+# Read UV version from pyproject.toml with more portable command
+UV_VERSION ?= $(shell grep 'required-version' pyproject.toml 2>/dev/null | sed -E 's/.*required-version = "([^"]+)".*/\1/' || echo "0.6.14")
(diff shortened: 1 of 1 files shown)


integration/Makefile
```diff
@@ -10,6 +10,8 @@ PROJECT_NAME := connect-extensions
 # Python settings
 PYTHON ?= $(shell command -v python || command -v python3)
 UV ?= uv
+# Read UV version from pyproject.toml
+UV_VERSION ?= $(shell grep -Po '(?<=required-version = ")[^"]+' pyproject.toml 2>/dev/null || echo "0.6.14")
```
dotNomad, line 14: I tried running `grep -Po '(?<=required-version = ")[^"]+' pyproject.toml` locally in the integration folder and got:

```
❮ grep -Po '(?<=required-version = ")[^"]+' pyproject.toml
grep: invalid option -- P
```

Without the flags I don't get any results from that `grep`, but we do have a `pyproject.toml` setup.

I'm a bit confused on the expectation of this command here.

cgraham-rs: Good catch. Fixed.

dotNomad (approved):
Worked perfectly with your new change ✅

PR #76, opened 2025-04-23
Add Connect example landing-page

cgraham-rs:
Note that this specific example is not currently available as a Jump Start in Connect. 

image


extensions/landing-page/manifest.json
```diff
@@ -0,0 +1,40 @@
...
+  "metadata": {
+    "appmode": "static",
+    "primary_rmd": null,
+    "primary_html": "index.html",
+    "content_category": null,
+    "has_parameters": false
+  },
+  "extension": {
+    "name": "landing-page",
+    "title": "Static HTML landing page",
```
dotNomad, line 14: I like the title change here 👌

dotNomad (approved):
> Note that this specific example is not currently available as a Jump Start in Connect.

Interesting, I wonder why that is. Maybe because it didn't fit into the R, Python, Quarto groupings?

---

Worked great 👍

PR #79, opened 2025-04-25
Content Usage Metrics Dashboard

toph-allen:
The long foretold "megadashboard".

This dashboard merges the "outer" `most-used-content` and inner `who-is-viewing-this-content` dashboards. It contains all the features and functionality from those dashboards.

You can find links to view the dashboard in the attached [issue](https://github.com/posit-dev/connect/issues/30933).

The code is pretty complex, but I went through and did a close read at the end of the process to try to organize it in a way that makes sense, erring on the side of verbose comments.

Features:

- View metrics.
- Click on a content item to navigate to a detail view, click the Back button to return to the overview.
- Select a date range to view, adjust visit merge window. These controls operate both on the outer list and the inner detail view.
- View-specific controls:
  - The outer list lets you filter by content type, show/hide the GUID column, and export data.
  - The inner view lets you filter by user and send an email to all selected users.
- Shiny Bookmarking is enabled: navigating directly to a content item's URL (when in "open solo" mode) will load that content item's data*
- Improved user experience upon initial deploy / when no Viewer API Key integration is selected: Instead of crashing, the app displays an undismissable modal dialog with setup instructions.

\* I couldn't get this to actually update the browser history entry, so your browser's back button won't work to navigate back.

---

Bugs / new features to fix:

- [x] When navigating "back", state in conditional sidebar filters is lost.
- [x] Bookmarking is not working (restore does not work)
- [x] Fail gracefully on first deploy (display a specific screen when client creation fails due to missing integration)

---

Subsequent issue: 
- Content scope control: https://github.com/posit-dev/connect/issues/31167


toph-allen:
I removed the `extract_guid` function because it's no longer used (we no longer parse a pasted-in URL)

```
extract_guid <- function(raw) {
  guid_pattern <- "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
  match <- regmatches(raw, regexpr(guid_pattern, raw))
  if (length(match) == 1) {
    match
  } else {
    ""
  }
}
```

extensions/who-is-visiting-this-content/www/styles.css
```diff
@@ -0,0 +1,52 @@
...
+  text-decoration-thickness: max(1px, 0.0625rem);
+}
+
+/* Styles for the column headers */
+.header {
+  border-bottom: 2px solid #555;
+  font-size: 0.8125rem;
+  font-weight: 400;
+  text-transform: uppercase;
+}
```
dotNomad, line 24: Are these being used? Maybe I was looking at the wrong column headers. They seem to use `rt-sort-header`

extensions/who-is-visiting-this-content/www/styles.css
```diff
@@ -0,0 +1,52 @@
...
+.metrics-tbl a {
+  color: inherit;
+  text-decoration: none;
+}
+
+.metrics-tbl a:hover,
+.metrics-tbl a:focus {
+  text-decoration: underline;
+  text-decoration-thickness: max(1px, 0.0625rem);
+}
```
dotNomad, line 16: I tried looking for anchors inside of a `metrics-tbl` I may have missed it though.

extensions/who-is-visiting-this-content/www/styles.css
```diff
@@ -0,0 +1,52 @@
...
+  height: 0.875rem;
+}
+
+.bar {
+  height: 100%;
+}
+
+#visit_merge_window_text {
+  margin-top: -10px !important;
+}
```
dotNomad, line 52: I never like seeing negative margin or `!important`. They have there uses, but this is usually a code smell for me.

Looks like this is getting around the `bslib-gap-spacing` but could we combine the Visit Merge Window bits into a single form group to avoid negative margin?

extensions/who-is-visiting-this-content/app.R
```diff
@@ -2,244 +2,313 @@ library(shiny)
...
+        footer = NULL,
+        HTML(paste(
+          "In the Access panel to the right, click <strong>\"Add integration\"</strong>,",
+          "then select a <strong>Visitor API Key</strong> integration.",
+          "If you don't see one in the list, an administrator must enable this feature on your Connect server.",
+          "See the <a href='https://docs.posit.co/connect/admin/integrations/oauth-integrations/connect/' target='_blank'>Admin Guide</a> for setup instructions.",
+          "<br><br>",
+          "For guidance on using visitor-scoped permissions in your own Connect apps, see the",
+          "<a href='https://docs.posit.co/connect/user/oauth-integrations/#obtaining-a-visitor-api-key' target='_blank'>User Guide</a>.",
+          sep = " "
```
dotNomad, line 252: This improvement is _HUGE_ for Gallery users. 🎉

extensions/who-is-visiting-this-content/app.R
```diff
@@ -633,6 +895,24 @@ server <- function(input, output, session) {
 
     plotlyOutput("visit_timeline_plot", height = paste0(height_px, "px"))
   })
+
+  # Global UI elements ----
+
+  output$page_title_bar <- renderUI({
```
dotNomad, line 901: Perhaps we can provide a URL / GUID somewhere in this title bar?

dotNomad (approved):
I'm not sure if this would be helpful, but now that these are combined I miss the supplying the GUID/URL functionality. I wonder if we can bring that back in someway?

We can search the content (can't search the GUID though). You may have an idea of how that could be incorporated again.

---

Fantastic work. I had a couple things (mainly CSS) that I had questions on. I did find I missed the URL/GUID input method if I knew what I wanted to look at immediately. Getting this into a megadashboard is fantastic.

extensions/who-is-visiting-this-content/www/styles.css
toph-allen: All of your CSS feedback is much appreciated!

toph-allen: Interesting. They definitely aren't active here, but I was following this example from the reactable site: https://glin.github.io/reactable/articles/twitter-followers/twitter-followers.html. I think the default formatting is nicer than what this was going for though (it was specifically replicating FiveThirtyEight's style) so I'll delete this.

toph-allen: Oh LOL the reactable example manually sets the `headerClass` argument.

extensions/who-is-visiting-this-content/www/styles.css
toph-allen: Yeah, this is trying to shrink the spacing between the text and slider input for the visit merge window.

Experimentally, the `!important` isn't required, but negative margin is probably still not ideal.

I think it would be much more fiddly to try to combine them (and I'm not even sure how that'd actually work in the R code). The default styling applied to sidebar input items is only applied to the direct children of the sidebar, and I've had a bunch of other back and forth with this merge working on getting elements to look and behave correctly (e.g. the auto-hiding elements — they were originally in a `div` and everything looked wrong).

I would also definitely be open to deleting the text input for the visit merge window. The text input was a request from Katie back when the maximum range was 600, and it was impossible to get consistent increments. The maximum range is now lower, and it's much easier to step the slider up and down by increments of 1. And the code to sync the slider and text input can still cause some finicky behavior in my experience.

extensions/who-is-visiting-this-content/app.R
toph-allen: I like this idea! I'll put the GUID there for now, as it seems like reasonable info to have while not being so long that it makes things difficult to lay out. (Plus, the Open in Connect button has the URL).

toph-allen: Screen Shot 2025-04-28 at 05 34 50 PM@2x

toph-allen:
> I'm not sure if this would be helpful, but now that these are combined I miss the supplying the GUID/URL functionality. I wonder if we can bring that back in someway?
> 
> We can search the content (can't search the GUID though). You may have an idea of how that could be incorporated again.
> 
> Fantastic work. I had a couple things (mainly CSS) that I had questions on. I did find I missed the URL/GUID input method if I knew what I wanted to look at immediately. Getting this into a megadashboard is fantastic.

@dotNomad I do kind of miss that functionality too. When I was initially thinking of adding it, I thought forward to how to add it in the combined dashboard. Talking to @jonkeane, we thought about using `reactable`'s built in filters. (You can add [custom filters](https://glin.github.io/reactable/articles/custom-filtering.html) that use arbitrary functions to filter the table, so we could process the entered text to extract any GUIDs and search multiple columns, I think.)

Some caveats:
- Given that the GUID is hidden by default, would we use the `Content` column? How would you make it clear you could paste in a URL to filter the list?)
- I was originally gonna write that it's too slow, but it seems faster now. Oh — maybe it's slow in Safari.

Two ideas:

- We could have an external search box like in the [Fuzzy text searching](https://glin.github.io/reactable/articles/custom-filtering.html#fuzzy-text-searching) example, with placeholder text that says something like "Search title, GUID, or URL". I'm not sure if these can search hidden columns — but it looks like you might be able to.
- We could have a button in the title bar on the outer screen that launches a modal to paste in a URL / GUID, and brings up the inner dashboard when clicked.

Initially I liked the idea of a button in the title bar, but filtering the table might be better.

Open questions:

- Worth including vanity URLs? Might be a "nice to have" but ultimately much more fiddly to implement feature

toph-allen:
@dotNomad I created a global search field that searches the title, GUID, dashboard URL, and owner username. (As a result, I removed the filter fields at the top of title and owner username, but could add them back.)

I can't figure out the CSS to make reactable's default search widget wider — any ideas @dotNomad?

extensions/who-is-visiting-this-content/www/styles.css
toph-allen: The little "open" up-right arrow icons are anchors. The second block here isn't needed though.

toph-allen:
@dotNomad Should be ready for another look :)

extensions/who-is-visiting-this-content/www/styles.css
dotNomad: My feedback is relatively minor. If it is more fiddly to combine them this is totally fine.

As far as removal, it is still very nice to have exact input even with the lower window. I prefer text input generally for things like this.

Thank you for the additional info

dotNomad:
> * We could have an external search box like in the [Basic custom search method](https://glin.github.io/reactable/articles/custom-filtering.html#basic-custom-search-method) example, with placeholder text that says something like "Search title, GUID, or URL". I'm not sure if these can search hidden columns — but it looks like you might be able to.

What you went with here is fantastic. I agree with your thoughts around not making it a column filter especially with the GUID originally hidden. The more general search input is much better.

> I created a global search field that searches the title, GUID, dashboard URL, and owner username. (As a result, I removed the filter fields at the top of title and owner username, but could add them back.)

The unified search is much better in my mind. Yes it has limitations since searching for a content name and owner at the same time is not really possible, but the more normal use case is much better and easier to understand since the bar is at the top rather than part of the column.

> I could add them in a more hacky way, but better to wait and do it the right way.

Agreed 👍

dotNomad (approved):
Thank you for adding in that search 🔍 

This dashboard is looking fantastic 🎉

PR #84, opened 2025-04-28
Add support for `requiredFeatures` specification

dotNomad:
Since Connect licenses have product tiers some items in the Gallery need a way to specify that they require features from Connect that may not be on all instances.

Those include
- [API Publishing](https://docs.posit.co/connect/user/fastapi/index.html)
- [OAuth Integrations](https://docs.posit.co/connect/admin/integrations/oauth-integrations/index.html)
- [Current User Execution](https://docs.posit.co/connect/admin/process-management/index.html#runas-current)

This PR adds support in the form of linting, writing to the extension list, and specifying the required features in the `manifest.json` similar to the `minimumConnectVersion`.

This will help users know if a Gallery item will work on their specific instance.

### DIFF
--- a/.github/actions/lint-extension/action.yml
+++ b/.github/actions/lint-extension/action.yml
@@ -36,6 +36,35 @@ runs:
         ' ./extensions/${{ inputs.extension-name }}/manifest.json
       shell: bash
 
+    # Ensures that the manifest.json has only valid requiredFeatures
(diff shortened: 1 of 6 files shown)


fizzboop (dismissed):
this makes sense to me! 🐱

fizzboop (approved):
🐱

PR #86, opened 2025-04-30
feat: add "content scope" control to megadashboard

toph-allen:
## Overview

This PR adds a dropdown menu to the sidebar allowing users to select the set of content they see.

Different account roles on Connect get different options:

- Administrators see "All Content", "Owned + Collaborating", and "Owned"
- Publishers see "Owned + Collaborating" and "Owned"
- Viewers still see the Publisher choices in the dropdown (i.e. the Publisher choices are the fallback), but since they don't own or collaborate on any content, they can't see its usage metrics).

These correspond to the following. I don't love the language on these — I've gone back and forth on many different options and would love feedback):

- "All Content" just gets… all content.
- "Owned + Collaborating": content where `user_role` is `owner` or `editor` — for publishers, this includes both their owned content and content they collaborate on.
- "Owned": content where `user_role` is `owner`

When a Viewer account loads the app, they see a message stating that Viewers can't see usage data. When a Publisher or Admin loads the account, but their filters lead there to be no usage data, a different message is displayed.

Screen Shot 2025-05-01 at 12 49 42 PM
Screen Shot 2025-05-01 at 11 48 42 AM@2x

### Test Deployments

It's deployed to Dogfood and connect.posit.it. Note that the draft deployment on connect.posit.it now displays a narrow set of content. Before, it displayed metrics for all content you had permission to _view_ — now it shows only content that you that is owned by you or that you are collaborating on.

- https://connect.posit.it/megadashboard/
- https://dogfood.team.pct.posit.it/megadashboard/

---

We will want to check this on a live server with different user account roles (e.g. a Publisher and a Viewer account) and app roles (i.e. a Publisher who is be a collaborator on a content item).

---

Earlier, I had planned a broader set of scopes, but the firehose only lets you retrieve metrics for content you own or collaborate on unless you're an Administrator.

> - Administrators see "All Content", "Content I Can See", "Content I Can Edit", and "Content I Own"
> - Publishers see "Content I Can See", "Content I Can Edit", and "Content I Own"
> - Viewers see "Content I Can See"
> 
> These correspond to the following:
> 
> - "All Content" just gets… all content.
> - "Content I Can See": content where `user_role` is not `none` — this is different for admins than "all content on the server" and the most inclusive set for other users.
> - "Content I Can Edit": content where `user_role` is `owner` or `editor` — for publishers, this includes both their owned content and content they collaborate on.
> - "Content I Own": content where `user_role` is `owner` — self-explanatory.

### DIFF
--- a/extensions/who-is-visiting-this-content/app.R
+++ b/extensions/who-is-visiting-this-content/app.R
@@ -457,7 +457,7 @@ server <- function(input, output, session) {
       "all" = content_unscoped(),
       "view" = content_unscoped() |> filter(app_role != "none"),
       "edit" = content_unscoped() |> filter(app_role %in% c("owner", "editor")),
-      "own" = content_unscoped() |> filter(app_role == "owner")
+      "own" = content_unscoped() |> filter(owner_guid == active_user_guid)
     )
   })
 
@@ -480,6 +480,7 @@ server <- function(input, output, session) {
   }) |> bindCache(active_user_guid)
 
   usage_data_raw <- reactive({
+    req(active_user_role %in% c("administrator", "publisher"))
     get_usage(
       client,
       from = date_range()$from,
@@ -524,6 +525,7 @@ server <- function(input, output, session) {
 
   # Create data for the main table and summary export.
   multi_content_table_data <- reactive({
+    req(nrow(usage_data_visits()) > 0)
     usage_summary <- usage_data_visits() |>
       group_by(content_guid) |>
       summarize(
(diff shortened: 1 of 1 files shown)


toph-allen:
Update: This is not ready for prime time as specified.

I had thought that publishers would be able to see metrics data for all content they have access to, but they are only able to see metrics for content they own or can collaborate on. I'll update the scope controls accordingly.

jonkeane (approved):
These changes look good — one thing that I'm curious about (though is a big directional difference, so not in scope for this PR, but this PR made me think about this):

Have you thought about leaning on the content search endpoint to enumerate accessible items and then join those with the usage data we get from the hits endpoint? That might simplify some of the filtering steps (or maybe not!)

PR #88, opened 2025-05-01
Add `requiredEnvironment` to extension list

dotNomad:
This PR includes a new field in the extension list (`extensions.json`) called `requiredEnvironment`. It sits inside of each content version (similar to `minimumConnectVersion`) and uses the `manifest.json`'s `environment` language version constraints specification to fill it. They use the same syntax.

This will enable the Connect UI to display and eventually behave differently based on the required language versions for each content item.

Each of the released items in the extension list have had the new field back-ported by manually adding it.

The CONTRIBUTING guide has been updated to recommend concise version constraints for better display in the Gallery, and `manifest.json` files have been adjusted to follow that guideline.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,14 +143,21 @@ Here is an example of what needs to be included in a `manifest.json`:
   ...
   "environment": {
     "python": {
-      "requires": ">=3.8, <4"
+      "requires": "~=3.8"
     },
     "r": {
-      "requires": ">=4.2, <5"
+      "requires": "~=4.2"
     }
   }
 }
 ```
+
+It is recommended to use the `~=` operator to specify the version of the
+language(s) that the content requires for concise display in the Connect Gallery
+UI.
+
+`~=` is the "Compatible release" operator. `~=4.2` means "any version greater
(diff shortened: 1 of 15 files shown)


toph-allen (approved):
Generally looks good to me, so I'm approving.

I do have a dumb question which might suggest a future change to `CONTRIBUTING.md`: Is `extensions.json` generated or manually composed? I just asked because I checked `CONTRIBUTING.md` and didn't see it mentioned anywhere except for in the sections on removing content (where there's it's typo'd as `extension.json` 🤫).

dotNomad:
> I do have a dumb question which might suggest a future change to `CONTRIBUTING.md`: Is `extensions.json` generated or manually composed?

Not a dumb question at all. The `extensions.json` file is _mainly_ generated. It is additive based on the automation in the repo, and [the `extension-list.ts` script](https://github.com/posit-dev/connect-extensions/blob/main/scripts/extension-list.ts) adds new content and content versions to the file.

However when new properties are added I've been back porting them to the file. Since the script is additive this doesn't create any problems.

toph-allen:
Awesome, that makes sense :)

PR #89, opened 2025-05-01
Update README and makefile to improve local testing experience

cgraham-rs:
Changes are based on feedback from others attempting to run tests locally.


fizzboop (approved):
NICE! thank you for walking through the integration testing and making these adjustments.

PR #95, opened 2025-05-07
Add quarto version constraints to Quarto docs

dotNomad:
This PR updates the `manifest.json` files of Quarto content with the new Quarto version constraints, backports the `requiredEnvironment` details to the released versions, and ups the content's version for re-release so the new manifest can be used when adding to Connect.

PR #96, opened 2025-05-07
Migrate jump start quarto-script-python

cgraham-rs:


dotNomad:
Let's wait until #95 is merged and add Quarto version constraints to this

extensions/quarto-script-python/manifest.json
```diff
@@ -0,0 +1,61 @@
...
+      "package_file": "requirements.txt"
+    }
+  },
+  "files": {
+    "requirements.txt": {
+      "checksum": "5a2845f6ec11bcefce420b49ef94c152"
+    },
+    ".Rprofile": {
+      "checksum": "57f9047cb06fdbef821b61f49e21406e"
+    },
```
dotNomad, line 44: Doesn't look like this file exists.

dotNomad (commented):
Looking good other than the `.Rprofile` mentioned in the `manifest.json`. We should be able to remove that and be good to go.

extensions/quarto-script-python/manifest.json
cgraham-rs: @dotNomad should we include some additional linting for this scenario?

jonkeane: Is the file needed by the script? I would be very surprised if it was since this is a Python script and that's an R profile.

Taking a step back: rather than write a linting job that prospectively checks things like this, we should make sure if this is a problem our integration tests catch it. I suspect in this case, it skips the file when deploying since it doesn't exist and since it's not needed for the actual script all is good. That's ok — this is why we have PRs + reviews for catching things like this.

dotNomad: > I suspect in this case, it skips the file when deploying since it doesn't exist and since it's not needed for the actual script all is good.

Yeah I haven't seen a failure for including a file that doesn't exist in the `manifest.json` (as long as that file isn't needed).

aronatkins: Connect does not enforce the contents of the `files` list. shinyapps.io does.

.github/workflows/extensions.yml
```diff
@@ -38,6 +38,7 @@ jobs:
             stock-api-flask: extensions/stock-api-flask/**
             landing-page: extensions/landing-page/**
             stock-api-fastapi: extensions/stock-api-fastapi/**
+            quarto-script-python: extensions/quarto-script-python/**
```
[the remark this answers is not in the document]
cgraham-rs: To add to this discussion:
- There is also a sibling `quarto-script-r` project, if we are going to rename one we should rename them both
- To create the manifest for this project you have to specify `quarto` using `rsconnect write-manifest quarto .`

extensions/quarto-script-python/manifest.json
cgraham-rs: Done.

dotNomad:
Just swapped this to use the example category.

.github/workflows/extensions.yml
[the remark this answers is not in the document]
cgraham-rs: Following up - do we want to rename this first or merge it as is?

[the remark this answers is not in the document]
dotNomad: I don't mind merging it. If we want to rename we can always do the work to remove / re-release. I'll let @nealrichardson chime in though.

dotNomad (approved):
Thanks for moving this over to `script-python` - it is looking solid 😎

PR #92, opened 2025-05-05
chore: release Content Usage Metrics app

toph-allen:
This PR releases the `megadashboard` to the Gallery.

The dashboard is deployed here:
- https://connect.posit.it/megadashboard/
- https://dogfood.team.pct.posit.it/megadashboard/

Main changes:

- Rename the directory from `who-is-viewing-this-content` to `content-usage-metrics`. I'm open to other suggestions for this name! I had hoped that having the dir rename happen in a single commit with no other changes would help GitHub display a useful diff for all files, but unfortunately, it is not displaying a useful diff for `README.md` or `get_usage.R`, which also had significant changes.
- Flesh out README. I wrote a brief usage guide describing the controls. I just did a single draft, so please give me edits.
- Remove development data shims from `get_usage.R`.
- Changes in `app.R`:
  - Format with `air` (this just happens when I save a file in Positron now).
  - Fix a bug in state restoration for bookmark URLs.
  - Change the cache expiration time to 1 hour. I think this is a better balance for usage. The firehose is pretty fast, and this ensures that users aren't seeing stale data in the app. This change was made as part of removing the dev data shims.
  - Accordingly, change the "Clear Cache" button to an Action Link, to de-emphasize it in the sidebar.
- Update manifest. Increment the app's version to 1.0.0.

PR #98, opened 2025-05-07
fix: rename `content-usage-metrics` to `usage-metrics-dashboard` and add to `extensions.yml`

toph-allen:
Merged #92 without adding the dashboard to `extensions.yml` and without renaming it to `usage-metrics-dashboard`. This PR does those things.


dotNomad (approved):
Got the "Will release!" message, passed CI, wonderful 😎

PR #99, opened 2025-05-08
fix: better first-run experience for `publisher-command-center` extensions

toph-allen:
This PR adds an initial setup screen to the `publisher-command-center` Gallery app.

Screen Shot 2025-05-08 at 11 31 02 AM@2x

The app checks to see if it can instantiate a client with a visitor API key, and if it can't, it displays an instructional panel. The app's router is only instantiated if the check succeeds.

I _think_ I have everything correct with the manifest etc.


dotNomad (dismissed):
Changes look good, and I deployed it. The initial setup experience is _much_ nicer 🎉

dotNomad (approved):
New wording looks great too

PR #100, opened 2025-05-08
Add stock-dashboard-python

cgraham-rs:
Screenshot from manual deployment via Gallery.

image


dotNomad (approved):
🎉  Fantastic

extensions/stock-dashboard-python/manifest.json
```diff
@@ -0,0 +1,44 @@
+{
```
jonkeane, line 1: This is so much nicer without needing to have all of the dependencies listed!

extensions/stock-dashboard-python/requirements.txt
```diff
@@ -0,0 +1,4 @@
+dash==2.16.0
+dash_bootstrap_components==1.5.0
+pandas==2.2.1
+plotly==5.24.1
```
jonkeane, line 4: We don't have to do this today, but we should consider using constraints here rather than strict pins.

jonkeane (approved):
Thanks for this! I'm going to go ahead and merge since you'll be out for a few days, though I know that's slightly non-standard for our team where the author merges.

PR #102, opened 2025-05-08
Some more words in the reaper description

jonkeane:


toph-allen:
I updated the bits and pieces (added a README.md with the same text) and bumped the version number to make sure the updates propagate.

The app itself needs an update before it goes public, so I'll write that up as an issue.

extensions/reaper/manifest.json
```diff
@@ -31,10 +31,10 @@
   },
   "extension": {
     "name": "reaper",
-    "title": "Reaper",
-    "description": "A view to a kill.",
+    "title": "Process monitor",
```
toph-allen, line 34:
```suggestion
    "title": "Process Monitor",
```

I think most of the titles of gallery content are in title case, so I'm updating this to match.

toph-allen:
Follow-up: https://github.com/posit-dev/connect-extensions/issues/105

dotNomad (approved):
Much much better

PR #103, opened 2025-05-08
Add stock-report-jupyter

cgraham-rs:


extensions/stock-report-jupyter/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+      "package_file": "requirements.txt"
+    }
+  },
+  "files": {
+    "stock-report-jupyter.ipynb": {
+      "checksum": "9f26b6fcc64283c7253edd2d37ad417c"
+    },
+    "requirements.txt": {
+      "checksum": "a21501182f426dfe1dc08a839ad47144"
+    }
```
dotNomad, line 36: Looks like the `quandl-wiki-tsla.json.gz` file wasn't included

cgraham-rs: The Notebook includes a big note at the top `This example requires quandl-wiki-tsla.json.gz to be included as an Additional File when publishing to Posit Connect as a notebook with source code.`  It is unclear why `rsconnect` does not source this file automatically as it is read in the Notebook code.

When creating the manifest we have to manually include it.
`rsconnect write-manifest notebook stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz --overwrite`

dotNomad: Yeah that is strange. Thanks for figuring that out and updating this

PR #110, opened 2025-05-09
Add quarto-script-r

cgraham-rs:
Screenshot from manual deploy to Connect Gallery.

image


extensions/quarto-script-r/manifest.json
```diff
@@ -0,0 +1,885 @@
...
+    "appmode": "quarto-static",
+    "primary_rmd": null,
+    "primary_html": null,
+    "content_category": null,
+    "has_parameters": false
+  },
+  "extension": {
+    "name": "quarto-script-r",
+    "title": "Run an R Script using Quarto",
+    "description": "This data transformation script uses Quarto and R. It is an example of how you might automate regular imports and transformations from a data source. JSON and CSV files are exported and made available for download. The script also generates a custom email, with the generated CSV files as attachments.",
```
toph-allen, line 15: Hmm I feel like the description could be improved in the future — it's not just an example script "using Quarto and R", it's an example of how to run R scripts on Connect by adding a _quarto.yml file to turn them into Quarto projects. _Maybe_ something like this (soft suggestion, update the README too if you take it)

```suggestion
    "description": "This example demonstrates running a simple R script on Connect as a Quarto project. It is an example of how you might automate regular imports and transformations from a data source. JSON and CSV files are exported and made available for download. The script also generates a custom email, with the generated CSV files as attachments.",
```

toph-allen (dismissed):
This looks fine to me! Approving but I did also leave a soft suggestion, so if you take it and need a re-review feel free to ping me.

extensions/quarto-script-r/manifest.json
dotNomad: This reads better to me as well as a gallery item.

dotNomad (approved):

extensions/quarto-script-r/manifest.json
jonkeane: Agreed! I've added these to the branch. I know normally you would do this @cgraham-rs , but I want to keep this moving along in your absences. We can always `git revert` if we need to!

jonkeane (approved):
Thanks for this!

Hmm it's a bit weird that _accepting a suggestion_ even still dismisses the old reviews. I know this is just how github + review dismissing works, but ooph that's a little silly!

PR #126, opened 2025-05-12
chore: unpublish `reaper` content

toph-allen:
This PR removes the `reaper` content from `extensions.json` and changes its version to `0.0.0`. I don't have permission to manage tags or releases on this repo, I don't think, so @dotNomad might have to do that.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -360,59 +360,6 @@
       ]
     },
     {
-      "name": "reaper",
-      "title": "Process Monitor",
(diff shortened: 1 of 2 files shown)


jonkeane:
> I don't have permission to manage tags or releases on this repo

You should be good to go now. 

And FTR: we're unreleasing the reaper since the publisher command center has much of this functionality and is more published. We might come back to this later if having a dedicated view to do this is helpful, but want to concentrate on the apps we have up now.

jonkeane (approved):
Thanks for this

dotNomad:
Thanks for following through with the removal of the tags and releases. Confirmed those are 🗑️

PR #128, opened 2025-05-14
feat: better first-run experience in Usage Metrics Dashboard

toph-allen:
The Usage Metrics Dashboard requires a Connect API Key integration to load data. Previously, it popped up a panel with instructions on how to add an integration.

Now, there are two variants of the panel:

- If there is a valid integration (Publisher, Admin) on the server, it grabs one, preferring Publisher integrations, and presents a button that, when clicked, adds the integration and reloads the page.
- If no eligible integration exists, displays a message with instructions on how to add, and a link to the page where Administrators can add integrations.

Play with this:
- Dogfood: https://dogfood.team.pct.posit.it/connect/#/apps/0234151b-9b6e-402c-8058-467617e0641b/access
- connect.posit.it: deployment failure :(

Possibly required features:

- [x] What happens if it's published by a Publisher? The API that we use to set the integrations does not say whether it requires Publisher or Admin credentials! https://docs.posit.co/connect/api/#put-/v1/content/-guid-/oauth/integrations/associations
  - It works fine! :)

Feedback please on:

- [x] I think we can improve the language. "Add Required Integration" might be confusing — especially because we use the term "add integration" both to mean "associate it with this content" and "create a new one". Maybe it's fine because it mirrors the "Add Integration" button in the sidebar, but we could also just call it "Complete Setup", or "Add Integration and Complete Setup".
- [x] @jonkeane We had talked about testing the code more thoroughly. Should I think about that now, for this PR? Happy to wait, but if you think this makes sense as a place to start adding them, I should add them. Although for this particular piece of functionality it'd require mocking a Connect server!

Screen Shot 2025-05-16 at 11 38 12 AM@2x
Screen Shot 2025-05-16 at 11 39 14 AM@2x

### DIFF
--- a/extensions/usage-metrics-dashboard/app.R
+++ b/extensions/usage-metrics-dashboard/app.R
@@ -309,14 +309,12 @@ server <- function(input, output, session) {
           "\"</strong> to this content."
         )
       } else if (nrow(selected_integration) == 0) {
-        print("PLACE ONE")
         integration_settings_url <- publisher_client$server_url(connectapi:::unversioned_url(
           "connect",
           "#",
           "system",
           "integrations"
         ))
-        print("PLACE TWO")
         message <- paste0(
           message,
           "<br><br>",
@@ -324,11 +322,10 @@ server <- function(input, output, session) {
           "<a href='",
           integration_settings_url,
           "' target='_blank'>Integration Settings page</a>. ",
-          "The 'Max Role' field must be set to 'Administrator' or 'Publisher'. ",
+          "The 'Max Role' field must be set to 'Administrator' or 'Publisher'; 'Viewer' will not work. ",
           "See the <a href='https://docs.posit.co/connect/admin/integrations/oauth-integrations/connect/' ",
           "target='_blank'>Admin Guide</a> for more setup instructions."
         )
-        print("PLACE THREE")
       }
 
       footer <- if (nrow(selected_integration) == 1) {
--- a/extensions/usage-metrics-dashboard/app.R
+++ b/extensions/usage-metrics-dashboard/app.R
@@ -15,6 +15,7 @@ shinyOptions(
 )
 
 source("get_usage.R")
+source("integrations.R")
 
 app_mode_groups <- list(
   "API" = c("api", "python-fastapi", "python-api", "tensorflow-saved-model"),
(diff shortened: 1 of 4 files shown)


dotNomad:
> * connect.posit.it: deployment failure :(

Do we know what caused this?

jonkeane:
>  @jonkeane We had talked about testing the code more thoroughly. Should I think about that now, for this PR? Happy to wait, but if you think this makes sense as a place to start adding them, I should add them. Although for this particular piece of functionality it'd require mocking a Connect server!

Let's not block shipping this on writing our first tests for this work. But we *should* do that separately from this PR.

extensions/usage-metrics-dashboard/app.R
```diff
@@ -274,25 +275,82 @@ ui <- function(request) {
 server <- function(input, output, session) {
   # Set up Connect client; handle error if Visitor API Key integration isn't
   # present.
+
+  publisher_client <- connect()
+
+  selected_integration_guid <- reactiveVal(NULL)
+  observeEvent(input$auto_add_integration, {
+    print("Clicked auto add integration button")
```
dotNomad, line 283: Should we clear out prints?

extensions/usage-metrics-dashboard/app.R
```diff
@@ -274,25 +275,82 @@ ui <- function(request) {
...
-          "then select a <strong>Visitor API Key</strong> integration.",
-          "If you don't see one in the list, an administrator must enable this feature on your Connect server.",
-          "See the <a href='https://docs.posit.co/connect/admin/integrations/oauth-integrations/connect/' target='_blank'>Admin Guide</a> for setup instructions.",
+      eligible_integrations <- get_eligible_integrations(publisher_client)
+      selected_integration <- slice_head(eligible_integrations, n = 1)
+      selected_integration_guid(selected_integration$guid)
+
+      message <- paste0(
+        "This content requires a <strong>Visitor API Key</strong> ",
+        "integration to show users the content they have access to."
```
dotNomad, line 301:
```suggestion
        "This content requires a Publisher or Admin <strong>Visitor API Key</strong> ",
        "integration to show users the content they have access to."
```

Minor suggestion to add the "Publisher" / "Admin" prior to the bolded text in the hopes it gets read.

extensions/usage-metrics-dashboard/app.R
dotNomad: Looks like there are a few more in here too.

extensions/usage-metrics-dashboard/app.R
```diff
@@ -388,6 +446,8 @@ server <- function(input, output, session) {
     session$reload()
   })
 
+  # Auto-add integration button ----
+
```
dotNomad, line 450: Left over?

```suggestion
```

dotNomad (commented):
> What happens if it's published by a Publisher? The API that we use to set the integrations does not say whether it requires Publisher or Admin credentials! https://docs.posit.co/connect/api/#put-/v1/content/-guid-/oauth/integrations/associations

Looking at the codebase it looks like this API requires a Publisher or higher auth role so should be good there. We should make an internal ticket to clarify in the docs though.

> I think we can improve the language. "Add Required Integration" might be confusing — especially because we use the term "add integration" both to mean "associate it with this content" and "create a new one". Maybe it's fine because it mirrors the "Add Integration" button in the sidebar, but we could also just call it "Complete Setup", or "Add Integration and Complete Setup".

I think the wording simplification is much nicer now that we don't have to explain as much.

I'd argue for using the verbiage "Add Integration", but I'm open to other options.

---

Are you thinking we will release the update in another PR once this merges?

extensions/usage-metrics-dashboard/app.R
toph-allen: Yes, clearing all but a few out.

extensions/usage-metrics-dashboard/app.R
toph-allen: After today's discussion, do you think we need to mention this detail to users? This text shows all the time, and I think that if an appropriate integration exists, we don't need to show those things to the user.

I do hear what you're saying though, good to have these two words more noticeable. Other options:

- We could have the entire message be conditional, and this detail could go only in the message we show if no integration exists. I think that could be better, I'm having second thoughts about the whole wording today.
- We could bold the words where they appear on line 324, the instructions on how to create a new integration.

extensions/usage-metrics-dashboard/app.R
toph-allen: Yep, deleted it.

extensions/usage-metrics-dashboard/app.R
dotNomad: With the fully automated add flow I do think the distinction is a lot less important. On the no-integration-exists case I do think we need to be clear about what is needed.

toph-allen: I know we discussed a fully automated flow in the demo session, but I thought afterwards we landed on still liking to have an interstitial with a button press. Happy to try fully automating it tomorrow though.

toph-allen:
> > * connect.posit.it: deployment failure :(
> 
> Do we know what caused this?

@dotNomad I don't know, it's pretty mystifying to me. I feel like I need to investigate tomorrow. Going to post in `connect-dev` about it I think.

> > What happens if it's published by a Publisher? The API that we use to set the integrations does not say whether it requires Publisher or Admin credentials! [docs.posit.co/connect/api#put-/v1/content/-guid-/oauth/integrations/associations](https://docs.posit.co/connect/api/#put-/v1/content/-guid-/oauth/integrations/associations)
> 
> Looking at the codebase it looks like this API requires a Publisher or higher auth role so should be good there. We should make an internal ticket to clarify in the docs though.

@dotNomad I checked and it does work.

I had an old PR that adds permissions language to ALL endpoints in Connect's API docs. I forget why it got stalled; it's still there. I was just thinking earlier today that I should try to find time to revive and update it.

> > I think we can improve the language. "Add Required Integration" might be confusing — especially because we use the term "add integration" both to mean "associate it with this content" and "create a new one". Maybe it's fine because it mirrors the "Add Integration" button in the sidebar, but we could also just call it "Complete Setup", or "Add Integration and Complete Setup".
> 
> I think the wording simplification is much nicer now that we don't have to explain as much.
> 
> I'd argue for using the verbiage "Add Integration", but I'm open to other options.
> 
> Are you thinking we will release the update in another PR once this merges?

@dotNomad I think I want to hold release until I make another change or two, but I could also just update the manifest.json for this version — now that you mention it there is _no_ reason not to!

extensions/usage-metrics-dashboard/app.R
```diff
@@ -274,25 +275,80 @@ ui <- function(request) {
...
-          "<a href='https://docs.posit.co/connect/user/oauth-integrations/#obtaining-a-visitor-api-key' target='_blank'>User Guide</a>.",
-          sep = " "
+          "Click the button below to allow this content to use the integration <strong>'",
+          selected_integration$name,
+          "'</strong>.",
+          "<br><br>",
+          "For more information, see ",
+          "<a href='https://docs.posit.co/connect/user/oauth-integrations/#obtaining-a-visitor-api-key' ",
+          "target='_blank'>Visitor API Key section of the User Guide</a>."
+        )
```
jonkeane, line 310: Lets change this to something like this, where the name of the integration is in the button and we mention that it's compatible rather than details about it.

> This content requires a Visitor API Key integration to show users the content they have access to. A compatible integration is listed below.
> 
> For more information, see documentation on Visitor API Key integrations.
> 
> [+ Add the "Connect API - Publisher role" integration]

extensions/usage-metrics-dashboard/integrations.R
```diff
@@ -0,0 +1,53 @@
+library(connectapi)
```
jonkeane, line 1: Do we have (an) issue(s) for adding this to connectapi?

jonkeane (dismissed):
I'm approving, but we should make the text cleaner like I suggest, but if you make that change, this can merge without me re-reviewing

extensions/usage-metrics-dashboard/integrations.R
toph-allen: I'm going to make sure we do when I close this

jonkeane: If you make the issues now you could comment here saying "when this other thing is merged delete these" whcih I have always found very very helpful

toph-allen: Yep — planning to reference this in the issue(s). Making them is on my burndown of things to do for today (next, in fact!)

toph-allen: There is an epic from last year referencing them. I'm going to add sub-issues for the individual recipes to make things more bite-size (so they better fit into guild work), as the epic lists _six_ different endpoints to support.

dotNomad (approved):
Thanks for all of those changes, this is such a huge improvement.

PR #132, opened 2025-05-16
Add `category` support for Gallery extensions

dotNomad:
This PR introduces a new `category` property for the `extension` section of the `manifest.json`. The `category` can be an `id` of one of the `categories` in the `extensions.json`.

The categories are used to group the content in the Connect UI.

Items tagged Jump Starts are being migrated over to `category: "example"` since we want one of the distinct categories in the Gallery to be Examples - containing the jump start examples. The other category is "extension" which contains the useful tools/dashboards we want to share.

The changes here include:
- linting to make sure the `category` is valid (similar to how the tag linting works)
- updates to the `extension-list` script to write new `category` values to extensions in the list
- documentation in the `CONTRIBUTING` guide describing the `category` property
- updates to the manifest and extension list to use the new categories
  - this is entirely metadata so no need for new releases (similar to tags)

---

Some example GitHub workflows to test the new linting:
- [An empty `category: ""`](https://github.com/posit-dev/connect-extensions/actions/runs/15074499637/job/42378659398?pr=132)
- [An invalid `category`](https://github.com/posit-dev/connect-extensions/actions/runs/15074576409/job/42378916043?pr=132)
- [A valid `category`](https://github.com/posit-dev/connect-extensions/actions/runs/15074590881/job/42378969579?pr=132)
- [No `category`](https://github.com/posit-dev/connect-extensions/actions/runs/15074608297/job/42379022737?pr=132)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,7 @@ to the `manifest.json`:
     "title": "My Content Name",
     "description": "A lovely, detailed description of the content.",
     "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/my-content-name",
+    "category": "Extensions",
     "tags": [],
     "minimumConnectVersion": "2025.04.0",
     "version": "0.0.0"
@@ -50,6 +51,15 @@ to the `manifest.json`:
 It is recommended to begin with `"version": "0.0.0"` to avoid triggering a
 release during development of your content. When you are ready to release to
 the gallery check the [Adding content to the Connect Gallery](#adding-content-to-the-connect-gallery) section.
+
+#### Category
+
+The `category` field in the `extension` section of the `manifest.json` is
+required to group content in the Gallery.
+
+Available categories are listed in the [`extensions.json`](./extensions.json)
+file. The category should match the `title`, and automations in the repository
+will check that the category is valid.
 
 #### Tags
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ to the `manifest.json`:
     "title": "My Content Name",
     "description": "A lovely, detailed description of the content.",
     "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/my-content-name",
-    "category": "Extensions",
+    "category": "",
     "tags": [],
     "minimumConnectVersion": "2025.04.0",
     "version": "0.0.0"
@@ -57,8 +57,20 @@ the gallery check the [Adding content to the Connect Gallery](#adding-content-to
 The `category` field in the `extension` section of the `manifest.json` is
 required to group content in the Gallery.
 
+```json
+// manifest.json
+
+{
+  ...
+  "extension": {
+    "category": "extension,
+    ...
+  }
+}
+```
+
 Available categories are listed in the [`extensions.json`](./extensions.json)
-file. The category should match the `title`, and automations in the repository
+file. The category should match the `id`, and automations in the repository
 will check that the category is valid.
 
 #### Tags
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,9 @@ Available categories are listed in the [`extensions.json`](./extensions.json)
 file. The category should match the `id`, and automations in the repository
 will check that the category is valid.
 
+The order of the categories determines the order in which they are displayed
+in the Connect Gallery.
+
 #### Tags
 
 The `tags` array in the `extension` section of the `manifest.json` is optional,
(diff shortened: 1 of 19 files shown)


CONTRIBUTING.md
```diff
@@ -51,6 +52,30 @@ It is recommended to begin with `"version": "0.0.0"` to avoid triggering a
...
+  ...
+  "extension": {
+    "category": "extension,
+    ...
+  }
+}
+```
+
+Available categories are listed in the [`extensions.json`](./extensions.json)
+file. The category should match the `id`, and automations in the repository
```
cgraham-rs, line 73: It appears it should be matching on `title` not `id`

It is slightly confusing as `extensions.json` has an `id` and `title` that have very similar values with only slight differences.
```
      "id": "example",
      "title": "Examples",
```

The linting tests seem to confirm everything is keyed from `title`

The linting for [A valid category](https://github.com/posit-dev/connect-extensions/actions/runs/15074590881/job/42378969579?pr=132) passes using the `title`

```
"category": "Examples",
```

The linting for [An invalid category](https://github.com/posit-dev/connect-extensions/actions/runs/15074576409/job/42378916043?pr=132) includes logging that uses the `title`
```
Valid categories are: Examples, Extensions
```

CONTRIBUTING.md
```diff
@@ -51,6 +52,30 @@ It is recommended to begin with `"version": "0.0.0"` to avoid triggering a
...
+will check that the category is valid.
+
+The order of the categories determines the order in which they are displayed
```
cgraham-rs, line 76: Suggestion to clarify we are specifically referring to the order of categories in `extensions.json`

extensions.json
```diff
@@ -1,7 +1,17 @@
 {
-  "tags": [
-    "Jump Start"
+  "categories": [
+    {
+      "id": "extension",
```
cgraham-rs, line 4: If `id` is not actively used we should consider removing it as it's a little confusing having two fields with extremely similar values.

dotNomad: `id` is what the items in the gallery use and the UI uses the title for the category tab text.

I figured a singular `id` form would be helpful if we wanted to show the category in the UI aside from the tab.

cgraham-rs: I don't have any strong opinions. But they seem nearly identical enough where we could consider collapsing them to one key for simplicity.

dotNomad: > I don't have any strong opinions. But they seem nearly identical enough where we could consider collapsing them to one key for simplicity.

I totally understand that. The idea to split them up comes from my experience dealing with plurals and singular representation. Since the Connect UI will not know the categories itself and they are given from the `extensions.json` having both the plural and singular representation here is very helpful and gives us a good deal of flexibility in how to present them.

They are very similar now, but depending on what categories we add in the future singular and plural representations can diverge a bit (aka not just have an `s` tacked on)

CONTRIBUTING.md
cgraham-rs: This is really nit picky, but if the idea is to be able to display singular vs plural in the Connect UI wouldn't we want to maintain the capitalization.

I don't think we need to re-do the tests. That's just where I started since it was provided in the original description.

dotNomad: Totally fair. Its really easy to do capitalization using CSS so I could adjust both to be lowercase.

CONTRIBUTING.md
dotNomad: Added a specific call out in https://github.com/posit-dev/connect-extensions/pull/132/commits/b3be3fd1f973b4dda0e7e7a69f3fef2468225329

CONTRIBUTING.md
toph-allen: Oof, yeah, my string initial reaction was that we should at *least* make the `id` a lower-case version of the `title`. When I was glancing through this PR earlier today I didn't even notice that `id` was singular.

[edit]

Looking through the conversation and changes that followed, I see the conversation when in the opposite way that I had been thinking — where `id` and `title` are now both lower-case but one is singular and one is plural.

extensions.json
```diff
@@ -1,7 +1,17 @@
...
+      "title": "extensions",
+      "description": "Useful tools that utilize the Posit Connect API to extend the functionality of Connect."
+    },
+    {
+      "id": "example",
+      "title": "examples",
```
toph-allen, line 10: I think that `id` and `title` is a confusing way to name these two peer values. There's nothing in that innately that tells me how they should be used, and nothing _structurally_ that tells me which I should use in my content. I think we should remove `id` for now in the spirit of YAGNI.

In general, I'm in favor of pluralizing the names and titles of categories and tags. Even displaying the category on the page of a content item, it would look fine and correct to have a label that read `category: Examples`.

toph-allen (changes requested):
My hot take is that it'd simplify things _and_ look better to remove the `id` and just go with the `title`. I think that the plural category name looks better even in contexts where you're editing a manifest, and the presence of both the `id` and `title` adds confusion.

toph-allen (approved):
Approving so that the PR can move forward — I don't want to hold it up over semantic issues

PR #135, opened 2025-05-19
Publisher Command Center: Allow publishers to publish

### LINKED ISSUE #127: Publisher Command Center: does not work if deployed by a Publisher
The Publisher Command Center does not work if the deploying account is a Publisher.

It looks like the code is calling the unversioned, private `/metrics/procs` endpoint. We need to remove this functionality.

[rstudio-connect.80169.0.271290.run_fastapi_app.B3fpJKjSU4MN2Cy3.log](https://github.com/user-attachments/files/20214573/rstudio-connect.80169.0.271290.run_fastapi_app.B3fpJKjSU4MN2Cy3.log)

toph-allen:
Fixes #127 

The Publisher Command Center could not be published by a Publisher because it called the unversioned, admin-only `metrics/procs` endpoint. It used this to display CPU and memory usage for processes.

I removed those calls and removed their display from the UI, so now the content will run successfully with a Publisher API key.

I deployed the content to my local dev Connect environment and confirmed that it could run from a Publisher account, kill processes, etc.

I bumped the version so that it releases to the gallery. @fizzboop this might affect your open PR on the same piece of content — we can coordinate on that if we need to!

Screenshots attached showing before and after:

Screen Shot 2025-05-19 at 04 31 26 PM@2x
Screen Shot 2025-05-19 at 05 13 11 PM@2x

### DIFF
--- a/extensions/publisher-command-center/app.py
+++ b/extensions/publisher-command-center/app.py
@@ -47,14 +47,9 @@ def get_visitor_client(token: str | None) -> connect.Client:
 async def contents(posit_connect_user_session_token: str = Header(None)):
     visitor = get_visitor_client(posit_connect_user_session_token)
 
-    response = client.get("metrics/procs")
-    processes = response.json()
-
     contents = visitor.me.content.find()
     for content in contents:
-        content["processes"] = [
-            process for process in processes if content["guid"] == process["app_guid"]
-        ]
+        content["active_jobs"] = [job for job in content.jobs if job["status"] == 0]
 
     return contents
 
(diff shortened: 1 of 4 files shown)


jonkeane (approved):
I'm fine to remove this for now — it would be nice to display this information if we could get to it — are there already issue(s) for exposing that (probably at least one in Connect itself?)

toph-allen:
I'll file one

tdstein (approved):
Agreed. It would be awesome to expose these types of metrics to publishers so they can understand application performance.

toph-allen:
Issue created: https://github.com/posit-dev/connect/issues/31544

PR #136, opened 2025-05-20
Usage Metrics Dashboard: Add loading spinners to outputs

toph-allen:
Adds loading spinners to tables and plots while data is being loaded and outputs are being rendered. See attached screen recording for an example, or go play with it at https://connect.posit.it/megadashboard-loading-spinners/.

https://github.com/user-attachments/assets/ff21c27d-0472-46b5-a6d9-ce33ea79d384

### DIFF
--- a/extensions/usage-metrics-dashboard/app.R
+++ b/extensions/usage-metrics-dashboard/app.R
@@ -9,9 +9,15 @@ library(lubridate)
 library(reactable)
 library(ggplot2)
 library(plotly)
+library(shinycssloaders)
 
 shinyOptions(
   cache = cachem::cache_disk("./app_cache/cache/", max_age = 60 * 60)
+)
+
+options(
+  spinner.type = 1,
+  spinner.color = "#7494b1"
 )
 
 source("get_usage.R")
@@ -38,7 +44,7 @@ bar_chart <- function(
   value,
   max_val,
   height = "1rem",
-  fill = "#00bfc4",
+  fill = "#7494b1",
   background = NULL
 ) {
   width <- paste0(value * 100 / max_val, "%")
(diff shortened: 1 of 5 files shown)


extensions/usage-metrics-dashboard/app.R
```diff
@@ -201,7 +207,7 @@ ui <- function(request) {
     div(
       id = "multi_content_table",
       textOutput("summary_text"),
-      reactableOutput("content_usage_table")
+      withSpinner(reactableOutput("content_usage_table"))
```
jonkeane, line 210: Nice!

extensions/usage-metrics-dashboard/get_usage.R
```diff
@@ -1,5 +1,4 @@
 library(connectapi)
-library(pins)
```
jonkeane, line 2: Was this leftover cruft? Or did we change something here that I'm not realizing removes the need?

extensions/usage-metrics-dashboard/renv.lock
```diff
@@ -1934,79 +1934,6 @@
       "Maintainer": "Kirill Müller <kirill@cynkra.com>",
       "Repository": "CRAN"
     },
-    "pins": {
```
jonkeane, line 1937: I'm _a little_ surprised there isn't at least one pins dependency that we can get rid of, but maybe they are all used by other things.

jonkeane (approved):
Nice and clean, love it. A few questions mostly

extensions/usage-metrics-dashboard/get_usage.R
toph-allen: Yes, it was leftover cruft that I missed when deleting the pins usage in an old PR.

extensions/usage-metrics-dashboard/renv.lock
toph-allen: Yeah, I think `renv` only writes necessary dependencies, so (to my understanding) there isn't anything in the lockfile that isn't required by one of the packages we load. We could probably reduce our dependencies somewhat, to be sure — I kind of approached writing this as though I was a publisher, not trying to keep things parsimonious and minimal like package dev.

Issue #122, opened 2025-05-12
Publisher Command Center: Change buttons on content table rows


toph-allen:
Feedback from @kmasiello:

> clicking anywhere on the row of a content item opens the detail view. but also there's a gear icon that makes me think i'll have more controls but I don't.  from this screen i'd like to be able to delete, lock, rename, edit content settings, edit gh repo.

Possible stories

- [x] Remove the gear button from table rows
- [x] Add table row button for deleting content
- [x] Add table row button for locking content
- [x] Add table row button for renaming content

fizzboop:
Completed by #134 and #138

PR #138, opened 2025-05-20
Add lock and rename content functionality to Publisher Command Center

fizzboop:
## Intent
Addresses the last two items of the #122 

## Approach
Created two endpoints to handle locking/unlocking content and rename content.

## Example
[Publisher Command Center Test](https://dogfood.team.pct.posit.it/connect/#/apps/52cbb388-814d-4869-a8dd-b785dfd8066c/access) on DogFood

## Screenshots
image

image

### DIFF
--- a/extensions/publisher-command-center/app.py
+++ b/extensions/publisher-command-center/app.py
@@ -1,6 +1,6 @@
 from http import client
 import asyncio
-from fastapi import FastAPI, Header
+from fastapi import FastAPI, Header, Body
 from fastapi.staticfiles import StaticFiles
 from posit import connect
 from posit.connect.errors import ClientError
@@ -76,13 +76,13 @@ async def lock_content(
 @app.patch("/api/content/{content_id}/rename")
 async def rename_content(
     content_id: str,
-    new_name: str = "",
+    title: str = Body(..., embed = True),
     posit_connect_user_session_token: str = Header(None),
 ):
     visitor = get_visitor_client(posit_connect_user_session_token)
     content = visitor.content.get(content_id)
 
-    content.update(title = new_name)
+    content.update(title = title)
     return content
 
 @app.get("/api/contents/{content_id}/processes")
--- a/extensions/publisher-command-center/app.py
+++ b/extensions/publisher-command-center/app.py
@@ -61,6 +61,29 @@ async def content(
     visitor = get_visitor_client(posit_connect_user_session_token)
     return visitor.content.get(content_id)
 
+@app.patch("/api/content/{content_id}/lock")
+async def lock_content(
+    content_id: str,
+    posit_connect_user_session_token: str = Header(None),
+):
+    visitor = get_visitor_client(posit_connect_user_session_token)
+    content = visitor.content.get(content_id)
+    is_locked = content.locked
+
+    content.update(locked=not is_locked)
+    return content
+
+@app.patch("/api/content/{content_id}/rename")
+async def rename_content(
+    content_id: str,
+    new_name: str = "",
+    posit_connect_user_session_token: str = Header(None),
+):
+    visitor = get_visitor_client(posit_connect_user_session_token)
+    content = visitor.content.get(content_id)
+
+    content.update(title = new_name)
+    return content
 
 @app.get("/api/contents/{content_id}/processes")
 async def get_content_processes(
--- a/extensions/publisher-command-center/src/components/ContentsComponent.js
+++ b/extensions/publisher-command-center/src/components/ContentsComponent.js
@@ -2,8 +2,9 @@ import m from "mithril";
 import { format } from "date-fns";
 import Contents from "../models/Contents";
 import Languages from "./Languages";
-import DeleteModal from "./DeleteModal";
 import LockContentButton from "./LockContentButton";
+import DeleteModal from "./DeleteModal";
+import RenameModal from "./RenameModal";
 
 const ContentsComponent = {
   error: null,
@@ -44,6 +45,8 @@ const ContentsComponent = {
           m("th", { scope: "col" }, "Date Added"),
           m("th", { scope: "col" }, ""),
           m("th", { scope: "col" }, ""),
+          m("th", { scope: "col" }, ""),
+          m("th", { scope: "col" }, ""),
         ]),
       ),
       m(
@@ -71,8 +74,21 @@ const ContentsComponent = {
               m("td", format(content["created_time"], "MMM do, yyyy")),
               m(
                 "td",
+                m("button", {
+                  class: "action-btn",
+                  ariaLabel: `Rename ${title}`,
+                  title: `Rename ${title}`,
+                  "data-bs-toggle": "modal",
+                  "data-bs-target": `#renameModal-${guid}`,
+                }, [
+                  m("i", { class: "fa-solid fa-pencil" })
+                ]),
+              ),
+              m(
+                "td",
                 m(LockContentButton, {
                   contentId: guid,
+                  contentTitle: title,
                   isLocked: content["locked"],
                 }),
               ),
@@ -80,7 +96,8 @@ const ContentsComponent = {
                 "td",
                 m("button", {
                   class: "action-btn",
-                  ariaLabel: "Delete Content",
+                  title: `Delete ${title}`,
+                  ariaLabel: `Delete ${title}`,
                   "data-bs-toggle": "modal",
                   "data-bs-target": `#deleteModal-${guid}`,
                 }, [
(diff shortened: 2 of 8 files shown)


extensions/publisher-command-center/src/components/RenameModal.js
```diff
@@ -0,0 +1,89 @@
...
+          m("div", { class: "form-group has-validation" }, [
+            m("label", {
+              for: "rename-content-input",
+              class: "mb-3",
+            }, "Enter new name for ", [
+              m("span", { class: "fw-bold" }, `${vnode.attrs.contentTitle}`)
+            ]),
+            m("input", {
+              oninput: function(e) {
+                RenameModalForm.isValid = e.target.validity.valid;
```
fizzboop, line 33: This ensures that the value from the input has a minimum of 3 characters and a max of 1024. This is based off of a warning a user will receive on the Info Pane of the Settings Panel.

image

extensions/publisher-command-center/src/models/Contents.js
```diff
@@ -34,6 +34,29 @@ export default {
     this.data = this.data.filter((c) => c.guid !== guid);
   },
 
+  lock: async function (guid) {
+    await m.request({
+      method: "PATCH",
+      url: `api/content/${guid}/lock`,
+    }).then((response) => {
+      const targetContent = this.data.find((c) => c.guid === guid);
+      Object.assign(targetContent, response);
```
fizzboop, line 43: Since Mihtril.JS doesn't have reactivity like Vue or React, we need to change the content of the data in order to reflect the new changes to the UI via the automatic `m.redraw()`. Without this, we would need a page reload to refresh the UI.

extensions/publisher-command-center/app.py
```diff
@@ -61,6 +61,29 @@ async def content(
...
+    return content
+
+@app.patch("/api/content/{content_id}/rename")
+async def rename_content(
+    content_id: str,
+    title: str = Body(..., embed = True),
+    posit_connect_user_session_token: str = Header(None),
+):
+    visitor = get_visitor_client(posit_connect_user_session_token)
+    content = visitor.content.get(content_id)
```
dotNomad, line 83: My first thought with both of these was "do we already have the `content` data?". If we do we could avoid an API call here.

It looks like there isn't a good way, given only a GUID, to call `content.update(title = title)` from the SDK though, perhaps that is something we could request to make things like this a bit easier so this could be faster.

dotNomad: Actually there is!

We can do:

```python
    visitor = get_visitor_client(posit_connect_user_session_token)
    context = connect.Context(visitor)
    content = connect.ContentItem(context, content_id)

    content.update(title = title)
```

with the additional imports:

```
from posit.connect.context import Context
from posit.connect.content import ContentItem
```

We can make a `Context` for the `ContentItem` and pass the `guid`. Since `ContentItem` only needs a `Context` and `guid` and all other attributes are optional this is enough to do the rename.

The lock has `is_locked = content.locked` which relies on the Content's `locked` attribute. If we can pass that in the FastAPI `/api/content/{content_id}/lock` API (or create two one for unlock and one for lock) then we can do the same there speeding it up by 2 times.

---

I don't think any of this is necessary in this PR to be fair. If it is very easy we can get it in otherwise we can follow-up since it would be good deal faster.

extensions/publisher-command-center/src/components/ContentsComponent.js
```diff
@@ -72,6 +76,28 @@ const ContentsComponent = {
...
+                }, [
+                  m("i", { class: "fa-solid fa-pencil" })
+                ]),
+              ),
+              m(
+                "td",
+                m(LockContentButton, {
+                  contentId: guid,
+                  contentTitle: title,
+                  isLocked: content["locked"],
```
dotNomad, line 92: Oh sweet we have `isLocked` for the component, we can use that in `Contents.lock` and send it up to the API (or have two endpoints)

extensions/publisher-command-center/src/components/ContentsComponent.js
```diff
@@ -83,6 +109,7 @@ const ContentsComponent = {
                 m("a", {
                   class: "fa-solid fa-arrow-up-right-from-square",
                   href: content["content_url"],
+                  title: `Open ${title} (opens in new tab)`,
```
dotNomad, line 112:
```suggestion
                  title: `Open ${title}`,
```
The "(opens in a new tab)" feels a bit verbose to me, but I'm not that adamant about removing it.

extensions/publisher-command-center/src/components/LockContentButton.js
```diff
@@ -0,0 +1,23 @@
+import m from "mithril";
+import Contents from "../models/Contents";
+
+const LockedContentButton = {
```
dotNomad, line 4: I found this button a bit tricky on first use and accidentally clicked it twice - partly because the API was slow since we do two Connect calls, but also because there wasn't any feedback.

I did a quick edit of the component that may be a good starting point for adjust this

<details>
  <summary>Code</summary>
  
```js
import m from "mithril";
import Contents from "../models/Contents";

const LockedContentButton = {
  oninit: function () {
    this.isLoading = false;
  },

  view: function (vnode) {
    const labelMessage = vnode.attrs.isLocked ?
      `Unlock ${vnode.attrs.contentTitle}` :
      `Lock Content ${vnode.attrs.contentTitle}`;
    const iconClassName = vnode.attrs.isLocked ? "fa-lock" : "fa-lock-open";

    return m("button", {
      class: "action-btn" + (this.isLoading ? " disabled" : ""),
      ariaLabel: labelMessage,
      title: labelMessage,
      disabled: this.isLoading,
      onclick: async () => {
        if (this.isLoading) return;

        this.isLoading = true;
        m.redraw();

        try {
          await Contents.lock(vnode.attrs.contentId);
        } finally {
          this.isLoading = false;
          m.redraw();
        }
      }
    }, [
      m("i", {
        class: `fa-solid ${this.isLoading ? "fa-spinner fa-spin" : iconClassName}`
      })
    ])
  }
};

export default LockedContentButton;

```
</details> 

This changes the button to a disabled spinner while the API call is going. This made is much easier to tell what was going on and avoid a double lock call.

extensions/publisher-command-center/src/components/RenameModal.js
```diff
@@ -0,0 +1,93 @@
...
+  onsubmit: function(e) {
+    if (RenameModalForm.isValid) {
+      e.preventDefault();  
+      Contents.rename(RenameModalForm.guid, RenameModalForm.newName);
+
+      const modalEl = document.getElementById(`renameModal-${RenameModalForm.guid}`);
+      const modal = Modal.getInstance(modalEl);
+      modal.hide();
+
+      RenameModalForm.newName = "";
```
dotNomad, line 18: Good catch!

extensions/publisher-command-center/src/models/Contents.js
dotNomad: Good call adjusting the data so we don't need the slower API call.

dotNomad (commented):
Both #146 and this PR are seeking to upgrade Publisher Command Center. Lets get both merged, then do a version bump

extensions/publisher-command-center/app.py
tdstein: Yeah, that looks good @dotNomad! For what it's worth, I wouldn't worry about the additional API call to retrieve the content in order to update it. It's all happening across the local network, so the overhead is fairly negligible.

extensions/publisher-command-center/src/components/ContentsComponent.js
fizzboop: I can move the "(opens in new tab)" portion to the `aria-label` instead of the tooltip. In general, if a link opens a new window or tab, the screen reader should alert the user of the action.

extensions/publisher-command-center/app.py
dotNomad: > For what it's worth, I wouldn't worry about the additional API call to retrieve the content in order to update it. It's all happening across the local network, so the overhead is fairly negligible.

Oh that is a good point - it will be much faster than local dev 👍

extensions/publisher-command-center/src/components/ContentsComponent.js
dotNomad: > In general, if a link opens a new window or tab, the screen reader should alert the user of the action.

Do screen readers not give that information when `target="_blank"`? I guess they don't

fizzboop: Added `aria-label` in commit [3e93e62](https://github.com/posit-dev/connect-extensions/pull/138/commits/3e93e62329ed387b387cfad8473152180f5fe1c9).

dotNomad (approved):
Looks fantastic thank you for those changes 🎉

PR #150, opened 2025-05-22
Bumps version for Usage Metrics Dashboard and Publisher Command Center

fizzboop:


extensions/usage-metrics-dashboard/manifest.json
dotNomad, line 1: I'm surprised the checksums on the files didn't change here given the changes in `app.R`. Can we check that again?

dotNomad (approved):
Looks great lets get these in 🎉

PR #155, opened 2025-05-29
Loosen `usage-metrics-dashboard` R requirement

dotNomad:
This releases a new version of the Usage Metrics Dashboard that loosens the R environment requirement. Instead of requiring `~=4.3.0` (`4.3.x`) this allows for `4.x` R versions.


toph-allen (approved):
Let's give this a shot!

PR #157, opened 2025-05-29
docs: add directions to use p3m to `CONTRIBUTING.md`

toph-allen:
Adds directions to use the Posit Public Package Manager to `CONTRIBUTING.md`.


dotNomad (approved):
Fantastic addition. Really like the note on _why_ we do this as well.

PR #173, opened 2025-06-11
runtime version scanner: improve table display and filter controls

toph-allen:
This PR contains a number of improvements to the runtime version scanner:

- Improved table display
  - Added GUID column
  - Improved styling and spacing of columns 
  - Added pagination back
  - (Checked version sorting; didn't appear to be broken in the current version)
- Updated filter controls
  - Changed version selector from a slider to a drop-down
  - Changed matching from <= to <
  - Added filtering based on hits (Considered adding to table, but added in sidebar so that it could affect a forthcoming "Export" button)
- Started to work with data on what runtimes are on the server
  - Added a control to highlight content with stale builds (i.e. it was last built against a runtime that no longer exists)
- Misc
  - Updated CSS

With the version selector control — I had wanted to do a "type in the version number" control, but I didn't have time to implement that, as it would have required significant refactoring of how the app handles versions. It might be worth postponing to a future iteration.

### DIFF
--- a/extensions/eol-runtime-scanner/app.R
+++ b/extensions/eol-runtime-scanner/app.R
@@ -129,7 +129,7 @@ ui <- page_sidebar(
 
     numericInput(
       "min_hits_filter",
-      label = "Minimum Hits",
+      label = "Minimum Views",
       value = 0,
       min = 0,
       step = 1
@@ -139,7 +139,7 @@ ui <- page_sidebar(
 
     checkboxInput(
       "show_guid",
-      label = "Show GUID"
+      label = "Show GUIDs"
     ),
 
     checkboxInput(
@@ -419,7 +419,7 @@ server <- function(input, output, session) {
         ),
 
         hits = colDef(
-          name = "Hits (1 Week)",
+          name = "Views (Last Week)",
           width = 125,
           class = "number-pre",
         )
--- a/extensions/eol-runtime-scanner/app.R
+++ b/extensions/eol-runtime-scanner/app.R
@@ -78,46 +78,43 @@ ui <- page_sidebar(
 
     checkboxInput(
       "use_r_cutoff",
-      "Match R versions…",
+      label = "R older than…",
       value = FALSE
     ),
     conditionalPanel(
       condition = "input.use_r_cutoff == true",
-      sliderTextInput(
-        "min_r_version",
+      selectizeInput(
+        "r_version_cutoff",
         label = NULL,
-        choices = "...",
-        pre = "≤ "
+        choices = NULL
       )
     ),
 
     checkboxInput(
       "use_py_cutoff",
-      "Match Python versions…",
+      label = "Python older than…",
       value = FALSE
     ),
     conditionalPanel(
       condition = "input.use_py_cutoff == true",
-      sliderTextInput(
-        "min_py_version",
+      selectizeInput(
+        "py_version_cutoff",
         label = NULL,
-        choices = "...",
-        pre = "≤ "
+        choices = NULL
       )
     ),
 
     checkboxInput(
       "use_quarto_cutoff",
-      "Match Quarto versions…",
+      label = "Quarto older than…",
       value = FALSE
     ),
     conditionalPanel(
       condition = "input.use_quarto_cutoff == true",
-      sliderTextInput(
-        "min_quarto_version",
+      selectizeInput(
+        "quarto_version_cutoff",
         label = NULL,
-        choices = "...",
-        pre = "≤ "
+        choices = NULL
       )
     ),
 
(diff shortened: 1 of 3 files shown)


extensions/eol-runtime-scanner/app.R
```diff
@@ -40,6 +39,27 @@ app_mode_lookup <- with(
...
+      # Check if version doesn't exist in server versions
+      if (!(value_str %in% server_versions)) {
+        return(list(
+          color = "#7D1A03", # Red color for versions not on server
+          fontWeight = "bold",
+          background = "#FFF0F0" # Light red background
+        ))
+      } else {
+        return(list(
+          color = "#276749" # Green color for versions on server
```
dotNomad, line 56: Nice job on the colors here 👍 They are accessible which is tough for color-on-color text.

extensions/eol-runtime-scanner/app.R
```diff
@@ -57,46 +77,43 @@ ui <- page_sidebar(
...
-      sliderTextInput(
-        "min_r_version",
+      selectizeInput(
+        "r_version_cutoff",
         label = NULL,
-        choices = "...",
-        pre = "≤ "
+        choices = NULL
       )
     ),
```
dotNomad, line 90: I noticed the `select` boxes here didn't have labels. They are not shown which makes sense, but they aren't accessible due to this change since there is no content in the `<label>`. If we could supply a label and hide it with some CSS like the below that would be great.

```
.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border-width: 0;
}
```

dotNomad (approved):
The code here looks great. I highlighted a couple bits for feedback, but totally fine with any of those being follow-up PRs 😄

extensions/eol-runtime-scanner/app.R
toph-allen: I actually think that highlighting in color is perhaps a feature which will go away (it doesn't really communicate something easily understandable) but thanks 😂

toph-allen:
@jonkeane Thanks for your feedback here! I'll go point by point.

> * I actually like the dropdowns for selecting versions better than I thought I would. I know I started off saying sliders, but the dropdowns aren't bad and it's nice to _see_ the version possibilities. One thing I would like is stronger communication to me as to what "R older than 4.3.3" means I know it's clear when I sit and think about it, but is there some way that we can communicate that slightly more obviously at a glance? It would especially nice if we communicated both which items were in the filter (and therefor are possibly in the category of "need to do something about" versus which are outside of the filter and are possibly int he category of "just fine"). A timeline showing the versions from 3.1.0 to 4.3.3 are red, and the versions 4.3.3 to 4.5.0 are green? Maybe a histogram of versions where the versions that are less than and the versions that are greater than are different colors?

I agree! I had been thinking about some kind of timeline view — it might be fiddly to code but worth it.

> * I like the count in "Showing 15 items with any of:" and "R older than 4.4.0 (15 items)" to show me how many items I have. Is it possible to also add in the total so I know if 15 is a lot? Like is it 15 of 17 items? Or 15 of 1750?
> * It's a little bit silly that there are two of these, but I get that you might have a mixture of languages — it would be _really slick (though not required)_ to have that text transform if there are multiple filters, otherwise be simpler. Where the total matched only shows up when you have multiple filters. Having multiple filters is not a super likely scenario (but it's good that we are supporting it).

Yep, I like these two pieces of feedback.

> * I really want the table to be sorted by the version I'm filtering on by default. Anyway we can make that happen?

Possibly! It might be a bit fiddly I'm not sure what to do if there are multiple runtimes selected for the filter, but perhaps we only do it when a single one is selected.

> * What does `Highlight stale builds` mean?

It means "Highlight items whose most recent environment build used a version of a runtime that doesn't exist on the server any more."

Communicating better about what this means feels related to better communicating about server versions and the currently selected versions. In the next iteration I'm focusing on different ways to communicate the versions — perhaps a timeline or a histogram.

jonkeane:
> It means "Highlight items whose most recent environment build used a version of a runtime that doesn't exist on the server any more."
> 
> Communicating better about what this means feels related to better communicating about server versions and the currently selected versions. In the next iteration I'm focusing on different ways to communicate the versions — perhaps a timeline or a histogram.

Can you punt on this right now? 🚮 the message + code etc. And come back to it when the other stuff is in a better place? That will help you stay focused on getting _something useful_ in this iteration. Like we talked about in our 1:1: this is an edge case, it would be nice to message about it, but this is not the most important thing / first thing we should be polishing.

toph-allen:
> > It means "Highlight items whose most recent environment build used a version of a runtime that doesn't exist on the server any more."
> > Communicating better about what this means feels related to better communicating about server versions and the currently selected versions. In the next iteration I'm focusing on different ways to communicate the versions — perhaps a timeline or a histogram.
> 
> Can you punt on this right now? 🚮 the message + code etc. And come back to it when the other stuff is in a better place? That will help you stay focused on getting _something useful_ in this iteration. Like we talked about in our 1:1: this is an edge case, it would be nice to message about it, but this is not the most important thing / first thing we should be polishing.

Yeah, definitely. I agree, I think that this is not helpful as a core functionality, the current messaging is more confusing than it is helpful, and trying to focus on it or whip it into shape this iteration will distract from the core functionality.

PR #174, opened 2025-06-12
Add proof of concept publisher content CVE dashboard

dotNomad:
This PR adds the proof of concept of a dashboard that enumerates the CVEs (common vulnerabilities and exposures) for content that the publisher has published.

The vulnerabilities are found using [Posit Package Manager](https://packagemanager.posit.co/) specifically the [`/repos/{repo}/vulns` API](https://packagemanager.posit.co/__api__/swagger/index.html#/default/get_repos__repo__vulns). Then the app uses Connect's [`v1/content/ endpoint`](https://docs.posit.co/connect/api/#get-/v1/content) to grab content, and finally [`v1/content/{guid}/packages`](https://docs.posit.co/connect/api/#get-/v1/content/-guid-/packages) to see packages utilized by that content cross referencing the vulnerabilities found from PPM.

<details>
  <summary>Preview</summary>
  
CleanShot 2025-06-11 at 17 15 25@2x

CleanShot 2025-06-11 at 17 16 53@2x

</details> 

The number of vulnerabilities are shown, details on each, links to https://osv.dev/, and the version to upgrade to fix the issue.

Vulnerabilities are grouped by package so there isn't repeated "To fix these vulnerabilities, upgrade to..." messages with conflicting version numbers. The latest is shown.

## Approach

I started with the same approach as the Publisher Command Center extension - a FastAPI project that served a Vite-built frontend. I used Vue for it which was a bit different.

Then I setup the API to call Connect and Posit Package Manager. Grabbing the data, storing in Pinia stores, then creating components to reference the data.

I used Tailwind for the styling.

## Review Assistance

Apologies for the size on this one. It got away from me a bit since it was so easy to prototype using Claude Code. I got setup using Vite's `vue-ts` template which is where several of these files came from

I've included several comments in the file diffs to help with reviewing.

Several of the files are either to setup Prettier, Vue, Vite, etc. The main files to investigate are:
- `main.py` this is the FastAPI that calls Connect using [`posit-sdk`](https://github.com/posit-dev/posit-sdk-py) and [Package Manager APIs](https://packagemanager.posit.co/__api__/swagger/index.html)
- The Vue files in `src`. These are the main components and stores (using Pinia). They contain most of the functionality.
- The `manifest.json` of course

Some files that can be mostly ignored for the purpose of the review (setup by the Vite template) are:
- tsconfigs
- `vite-env.d.ts`

There is a `CONTRIBUTING.md` doc to look at if you'd like to run this locally.


extensions/publisher-cves/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// https://vite.dev/config/
+export default defineConfig({
+  base: "",
+  plugins: [vue(), tailwindcss()],
+  server: {
+    proxy: {
+      "/api": {
+        target: "http://localhost:8000",
+        changeOrigin: true,
```
dotNomad, line 13: Mostly setup from Vite's templating other than this which is to develop locally alongside the FastAPI server.

extensions/publisher-cves/src/components/icons/ArrowTopRight.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+  <svg
```
dotNomad, line 2: This is this SVG:
CleanShot 2025-06-11 at 17 30 49@2x

extensions/publisher-cves/src/components/VulnerabilityChecker.vue
```diff
@@ -0,0 +1,272 @@
...
+  return null;
+}
+
+// Go back to content list
+function goBack() {
+  contentStore.currentContentId = undefined;
+}
+
+// Find vulnerable packages by comparing package data with vulnerability data
+const vulnerablePackages = computed<VulnerablePackageItem[]>(() => {
```
dotNomad, line 92: This collection of vulnerable packages is more complex than I'd like it to be. I figured I could create a follow-up to come back to this, move some of the logic into the store, and refactor it a bit.

I'm happy to address feedback on this PR, but it felt like a good follow-up to me as I was digging into it.

extensions/publisher-cves/src/components/ContentList.vue
```diff
@@ -0,0 +1,77 @@
...
+import { usePackagesStore } from "../stores/packages";
+import { useContentStore } from "../stores/content";
+import StatusMessage from "./ui/StatusMessage.vue";
+import ContentCard from "./ContentCard.vue";
+
+const packagesStore = usePackagesStore();
+const contentStore = useContentStore();
+
+// Fetch packages in batches to avoid overwhelming the server
+async function fetchPackagesInBatches(batchSize = 3) {
```
dotNomad, line 11: I debated firing off all of the fetches to grab Connect content packages, but figured batching it didn't cause many problems. It starts with the top content item and slowly loads in which is nice from the user perspective.

extensions/publisher-cves/src/stores/vulns.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const useVulnsStore = defineStore("vulns", () => {
+  // State
+  const pypi = ref<VulnerabilityMap>({});
+  const cran = ref<VulnerabilityMap>({});
+  const isLoading = ref(false);
+  const error = ref<Error | null>(null);
+
+  // Track fetch status
+  const isFetched = ref(false);
```
dotNomad, line 32: A lot of the stores have this `isFetched`, or similar, variable to avoid calling APIs multiple times. Each is only called once right now.

extensions/publisher-cves/.prettierrc
dotNomad, line 1: Turns out an empty `.prettierrc` prevents some defaults from the VS Code extension

.github/workflows/publisher-cves.yml
dotNomad, line 1: This workflow is mostly copied from `.github/workflows/publisher-command-center.yml` for comparison.

extensions/publisher-cves/src/components/VulnerabilityChecker.vue
```diff
@@ -0,0 +1,272 @@
...
+  }, 0);
+});
+</script>
+
+<template>
+  <!-- Only show the component if we have content to analyze -->
+  <div
+    v-if="hasPackages || isLoadingPackages"
+    class="max-w-4xl mx-auto p-5 bg-white rounded-lg shadow-md"
+  >
```
toph-allen, line 213: Perhaps this is why things without packages don't show anything when you click on them 🤔

extensions/publisher-cves/src/components/vulnerability/VulnerabilityDetails.vue
```diff
@@ -0,0 +1,55 @@
...
+      </p>
+    </div>
+
+    <p class="text-base mb-3">
+      <strong>{{ summary }}</strong>
+    </p>
+    <p
+      class="text-sm text-gray-600 whitespace-pre-line break-words leading-relaxed bg-gray-50 p-3 rounded-md my-3"
+    >
+      {{ details }}
```
fizzboop, line 47: Maybe a nice-to-have in the future: We use the `markdown-it` [package](https://github.com/markdown-it/markdown-it#readme) in Connect. Could we introduce it to this extension to render the markdown of the details of the vulnerability.

jonkeane:
This is looking really nice! I've been poking at actually using this extension (not yet read through the code in detail). Some of this I'm sure you've gotten from other folks offering feedback, but nevertheless my impressions:

* At the top it says `Your Connect Content` could you add in the API calls needed to pull the (first) publisher's name in there (I'm going to use "the publisher" throughout to refer to the first publisher — that is the person whose API key is being auto provisioned and used). It's a little silly (why show someone their own name? They know it!) but that would be one way to (more) explicitly show that this is all and only about the packages that _this one person_ has access to.
* Using `v1/content/` will return all items that the publisher has at least view access to. Which will include items the publisher has only view permissions on. Is that what we want here? Would we rather list all and only the items the publisher can actually take action on?
* You've already got this feedback elsewhere, but being able to sort the cars would be really lovely: by number of vulnerabilities, but also by [one simple metric like the runtime scanner has](https://github.com/posit-dev/connect-extensions/blob/main/extensions/eol-runtime-scanner/get_usage.R) so that folks know "is this an app that I _need_ to remediate, or maybe I should just delete it"

.github/workflows/publisher-cves.yml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          extension-name: ${{ env.EXTENSION_NAME }}
+          artifact-name: ${{ env.EXTENSION_NAME }}
+
+  connect-integration-tests:
+    needs: extension
+    uses: ./.github/workflows/connect-integration-tests.yml
+    secrets: inherit
+    with:
+      extensions: '["publisher-cves"]'  # JSON array format to match the workflow input schema
```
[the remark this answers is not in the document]
dotNomad: Great question. The environment isn't available in the `jobs.<job_id>.with` block.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idwith

extensions/publisher-cves/src/components/vulnerability/VulnerabilityDetails.vue
dotNomad: This was listed in some feedback we got from other teams too. I'll add it as a follow-up!

extensions/publisher-cves/src/components/VulnerabilityChecker.vue
dotNomad: Definitely part of it. I'll have to add some fallback behavior if there aren't any packages.

jonkeane:
> This uses the [posit-sdk](https://github.com/posit-dev/posit-sdk-py)'s client.me.content.find() which is all content the publisher has access to where owner_guid=self["guid"] so only content they can access where they are the owner. Is that closer to how you are expecting this to behave?

Aaah ok ok that's good. I thought you were just grabbing content from the endpoint without then filtering it, but yeah if you're using that posit-sdk funciton it will do what we want there

PR #178, opened 2025-06-12
App Canary: display GUIDs and better instructions

cgraham-rs:
Minor, iterative improvements:

- Display the CANARY_GUIDS for the user to enable them to see the value and copy-paste into the env vars for editing purposes
- Use the [posit-sdk](https://github.com/posit-dev/posit-sdk-py) instead of `requests` where possible
- Reformat the display of the setup instructions to be less error-y

### DIFF
--- a/extensions/app-canary/.gitignore
+++ b/extensions/app-canary/.gitignore
@@ -11,3 +11,5 @@ app-canary.html
 preview.html
 
 /.posit/
+
+.env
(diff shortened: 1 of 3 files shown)


toph-allen (approved):
I think I have reviewed later PRs than this one — I guess this implicitly means I approve this one?

Even disregarding that, I explicitly approve this pull request, all of the iterative improvements seem to be good ones. I think the fact that there are newer PRs means that I should avoid giving UI critique specific to this PR, but @cgraham-rs let me know if I'm misunderstanding how things are organized here.

PR #185, opened 2025-06-18
Usage Metrics Dashboard: Persist search string across filter changes

### LINKED ISSUE #153: Usage: "search" box doesn't persist when I change filters
In the main view:

* Type something in the box that says "Search by title, URL, GUID, or owner" (aside: it's not search)
* Select something in the "Filter by Content Type" dropdown on the left

When the view responds to that, the "search" term is gone. I would expect it to still be applied.

toph-allen:
Fixes #153

This PR uses some JavaScript to store the state of the search string in session storage and restore it after something triggers a `reactable` rerender.

This is definitely a bit of a workaround — but I tried a few other approaches (including using `updateReactable(data = new_data)`) and none of them worked simply. Perhaps it could be made more seamless if the `setTimeout()` value was lower. Definitely would appreciate a look from someone with more JS experience than me!

Available:
- https://dogfood.team.pct.posit.it/usage-metrics-dashboard-auto-add/
- https://connect.posit.it/usage-metrics-dashboard-2025-06-05/

### DIFF
--- a/extensions/usage-metrics-dashboard/app.R
+++ b/extensions/usage-metrics-dashboard/app.R
@@ -60,8 +60,9 @@ full_url <- function(session) {
     session$clientData$url_protocol,
     "//",
     session$clientData$url_hostname,
-    if (nzchar(session$clientData$url_port))
-      paste0(":", session$clientData$url_port),
+    if (nzchar(session$clientData$url_port)) {
+      paste0(":", session$clientData$url_port)
+    },
     session$clientData$url_pathname
   )
 }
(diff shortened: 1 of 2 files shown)


extensions/usage-metrics-dashboard/app.R
```diff
@@ -746,10 +747,41 @@ server <- function(input, output, session) {
...
   })
 
+  # JavaScript for persisting search terms across table rerenders
+  table_js <- "
+  function(el, x) {
+    const tableId = el.id;
+
+    // Initialize global store for search terms if needed
+    if (!window.currentSessionSearches) {
+      window.currentSessionSearches = {};
```
dotNomad, line 757: Rather than directly manipulating a window attribute you could use the `localStorage` API https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

dotNomad (commented):
> I suspect that the real answer for much of the awkwardness of getting this working is actually coming from the fact that we aren't using Connect's search API and returning items based on that in the table. Rather than always taking all of the items and then filtering them using reactable

Perhaps the answer for a full fledged version of this is to add search to `connectapi` so we can have that reactive element. I'm not deeply familiar with `reactable`, but I image we could call a function that searches for content with the given filters we have selected.

The flicker is a bit distracting, but I'm not sure if it should stop us from getting this temporary feature in. Its pretty handy to not have to retype a search. If we think its worth merging while we work on the search functionality then I added a suggestion.

extensions/usage-metrics-dashboard/app.R
toph-allen: I used `sessionStorage` :)

dotNomad: Smart I'll take a look at the new implementation.

dotNomad (approved):
This is working great!

PR #190, opened 2025-06-20
Chat with Content Extension V1

### LINKED ISSUE #186: Extension Proposal: Chat with content
## Purpose

There are numerous ideas pertaining to a chat assistant within Connect that could provide features like: chat with content, chat with docs, chat with Connect API, chat with Connect UI, chat with content logs, etc. The common factor among these is a context-aware chat assistant that acts as the medium to enable these things. This proposal would focus on the “chat with content” use case as a starting point with a minimally invasive design to introduce the concept.

## Diagram

Image

## Deliverables

* Shiny Chat App  
  * Content selector for one or more pieces of rendered content  
  * Shiny Chat UI backed by Chatlas  
    * Tools registers to interact with Connect API using visitor api key  
  * Iframe to load currently selected content  
* Stretch  
  * Basic citation highlighting

mconflitti-pbc:
Fixes #186 

~~Scoped to only public rendered content to avoid ACL shenanigans.~~

Setting up env vars for testing:
Run:
```
$> aws sso login
$> aws configure export-credentials
{
  "Version": 1,
  "AccessKeyId": "<access key>",
  "SecretAccessKey": "<secret key>",
  "SessionToken": "<session token>",
  "Expiration": "2025-07-02T02:37:29+00:00"
}
```
Then set the following env vars as follows:
```
CHATLAS_CHAT_PROVIDER: "bedrock-anthropic"
CHATLAS_CHAT_ARGS: '{"model": "us.anthropic.claude-sonnet-4-20250514-v1:0", "aws_region": "us-east-1", "aws_access_key": "<access key>", "aws_secret_key": "<secret key>", "aws_session_token": "<session token>"}'
```

- Setup screen to walk user through LLM and visitor api key setup
- Pulls all static content a user has access to
- Dropdown is selectize to be able to search for content
  - Can see title/name, owner, and human readable last deployed time

Once content is selected, a prompt is run behind the scenes to generate a summary and up to 3 answerable prompts to help the user interact with their content.

Demo ([dogfood](https://dogfood.team.pct.posit.it/connect/#/apps/e2ae61f0-e788-4e50-be67-fe96c2e15e2c/access)):
Screenshot 2025-06-26 at 11 33 47 AM

Setup:
Screenshot 2025-06-26 at 12 15 41 PM


jonkeane:
Thanks for this! Do you have this up on a connect instance somewhere already? I would love to poke at it myself.

> Scoped to only public rendered content to avoid ACL shenanigans.

Could you explain a bit more what you mean here?

mconflitti-pbc:
> Thanks for this! Do you have this up on a connect instance somewhere already? I would love to poke at it myself.
> 
> > Scoped to only public rendered content to avoid ACL shenanigans.
> 
> Could you explain a bit more what you mean here?

I can't scope it to the visitor since we need use the AWS integration aka we need multiple integrations :)

jonkeane:
> I am not convinced that this is very useful for the person that published the report. The goal of the extension is to show it is possible to chat with a report in general, but the target persona imo would be someone that did not publish the report trying to extract insights from it.

Right, exactly. But I think my suggestion was poorly worded cause this is flipped: what I'm suggesting is that you use the one integration slot you have now for the Connect visitor integration _so that you can chat with not-your-reports_, and guide folks through setting up their LLM with envvars (for now, until we can have multiple integrations).

jonkeane:
It's possible that the AWS setup won't work with something like an API key in an envvar, that's ok, maybe we use a different service that would let us set it up in this way?

jonkeane:
> how to make that as easy as possible for a user to attach something that works. May need some sort of cascade of checks. If we rely on env vars for chat and use that header value then no integration is needed at all which is even simpler.

Yeah, totally. You might be interested in some of what @toph did in https://github.com/posit-dev/connect-extensions/tree/main/extensions/publisher-command-center and the user metrics dashboard for doing something like this (though for both of those that's setting the visitor integration, we should do that _as well as_ stepping folks through setting the env var, set it and the app restarts working)

jonkeane:
This is a really nice demo and love the direction. I think you're already working on this, but you could add to the sidebar a place to put the API key / refresh the token and have it call the API, set the envvar, and that will relaunch the process. Which should make for a slightly smoother demonstration (before we get the ultimate multiple integrations).

One small thing I noticed: if the html is too large (one of the deployed conenctwidgets examples did this), I would get `Error in Chat('chat'): Error code: 400 - {'message': 'Input is too long for requested model.'}`

mconflitti-pbc:
> This is a really nice demo and love the direction. I think you're already working on this, but you could add to the sidebar a place to put the API key / refresh the token and have it call the API, set the envvar, and that will relaunch the process. Which should make for a slightly smoother demonstration (before we get the ultimate multiple integrations).
> 
> One small thing I noticed: if the html is too large (one of the deployed conenctwidgets examples did this), I would get `Error in Chat('chat'): Error code: 400 - {'message': 'Input is too long for requested model.'}`

Oh! I didn't know we could do that through the API, makes sense! I may add that to the setup screen then as well with a textarea containing boilerplate json for someone to edit. For things like bedrock there are more fields than just the api key unfortunately.

mconflitti-pbc:
> One small thing I noticed: if the html is too large (one of the deployed conenctwidgets examples did this), I would get Error in Chat('chat'): Error code: 400 - {'message': 'Input is too long for requested model.'}

This will be difficult to address since it is highly dependent on the model being used. Cant try to add better handling

mconflitti-pbc:
> Yup, totally can. And IIRC, you "accidentally" also get a content restart for free when you do so you don't need to do that either.

I am going to leave this as-is where the user is instructed in the setup screen to set their env var for the app. That should be sufficient and I dont like the UX when updating the env vars and the app dies without Connect refreshing the page automatically. Seems cleaner to prohibit entry to the app until all pre-reqs are set.

extensions/chat-with-content/manifest.json
```diff
@@ -0,0 +1,55 @@
...
+  "extension": {
+    "name": "chat-with-content",
+    "title": "Chat with Content",
+    "description": "Provides a way to interact with and query content using a chat interface.",
+    "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/chat-with-content",
+    "category": "extension",
+    "tags": ["llm", "shiny", "chat", "python"],
+    "minimumConnectVersion": "2025.04.0",
+    "requiredFeatures": ["OAuth Integrations"],
+    "version": "0.0.1"
```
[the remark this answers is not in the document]
mconflitti-pbc: thank you will do!

extensions/chat-with-content/manifest.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "environment": {
+    "python": {
+      "requires": ">=3.10.0"
+    }
+  },
+  "extension": {
+    "name": "chat-with-content",
+    "title": "Chat with Content",
+    "description": "Provides a way to interact with and query content using a chat interface.",
```
dotNomad, line 24: Should the description or title contain anything about LLMs? There is the "LLM" tag but perhaps something like:


```suggestion
    "title": "Chat with Content",
    "description": "Provides a way to interact with and query content using an LLM chat interface.",
```

extensions/chat-with-content/app.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Update iframe when content selection changes
+    @reactive.Effect
+    @reactive.event(input.content_selection)
+    async def _():
+        if input.content_selection() and input.content_selection() != "":
+            content = client.content.get(input.content_selection())
+            print(
+                f"Selected content: {content.last_deployed_time} {content.title} {content.name} ({content.app_mode}) - {content.owner.first_name} {content.owner.last_name} => {content.content_url}"
+            )
```
dotNomad, line 259: Should we leave the print line in? I'm not seeing it in the logs or anything, so perhaps it can stay?

dotNomad: Actually I am seeing Console logs

> Wed Jul 02 2025 12:28:35 GMT-0700 (Pacific Daylight Time) [DBG]: 2 message(s) discarded from buffer

Not sure if those are related.

extensions/chat-with-content/app.py
```diff
@@ -0,0 +1,304 @@
...
+        });
+    """),
+    fillable=True,
+)
+
+screen_ui = ui.page_output("screen")
+
+CHATLAS_CHAT_PROVIDER = os.getenv("CHATLAS_CHAT_PROVIDER")
+CHATLAS_CHAT_ARGS = os.getenv("CHATLAS_CHAT_ARGS")
+IS_INTERNAL = "connect.posit.it" in os.getenv("CONNECT_SERVER", "")
```
dotNomad, line 179: I wonder if we should be making the distinction on what chat provider to use on something else - something we can tackle in the future surely.

dotNomad (approved):
It looks like the README has not contents. It doesn't need a lot of detail, but a least a header with the title and the description of the app would be good if the GitHub link on the Gallery card is pressed.

---

Approving with a few comments. These are either minor adjustments or things I'm fine taking a look at for future releases.

extensions/chat-with-content/app.py
mconflitti-pbc: this was for debugging. can remove it!

extensions/chat-with-content/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+    **Example for Google Gemini:**
+
+    -   `CHATLAS_CHAT_PROVIDER`: `google`
+    -   `CHATLAS_CHAT_ARGS`: `{"model": "gemini-1.5-flash"}`
+    -   `GOOGLE_API_KEY`: `<your-google-api-key>` (Set this as a secret)
+
+    **Example for Anthropic on AWS Bedrock:**
+
+    If the Connect server is running on an EC2 instance with an IAM role that grants access to Bedrock, no environment variables are needed. The application will automatically detect and use AWS credentials. It defaults to the `us.anthropic.claude-sonnet-4-20250514-v1:0` model.
```
mconflitti-pbc, line 42:
```suggestion
    If the Connect server is running on an EC2 instance with an IAM role that grants access to Bedrock, no environment variables are needed. The application will automatically detect and use AWS credentials. It defaults to the `us.anthropic.claude-sonnet-4-20250514-v1:0` model. Otherwise, you can set the following environment variables with your AWS credentials:
```

PR #196, opened 2025-06-20
Runtime Version Scanner: Version Selector Improvements

toph-allen:
This PR includes:

- A totally custom UI element for selecting versions. Versions are displayed in a row of buttons. Click a button to display content using that version and below. Selected versions are highlighted.
  - Server versions are included in the list of possible versions regardless of whether the user has content using them.
  - In the selector, versions are displayed in regular typeface; versions of runtimes no longer on the server are italic.
- Removed "Highlight stale builds" feature as it was just confusing.
- Improved summary text display when there is only one runtime selected for filtering.

Not implemented yet, but possibly planned for early next week:

- I think it might be helpful to show a parenthetical count of number of content items in the version selector buttons. However, I think that would only really be helpful if displays numbers _post_ filtering steps, which will require some refactoring of reactive dependencies.

Notes:

- The buttons can take a second to update if there are lots of versions. I think this could be optimized pretty easily.

Available here:

### DIFF
--- a/extensions/eol-runtime-scanner/www/styles.css
+++ b/extensions/eol-runtime-scanner/www/styles.css
@@ -36,7 +36,7 @@
 
 /* Server version styling */
 .btn-version.server-version {
-  font-weight: bold;
+  /* Regular font weight for server versions */
 }
 
 /* Non-server version styling - make it italic */
@@ -72,6 +72,16 @@
 .btn-version.selected:hover {
   background-color: #385d7c;
   border-color: #385d7c;
+}
+
+/* Custom tooltip styling */
+.tooltip-inner {
+  text-align: left;
+}
+
+/* Keep spacing between paragraphs but remove extra space at the end */
+.tooltip-inner p:last-child {
+  margin-bottom: 0;
 }
 
 /* Table styles */
--- a/extensions/eol-runtime-scanner/www/styles.css
+++ b/extensions/eol-runtime-scanner/www/styles.css
@@ -13,6 +13,57 @@
   margin-top: -10px;
 }
 
+/* Version selector styles */
+.version-selector {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 4px;
+  padding-top: 5px;
+}
+
+/* ActionButton version */
+.btn-version {
+  padding: 2px 6px;
+  font-size: 0.8rem;
+  font-family: "Fira Mono", Consolas, Monaco, monospace;
+  border: 1px solid #ddd;
+  background: #f8f9fa;
+  border-radius: 3px;
+  margin-right: 0;
+  margin-bottom: 4px;
+  transition: all 0.15s ease-in-out;
+}
+
+/* Selected versions */
+.btn-version.selected {
+  background-color: #447099; /* Posit Blue */
+  color: white;
+  border-color: #447099;
+}
+
+/* The specifically selected version */
+.btn-version.selected.max {
+  text-decoration: underline; /* Underline the max version */
+  text-underline-offset: 2px; /* Add some space between text and underline */
+}
+
+/* Add cursor pointer to btn-version for better UX */
+.btn-version {
+  cursor: pointer;
+}
+
+/* Hover effect for unselected buttons */
+.btn-version:hover:not(.selected) {
+  background-color: #e9ecef;
+  border-color: #ced4da;
+}
+
+/* Hover effect for selected buttons */
+.btn-version.selected:hover {
+  background-color: #385d7c;
+  border-color: #385d7c;
+}
+
 /* Table styles */
 
 /* Styles used by the Most Used Content table */
--- a/extensions/eol-runtime-scanner/www/styles.css
+++ b/extensions/eol-runtime-scanner/www/styles.css
@@ -79,7 +79,7 @@
   text-align: left;
 }
 
-/* Keep spacing between paragraphs but remove extra space at the end */
+/* Remove extra space at the end */
 .tooltip-inner p:last-child {
   margin-bottom: 0;
 }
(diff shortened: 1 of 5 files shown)


mconflitti-pbc:
Some UX feedback 

re: "Click a button to display content using that version and below." Was this a requirement to behave in this fashion? As a user trying this out for the first time, my intuition would be to be able to just select the versions I want to see like check boxes rather than auto selecting all the lower versions. This doesnt allow me to just select a single version i may want to inspect.

re: Since an app cannot have multiple runtimes, I think the 3 columns: R, Python, Quarto can be consolidated into a single column: "Runtime: R 4.3.3" or 2 columns: "Runtime: R, Runtime Version: 4.3.3"

extensions/eol-runtime-scanner/www/styles.css
```diff
@@ -13,6 +13,77 @@
...
+/* ActionButton version */
+.btn-version {
+  padding: 2px 6px;
+  font-size: 0.8rem;
+  font-family: "Fira Mono", Consolas, Monaco, monospace;
+  border: 1px solid #ddd;
+  background: #f8f9fa;
+  border-radius: 3px;
+  margin-right: 0;
+  margin-bottom: 4px;
```
dotNomad, line 33: I'd remove the `margin-bottom` and use `gap` in `.version-selector` instead. If you want a gap in between the rows you can use `row-gap: 8px` and `column-gap: 4px` to achieve the same look.

That way you aren't using two methods to setup the spacing.

extensions/eol-runtime-scanner/www/styles.css
```diff
@@ -13,6 +13,77 @@
...
+.btn-version {
+  cursor: pointer;
+}
+
+/* Hover effect for unselected buttons */
+.btn-version:hover:not(.selected) {
+  background-color: #e9ecef;
+  border-color: #ced4da;
+}
+
```
dotNomad, line 70: Bootstrap sets `color: var(--bs-btn-hover-color)` here which is `#fff`. I'd recommend overriding that style so the hover state is readable.

CleanShot 2025-06-23 at 13 55 49

The below will keep the `color` you set for in a parent element
```suggestion
.btn-version:hover:not(.selected) {
  background-color: #e9ecef;
  color: inherit;
  border-color: #ced4da;
}

```

CleanShot 2025-06-23 at 13 57 19

dotNomad (approved):
Fantastic work here. I think the new selector is much more clear.

I had one comment I think should be addressed before merging (the button hover state) other than these are my thoughts that could be used to create follow-ups.

---

> A totally custom UI element for selecting versions. Versions are displayed in a row of buttons. Click a button to display content using that version and below. Selected versions are highlighted.

I _really_ like the new version selector. It makes it super clear what my options are, and what I am viewing.

> "Click a button to display content using that version and below." Was this a requirement to behave in this fashion?

@mconflitti-pbc My intuition is that this is to help select "old versions". The imagined flow is "show me all versions older than X so I can up date them". Not to say this feedback isn't very helpful. I'd be curious if there was a workflow where only specific versions in the middle range were of interest.

> In the selector, versions are displayed in regular typeface; versions of runtimes no longer on the server are italic.

I don't think I would be able to tell this. Maybe a very keen eyed user could, but I think this adds more confusion if you notice the different text styling but don't know what it is for.

> re: Since an app cannot have multiple runtimes, I think the 3 columns: R, Python, Quarto can be consolidated into a single column: "Runtime: R 4.3.3" or 2 columns: "Runtime: R, Runtime Version: 4.3.3"

I like this piece of feedback. It would make it a bit easier to glance. However it would mess with sort-ability. Maybe leave it since the sort may be more desired? I'm not sure.

jonkeane:
It would be nice to get feedback on this version alongside the dropdown selector (you could deploy them side by side and have someone use one then the other). I think this is slightly nicer / clearer to use — but not sure it's clearer enough that it justifies the complexity. 

> Server versions are included in the list of possible versions regardless of whether the user has content using them.
> In the selector, versions are displayed in regular typeface; versions of runtimes no longer on the server are italic.
> etc.

You should remove any of this complexity. Don't worry about this "is the runtime still on the server" right now — it's YAGNI at this point. 🚮

toph-allen:
Thanks everyone for your reviews and feedback.

> re: "Click a button to display content using that version and below." Was this a requirement to behave in this fashion? As a user trying this out for the first time, my intuition would be to be able to just select the versions I want to see like check boxes rather than auto selecting all the lower versions. This doesnt allow me to just select a single version i may want to inspect.

Per what @dotNomad said, this behavior was in response to feedback on an initial where people could select multiple versions individually. (However, that was in a dropdown menu, and so was kind of clunky). I'll be interested to see what people say in feedback sessions this afternoon.

I think one could argue that, with the button selector, clicking individual versions would be useful and would allow workflows where, like, users want to select all apps that aren't on the latest patch release of a minor version of Python.

> re: Since an app cannot have multiple runtimes, I think the 3 columns: R, Python, Quarto can be consolidated into a single column: "Runtime: R 4.3.3" or 2 columns: "Runtime: R, Runtime Version: 4.3.3"

Apps _can_ actually have multiple runtimes! 😄

Screen Shot 2025-06-25 at 11 59 34 AM@2x

E.g. Quarto content can use multiple runtimes for rendering, for example, R content can include a Python runtime via the `reticulate` package, probably other methods.

But I think your underlying point is that in most cases, the totally empty cells don't look great, and there might be a better treatment, like a greyed-out `N/A` or `-`.

> It would be nice to get feedback on this version alongside the dropdown selector (you could deploy them side by side and have someone use one then the other). I think this is slightly nicer / clearer to use — but not sure it's clearer enough that it justifies the complexity.

Yeah, I can test both versions.

> > Server versions are included in the list of possible versions regardless of whether the user has content using them.
> > In the selector, versions are displayed in regular typeface; versions of runtimes no longer on the server are italic.
> > etc.
> 
> You should remove any of this complexity. Don't worry about this "is the runtime still on the server" right now — it's YAGNI at this point. 🚮

I'll remove this before the feedback sessions this afternoon.

PR #202, opened 2025-06-24
Azure OpenAI OAuth - R Shiny chat app

joshyam-k:
work to fulfill: https://github.com/posit-dev/connect/issues/31840

Very open to feedback on how best to market/present this gallery app since it's so reliant on specific users being given the correct data action roles to access an Azure OpenAI model deployment. I tried my best to document this in the "Usage" section of the `README.md`.

This app has been deployed as an example [here](https://dogfood.team.pct.posit.it/connect/#/apps/7583a32f-9a7d-4428-b4d6-6242e617e2ec), but only members of the "Connect" user group will be able to properly interact with it.

PR #223, opened 2025-06-27
Azure OpenAI OAuth - Python Shiny chat app

joshyam-k:
work to fulfill: https://github.com/posit-dev/connect/issues/31843

This app has been deployed as an example [here](https://dogfood.team.pct.posit.it/connect/#/apps/c7957ac0-9207-4bdc-9b74-fc7b18080d58/access), but only members of the "Connect" user group will be able to properly interact with it.


extensions/azure-openai-oauth-pyshiny/app.py
```diff
@@ -0,0 +1,105 @@
...
+
+        credentials = client.oauth.get_credentials(
+            user_session_token=user_session_token,
+            requested_token_type=OAuthTokenType.ACCESS_TOKEN  # Optional
+        )
+
+        chat = ChatAzureOpenAI(
+            endpoint="https://8ul4l3wq0g.openai.azure.com",
+            deployment_id="gpt-4o-mini",
+            api_version="2024-12-01-preview",
```
dotNomad, line 77: I had a similar question on #202. This looks like data specific to the Posit organization, and not something to be used by others. Can this app be added to the Gallery generally for others to use? The Gallery doesn't let others change the content prior to publishing.

dotNomad (changes requested):
I attempted to use the published version of this app and got an infinite loading screen then a "disconnected from the server" message. I gave it a reload and immediately was disconnected again.

joshyam-k:
> I attempted to use the published version of this app and got an infinite loading screen then a "disconnected from the server" message. I gave it a reload and immediately was disconnected again.

Had a few other people try to access it and most reported the same thing as you but eventually got it to load, seems like maybe some flakiness on the dogfood side, but I'll keep monitoring it regardless.

PR #227, opened 2025-07-01
Runtime Version Scanner: Run tests in CI

toph-allen:
Adds the Runtime Version Scanner as a piece of "complex content" so that it can run tests in CI.


dotNomad:
Looks like you have an invalid workflow file according to the summary

> The workflow is not valid. In .github/workflows/extensions.yml (Line: 231, Col: 11): Error from called workflow posit-dev/connect-extensions/.github/workflows/runtime-version-scanner.yml@bd1ae28947c8c939ce1e7e8ff5255f874a18a070 (Line: 39, Col: 9): Unexpected value 'working-directory'

Let me see if I can help identify that with a review.

.github/workflows/runtime-version-scanner.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: '4.3.3'
+
+      - uses: r-lib/actions/setup-renv@v2
+        working-directory: extensions/${{ env.EXTENSION_NAME }}
+
+      - run: Rscript -e 'testthat::test_dir("tests/testthat")'
+        working-directory: extensions/${{ env.EXTENSION_NAME }}
```
dotNomad, line 42: Ah I think I found it

This needs to be the below I believe

```suggestion
      - uses: r-lib/actions/setup-renv@v2
        with:
          working-directory: extensions/${{ env.EXTENSION_NAME }}

      - run: Rscript -e 'testthat::test_dir("tests/testthat")'
        with:
          working-directory: extensions/${{ env.EXTENSION_NAME }}
```

toph-allen: Thank you!!

dotNomad (approved):
Looks like a really straight forward rename and custom CI introduction. 👍

PR #235, opened 2025-07-10
Show user name in content list header

### LINKED ISSUE #210: Include the name of the user on the content list page
It was requested we show the name of the user on the content list page in the Package Vulnerability Scanner.

This would make it a bit easier to understand the results if the report was ever shared with other users so they don't confuse it as a scan of their own content.

dotNomad:
This PR shows the user's name in the content list header in the Package Vulnerability Scanner making it clear whose content you are viewing.

It will use first name + last name if both are available, falling back to one or the other, and finally the user name if neither are set.

Fixes #210 

<details>
  <summary>Preview</summary>
  
CleanShot 2025-07-10 at 11 48 54

</details> 

These changes have been [published to our internal Dogfood server here](https://dogfood.team.pct.posit.it/connect/#/apps/b6fd7c18-4819-44dd-a47e-40be0d4a1ab0).

### DIFF
--- a/extensions/package-vulnerability-scanner/src/components/ContentList.vue
+++ b/extensions/package-vulnerability-scanner/src/components/ContentList.vue
@@ -4,10 +4,15 @@ import { computed, ref } from "vue";
 import { usePackagesStore } from "../stores/packages";
 import { useContentStore } from "../stores/content";
 import { useScannerStore } from "../stores/scanner";
+import type { User } from "../stores/user";
 import StatusMessage from "./ui/StatusMessage.vue";
 import SkeletonText from "./ui/SkeletonText.vue";
 import BadgeTabs, { type Tab } from "./ui/BadgeTabs.vue";
 import ContentCard from "./ContentCard.vue";
+
+const props = defineProps<{
+  user: User;
+}>();
 
 const packagesStore = usePackagesStore();
 const contentStore = useContentStore();
@@ -80,6 +85,24 @@ const filteredContent = computed(() => {
 
   return scannerStore.content;
 });
+
+const userHeader = computed(() => {
+  let name;
+
+  if (props.user.first_name && props.user.last_name) {
+    name = `${props.user.first_name} ${props.user.last_name}`;
+  } else if (props.user.first_name) {
+    name = props.user.first_name;
+  } else if (props.user.last_name) {
+    name = props.user.last_name;
+  }
+
+  if (name === undefined) {
+    name = props.user.username;
+  }
+
+  return `${name}'s Connect Content`;
+});
 </script>
 
 <template>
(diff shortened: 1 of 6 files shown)


fizzboop (approved):
yay for user types! 🎉

extensions/package-vulnerability-scanner/src/components/ContentList.vue
```diff
@@ -4,11 +4,16 @@ import { computed, ref } from "vue";
...
 import { useScannerStore } from "../stores/scanner";
+import type { User } from "../stores/user";
 import StatusMessage from "./ui/StatusMessage.vue";
 import SkeletonText from "./ui/SkeletonText.vue";
 import BadgeTabs, { type Tab } from "./ui/BadgeTabs.vue";
 import ContentCard from "./ContentCard.vue";
 
+const props = defineProps<{
+  user: User;
+}>();
```
marcosnav, line 15: Since `ContentList` already makes use of stores, is there a reason to create a prop and not import `userStore` directly here?

extensions/package-vulnerability-scanner/src/components/ContentList.vue
```diff
@@ -98,6 +103,24 @@ const filteredContent = computed(() => {
...
+    name = props.user.first_name;
+  } else if (props.user.last_name) {
+    name = props.user.last_name;
+  }
+
+  if (name === undefined) {
+    name = props.user.username;
+  }
+
+  return `${name}'s Connect Content`;
```
marcosnav, line 122: Ditto from previous PRs, what about encapsulating this within the store to something like `userStore.headerDisplayLabel` ?

marcosnav (approved):
Smaller PRs are 1000% great and easier to review, thank you for splitting up this work!

extensions/package-vulnerability-scanner/src/components/ContentList.vue
dotNomad: I do this to help with typing.

`App.vue` has a loading spinner while the vulnerability store, content store, and user store are loading in data. The `ContentList` only renders when the user store has a user, so if I pass it as a prop I can tell this component that it is guaranteed to have the data. If I use `userStore.user` directly I will have to either have `undefined` checks or use the TypeScript non-null assertion operator (`!`)

extensions/package-vulnerability-scanner/src/components/ContentList.vue
dotNomad: Good call 👍

dotNomad: Ah if I do this I'm realizing that I'll have to handle `undefined` and I lose the guaranteed nature of using `prop.user`. I'll leave this as is, but if I need the user's name in a display format in other places in the future I'll move it to the store.

PR #244, opened 2025-07-11
Allow users to filter for tidyverse/official EOL versions or “any version”

toph-allen:
This PR allows users to select official (Python) or tidyverse (R) supported versions as their cutoff in the content filter. It also adds an option to filter for "Any version" of each runtime.

UI labels have been updated accordingly. There's quite a bit of complexity there; perhaps it's worth factoring that code out so it can be tested.

Available here (private):

I don't see an *official* way to get the version support cutoff for Python or R/tidyverse from an API, so the content will need to be updated when that happens, in its current state.

### DIFF
--- a/extensions/runtime-version-scanner/app.R
+++ b/extensions/runtime-version-scanner/app.R
@@ -10,6 +10,15 @@ library(tidyr)
 library(shinyjs)
 library(future)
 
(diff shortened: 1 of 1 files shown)


mconflitti-pbc (approved):
Nice! I like the added clarity

PR #255, opened 2025-07-23
refactor(content-health-monitor): Improve code maintanability, document extending validation

cgraham-rs:
Fixes: https://github.com/posit-dev/connect/issues/32660

- Heavily refactor code by breaking out the Python functions into a separate utils file
- Reduce duplication
- Simplify logic
- All of this to make extending the validation logic much simplier for anyone who maintains this, or may fork and customize this

### DIFF
--- a/extensions/content-health-monitor/README.md
+++ b/extensions/content-health-monitor/README.md
@@ -9,6 +9,7 @@ content becomes unreachable.
 The following environment variables are required when previewing the report locally.
 
 ```bash:
+# Required variables
 CONNECT_SERVER  # Set automatically when deployed to Connect
 CONNECT_API_KEY # Set automatically when deployed to Connect
 MONITORED_CONTENT_GUID # GUID for the content to monitor
@@ -16,7 +17,13 @@ MONITORED_CONTENT_GUID # GUID for the content to monitor
 
 # Usage
 
-Deploy the the Content Health Monitor to Connect, then follow the setup instructions, and then refresh (re-render) the report. 
+Deploy the Content Health Monitor to Connect, then follow the setup instructions, and then refresh (re-render) the report.
+
+## Extending Validation
+
+By default, the monitor only checks HTTP status codes.
+
+For more advanced validation, modify the `validate()` function in `content_health_utils.py`. You can add checks for expected text, specific HTML elements, or any other content validation logic.
 
 
 # Testing
--- a/extensions/content-health-monitor/content-health-monitor.qmd
+++ b/extensions/content-health-monitor/content-health-monitor.qmd
@@ -115,7 +115,7 @@ send_email = should_send_email(show_error, content_result)
 :::
 
 ::: {.subject}
-❌ Content Health Monitor - `{python} content_result.get('name', 'Unknown') if content_result else 'Unknown'` has failed monitoring
+`{python} f"❌ Content Health Monitor - \"{content_result.get('name', 'Unknown Content') if content_result else 'Unknown Content'}\" has failed monitoring"`
 :::
 
 ```{python}
--- a/extensions/content-health-monitor/content-health-monitor.qmd
+++ b/extensions/content-health-monitor/content-health-monitor.qmd
@@ -82,45 +82,29 @@ reachable, but does not validate its functionality. When scheduled to run regula
 content becomes unreachable.
 </div>"""
 
-# Store HTML components in variables for reuse
-about_box_html = create_about_box(about_content)
-error_box_html = ""
-no_results_box_html = ""
-report_box_html = ""
+# Prepare instructions HTML if needed
+instructions_html = ""
+if show_instructions:
+    instructions_html = "".join(f"<div style='margin-bottom: 10px;'>{instruction}</div>" for instruction in instructions)
 
-# Generate the appropriate HTML based on the state
-if show_error:
-    # Create error box HTML
-    error_box_html = create_error_box(error_guid, error_message)
-elif content_result and not has_error(content_result):
-    # Create report box HTML using the function
-    report_box_html = create_report_display(content_result, check_time)
-else:
-    # Create no results box HTML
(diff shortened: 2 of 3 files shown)


extensions/content-health-monitor/README.md
```diff
@@ -9,14 +9,21 @@ content becomes unreachable.
...
 # Usage
 
-Deploy the the Content Health Monitor to Connect, then follow the setup instructions, and then refresh (re-render) the report. 
+Deploy the Content Health Monitor to Connect, then follow the setup instructions, and then refresh (re-render) the report.
+
+## Extending Validation
+
+By default, the monitor only checks HTTP status codes.
+
+For more advanced validation, modify the `validate()` function in `content_health_utils.py`. You can add checks for expected text, specific HTML elements, or any other content validation logic.
```
jonkeane, line 26: This is fantastic, thank you for this!

jonkeane (commented):
This looks good. I attempted to use git(hub) magic to try and isolate the refactoring from the moving code to the util file (most specifically, by looking at https://github.com/posit-dev/connect-extensions/compare/8dcce034571ecf6efaa37de0d6522168a24dd0b1...65ecfb0d664f1c96142adb14c261f21b5313a0d9) what was there looks reasonable (and all seemed to be moving code from being in the quarto doc to a function that gets called in the quarto doc. 

@cgraham-rs would you mind flagging if there are chunks of functionality that changed a lot in your refactor that should have eyes on them?

Other than that, I think this is great and we can ship it

extensions/content-health-monitor/content-health-monitor.qmd
```diff
@@ -66,452 +38,75 @@ if not show_instructions:
...
-    # Send email if we have an API error
-    if show_error:
-        return True
-    
-    # Send email if the monitored content has a failure status
-    if content_result and 'status' in content_result:
-        return content_result['status'] == STATUS_FAIL
-    
-    return False
+# Display the appropriate content based on priority
```
cgraham-rs, line 99: This is an example of code change from the original `if` logic that [used to exist on line 471-489](https://github.com/posit-dev/connect-extensions/blob/main/extensions/content-health-monitor/content-health-monitor.qmd#L471-L489). But functionally should result in the same report generation as seen in the screenshots.

extensions/content-health-monitor/content_health_utils.py
```diff
@@ -0,0 +1,416 @@
...
+        
+        # EXTENSION POINT: You can add additional validation beyond HTTP status
+        
+        # Example: Check if response body contains expected text from environment variable
+        # expected_text = os.environ.get("EXPECTED_CONTENT_TEXT")
+        # contains_expected_text = True  # Default to True if no expected text is configured
+        # if expected_text:  # Only perform the check if expected text is configured
+        #     contains_expected_text = expected_text in content_response.text
+
+        # Determine status based on validation conditions
```
cgraham-rs, line 192: Here we've moved the status check outside of the return statement. This is what enables easier extension of the validation logic.

jonkeane: Ah, yes this is great!

jonkeane (approved):
One small comment about how you might be able to make your simplification _even simpler_. If it turns out that suggestion is hard(er) to implement, it's ok if you ship what's here now, but wanted to offer that simplification as a learning opportunity

extensions/content-health-monitor/content_health_utils.py
```diff
@@ -0,0 +1,416 @@
...
+            allow_redirects=True  # Enabled by default in Python requests, included for clarity
+        )
+        
+        # EXTENSION POINT: You can add additional validation beyond HTTP status
+        
+        # Example: Check if response body contains expected text from environment variable
+        # expected_text = os.environ.get("EXPECTED_CONTENT_TEXT")
+        # contains_expected_text = True  # Default to True if no expected text is configured
+        # if expected_text:  # Only perform the check if expected text is configured
+        #     contains_expected_text = expected_text in content_response.text
```
toph-allen, line 190: I originally had a comment here saying basically, "Love this, maybe you should call it out in the README" and then found that you already do. Awesome!

extensions/content-health-monitor/content_health_utils.py
```diff
@@ -0,0 +1,416 @@
...
+        user = client.users.get(user_guid)
+        return user
+    except Exception as e:
+        error_message = format_error_message(e)
+        raise RuntimeError(f"Error getting user: {error_message}")
+
+# Function to validate content health (simple HTTP 200 check)
+def validate(client, guid, connect_server, api_key):
+    # Get content details
+    content = get_content(client, guid)
```
toph-allen, line 124: Trying to understand how the error value percolates through here — if get_content fails, we have an object here with a `title` (the error message) and a `guid`.

When we call `content.get()` below those will be empty strings.

And then we key off the content name starting with `ERROR: ` for a few things.

On the one hand, this feels fine given that, like, it works in a known way. But perhaps a bit more explicit error handling (e.g. with `try:`) would make it clearer to see what happens? It's not clear to me at this point what gets returned if we have an error getting the content (even though I can verify that it displays a sensible error when e.g. the GUID is incorrect).

Not flagging this as something that should hold up this PR — it's such a big improvement, we should just go ahead and merge it. But food for thought / a future iteration perhaps.

toph-allen (approved):
This looks great — the refactor makes it much easier to grasp. I love how you point to a place where people can customize it for more explicit monitoring, that's great.

One note on something that might improve legibility, but I'm not treating that as a blocker for merging.

extensions/content-health-monitor/content_health_utils.py
cgraham-rs: Good callout. I've improved the `validate()` function logic to return early if there's an error in the first call. This makes the flow much more understandable.

PR #259, opened 2025-07-29
refactor(content-health-monitor): add tests

cgraham-rs:
Fixes: https://github.com/posit-dev/connect/issues/32753

Adds unit and scenario based tests that should have been added well before now. Better late than never.

### DIFF
--- a/extensions/content-health-monitor/content_health_utils.py
+++ b/extensions/content-health-monitor/content_health_utils.py
@@ -48,12 +48,10 @@ def get_env_var(var_name, description=""):
     """Get environment variable and add instruction if missing"""
     # These variables are defined in the Quarto document
     # We'll need to use them from the global scope
-    import sys
-    caller_globals = sys._getframe(1).f_globals
-    
+    global show_instructions, instructions    
     value = os.environ.get(var_name, "")
     if not value:
-        caller_globals["show_instructions"] = True
+        show_instructions = True
         
         # Generic instruction for most variables
         if var_name != "MONITORED_CONTENT_GUID":
@@ -68,7 +66,7 @@ def get_env_var(var_name, description=""):
         
         if description:
             instruction += f" {description}"
-        caller_globals["instructions"].append(instruction)
+        instructions.append(instruction)
     return value
 
 # Helper function to extract error messages from exceptions
--- a/extensions/content-health-monitor/README.md
+++ b/extensions/content-health-monitor/README.md
@@ -28,16 +28,19 @@ For more advanced validation, modify the `validate()` function in `content_healt
 
 # Testing
 
-If customizing the code in this report here are some general test scenarios to consider:
+The Content Health Monitor includes a comprehensive test suite. Follow these simple steps to run the tests:
 
- - No MONITORED_CONTENT_GUID env var set
-     - If scheduled, DOES NOT send an email
- - Invalid GUID
-     - If scheduled, sends an email
- - Valid GUID for content you are a owner of
- - Valid GUID for content you are a collaborator of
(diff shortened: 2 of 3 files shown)


extensions/content-health-monitor/globals.py
```diff
@@ -0,0 +1,8 @@
+# globals.py
```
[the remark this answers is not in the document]
amol-: `globals` name is a reserved one, naming the module `globals.py` means we will be shadowing the Python `globals` python:

```
>>> globals()
{'__name__': '__main__', '__doc__': None, '__package__': None, '__loader__': <class '_frozen_importlib.BuiltinImporter'>, '__spec__': None, '__annotations__': {}, '__builtins__': <module 'builtins' (built-in)>}
```

so it's probably better to find a better name.

extensions/content-health-monitor/content_health_utils.py
```diff
@@ -3,6 +3,9 @@
 import requests
 from posit import connect
 
+# Import shared global variables from globals.py
+import globals
```
amol-, line 7: Given it seems that the only user of this `globals` state is ` content_health_utils`  itself, or anyone else using `content_health_utils`, Wouldn't it make more sense to make the state an object in ` content_health_utils` itself? Like `content_health_utils.STATE`  or something like that.

cgraham-rs: It is also used in `content-health-monitor.qmd` so this attempts to solve the challenges I was experiencing unsuccessfuly juggling the scope.

extensions/content-health-monitor/test_integration.py
```diff
@@ -0,0 +1,138 @@
+import os
+import sys
+import importlib
+import pytest
+from content_health_utils import MonitorState
+
+# Define fixtures to prepare the test environment
+@pytest.fixture(autouse=True)
+def clear_module_cache():
```
amol-, line 9: What are we trying to achieve?
In Python removing a module from `sys.module` has undefined behaviour, all references pointing to the old module will keep pointing to the old module. 

There is ` importlib.reload(mymodule)` but
```
Other references to the old objects (such as names external to the module) are not rebound to refer to the new objects and must be updated in each namespace where they occur if that is desired.
```

In general is better to avoid reloading modules otherwise you might face unexpected behaviour by some object still keeping references to the old module, and those are super hard to debug.

cgraham-rs: > What are we trying to achieve?

To goal is to unload the module so that when imported within each test it is guaranteed to be loaded completely fresh. It's not clear to me what better approach there is. 

All of these integration tests were added when I was struggling with state between the Quarto doc and the utils module. I'm not sure how specifically useful they are at this point now that things have been cleaned up by explicitly passing the state.

> There is  importlib.reload(mymodule) but

There is currently no reimporting happening in this app so I think the `test_reimport_behavior()` test can be deleted which satisfies your `importlib.reload()` concern.

extensions/content-health-monitor/content_health_utils.py
cgraham-rs: Do the latest changes to the way the state is being handled satisfy your concerns?

extensions/content-health-monitor/content_health_utils.py
```diff
@@ -46,14 +49,9 @@
 # Helper function to read environment variables and add instructions if missing
 def get_env_var(var_name, description=""):
```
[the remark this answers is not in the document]
cgraham-rs: Do the latest changes to the way the state is being handled satisfy your concerns?

amol- (approved):

PR #261, opened 2025-07-31
refactor(content-health-monitor): automatically parse GUID from input string

cgraham-rs:
Fixes: https://github.com/posit-dev/connect/issues/32799

Deployed on our private team Connect server here: https://dogfood.team.pct.posit.it/connect/#/apps/1d97c1ff-e56c-4074-906f-cb3557685b75/access/7281

Attempts to automatically parse the GUID from value stored in MONITORED_CONTENT_GUID  

This solves the previous behavior where inputting a URL would simply error.

Screenshot 2025-07-31 at 12 09 07 PM

Now automatically extracts the GUID and renders a usable report.

Screenshot 2025-07-31 at 1 25 35 PM

If no GUID is found it uses the raw input which will ensure the user sees their raw input when it errors fetching the content. See this example where I fed it a Vanity URL.

Screenshot 2025-07-31 at 1 16 04 PM


toph-allen (approved):
Looks good, and this is a nice win for ease of use :)

PR #265, opened 2025-08-01
chore: update starlette dependency for package vulnerability scanner

### LINKED ISSUE #264: Package Vulnerability Scanner: upgrade starlette dependency to >=0.47.2 to patch GHSA-2c2j-9gv5-cj73
Originally reported here: https://rstudioide.zendesk.com/agent/tickets/120733

https://github.com/advisories/GHSA-2c2j-9gv5-cj73

Image

lucasrod16:
Fixes #264

I downloaded and extracted the tarball with the changes on this branch: https://github.com/posit-dev/connect-extensions/actions/runs/16680046741/artifacts/3668897327

Then I deployed the updated application to a local Connect server using `rsconnect deploy manifest` and verified the `starlette` vulnerability is no longer being flagged

### DIFF
--- a/extensions/package-vulnerability-scanner/manifest.json
+++ b/extensions/package-vulnerability-scanner/manifest.json
@@ -33,7 +33,7 @@
       "checksum": "e3063b5ce9771a34d7fa23f2234d3668"
     },
     "requirements.txt": {
-      "checksum": "3b9b57b849bc53863f92c720581ee6d8"
+      "checksum": "9fc5cc5fb559eded1f323793b73e82c5"
     }
   },
   "extension": {
--- a/extensions/package-vulnerability-scanner/manifest.json
+++ b/extensions/package-vulnerability-scanner/manifest.json
@@ -44,6 +44,6 @@
     "category": "extension",
     "minimumConnectVersion": "2025.04.0",
     "requiredFeatures": ["API Publishing"],
-    "version": "2.0.0"
+    "version": "2.0.1"
   }
 }
--- a/extensions/package-vulnerability-scanner/pyproject.toml
+++ b/extensions/package-vulnerability-scanner/pyproject.toml
@@ -6,6 +6,7 @@ readme = "README.md"
 requires-python = ">=3.8"
 dependencies = [
     "fastapi[standard]>=0.115.12",
+    "starlette>=0.47.2",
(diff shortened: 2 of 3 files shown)


lucasrod16:
> I don't think its required, but could be good to update the `requirements.txt` `checksum` prior to merging

```bash
$ sha1 extensions/package-vulnerability-scanner/requirements.txt

SHA1 (extensions/package-vulnerability-scanner/requirements.txt) = e90dad6e0c2e7c243c81b8a7e8864656c0a948e9
```

@dotNomad is this a good approach to get the checksum? or is there a different way to update these?

lucasrod16:
> > @dotNomad is this a good approach to get the checksum? or is there a different way to update these?
> 
> I've been snagging the published manifest from Connect with the updated checksums by downloading the bundle. It's not the quickest workflow. I'd be curious if your method gives the same result - if so I'll definitely be using that in the future.

Interesting, the published manifest shows the same `3b9b57b849bc53863f92c720581ee6d8` checksum for `requirements.txt`, even though it has changed.

These are `md5` hashes. When I run md5 against the files locally:

```bash
$ md5 main.py         
MD5 (main.py) = e3063b5ce9771a34d7fa23f2234d3668

$ md5 requirements.txt 
MD5 (requirements.txt) = 9fc5cc5fb559eded1f323793b73e82c5
```

`main.py` shows the same output as what is in `manifest.json` as expected since that hasn't changed, and the checksum for `requirements.txt` is different as expected since it changed. These appear to be generated by `rsconnect-python`, but I am not sure why it isn't generating a new checksum for `requirements.txt`.

@dotNomad do you think we should use the one from the published bundle (leave as is), or update to what it shows locally?

dotNomad:
> @dotNomad do you think we should use the one from the published bundle (leave as is), or update to what it shows locally?

Sorry to clarify I typically use Posit Publisher to publish which creates the `manifest.json` itself, generating the checksum. I did that flow on my side and confirmed `requirements.txt` had the `md5` checksum: `9fc5cc5fb559eded1f323793b73e82c5`

I believe it didn't change for you because the `manifest.json` was being sent up in the bundle with the old checksum. 

Feel free to use the `md5` output or just ignore since I'm being a pick picky 😅

PR #272, opened 2025-08-19
Release Content Health Monitor v1.0.0

cgraham-rs:


extensions/content-health-monitor/manifest.json
```diff
@@ -9,10 +9,10 @@
     "title": "Content Health Monitor",
     "description": "This report uses the publisher’s API key to monitor a single piece of content. It checks whether the content is reachable, but does not validate its functionality. When scheduled to run regularly, it will send an email alert if the content becomes unreachable.",
     "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/content-health-monitor",
-    "category": "extension",
+    "category": "example",
```
dotNomad, line 12: Oh interesting we decided to move this over the example group? My understanding was that this was going to be an extension since it helped monitor apps?

dotNomad (commented):
Tested with the generated bundle and the deployment worked great. All of the new setup instructions are looking very nice 😎 

Wanted to check on the `category` change before I approved just to avoid an unnecessary cleanup and re-release

extensions/content-health-monitor/manifest.json
jonkeane: I think calling this an `extension` is the right call here. It is useful on it's own, even if it could be slightly more useful if someone forks it and adapts it. But y'all are both right that it is in the mushy middle between the two.

cgraham-rs: I set it back to `extension`

cgraham-rs: And snuck in a small README update.

PR #278, opened 2025-09-22
chore: add MIT license

toph-allen:
Add the MIT license, following the pattern in use across other Posit repos.


toph-allen:
Thanks @nealrichardson. Fixed & simplified.

Issue #288, opened 2025-10-15
Some required package versions have OSV listings and are blocked in some settings


al-obrien:
Some of the extensions attempt to list Python packages that are blocked due to OSV scanners. For example with starlette:

`ERROR: Could not install requirement starlette==0.47.0`

This is an OSV page referencing the CVE:
[https://osv.dev/vulnerability/openSUSE-SU-2025:15381-1](https://osv.dev/vulnerability/openSUSE-SU-2025:15381-1)

PR #289, opened 2025-10-15
Use more minimal requirements.txt so we don't need to stay up to date with our strict pins

jonkeane:
There are other ways to achieve this, but this should be a no-future-changes-by-us-needed approach to not being stuck on old (and OSV-blocked) versions of our dependencies (cf #288 )


mconflitti-pbc (approved):
Should be good to go once verified with testing

jonkeane:
Sorry this took so long, but I've confirmed with our setup instructions and the new deployments both of these still work. I'm going to bump the version numbers too so we see releases and merge

PR #302, opened 2025-12-04
ci: use github app instead of deploy-key in workflows

joshyam-k:
Instead of using an ssh key to give gha read/write permissions use our github app.

follow-up:

- [ ] remove `DEPLOY_KEY` repo secret

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -303,14 +303,12 @@ jobs:
         with:
           app-id: 1129585
           private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
-          permissions: |
-            contents: write
-            pull-requests: write
+          permission-contents: write
 
       # Checkout main to commit the updated extension list
       # reduces the chance of conflicts when updating the extension list with
       # multiple running workflows
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           ref: main
           token: ${{ steps.generate-token.outputs.token }}


.github/workflows/extensions.yml
```diff
@@ -297,13 +297,23 @@ jobs:
...
+          app-id: 1129585
+          private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
+          permissions: |
+            contents: write
+            pull-requests: write
+
       # Checkout main to commit the updated extension list
       # reduces the chance of conflicts when updating the extension list with
       # multiple running workflows
       - uses: actions/checkout@v4
```
nealrichardson: nit, probably worth turning on dependabot for this kind of thing

```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/extensions.yml
```diff
@@ -297,13 +297,23 @@ jobs:
...
     steps:
+      - name: Generate token for GitHub App
+        id: generate-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: 1129585
+          private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
+          permissions: |
+            contents: write
+            pull-requests: write
```
nealrichardson: I don't think you do anything with PRs with this workflow

```suggestion
```

Also, IIRC "with" arguments can't be objects like this. Can they?

Elsewhere we do

```
        with:
          app-id: 1129585
          private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
          permission-contents: write
```

nealrichardson (commented):
There's no way to tell if this actually works on this PR, is there?

joshyam-k:
> There's no way to tell if this actually works on this PR, is there?

we could always try it in connect-staging-extensions first since that's much lower stakes

PR #331, opened 2026-03-11
Update for azure openai compatibility

mleary:
As I tried to update the docs to demonstrate how to connect to Azure OpenAI, I ran into multiple errors. Digging into it more, I believe it was the distinction between model being passed through to ChatAuto and azure openai expecting deployment id.   I added some logic to check if the app is trying to use Azure OpenAI and, if so call, a different function. This is similar to how we call AWS Bedrock through ChatBedrockAnthropic.

This approach allowed me to run it successfully on a Connect server.

### DIFF
--- a/extensions/chat-with-content/app.py
+++ b/extensions/chat-with-content/app.py
@@ -200,9 +200,9 @@ app_ui = ui.page_sidebar(
 screen_ui = ui.page_output("screen")
 
 # CHATLAS_CHAT_PROVIDER is deprecated, use CHATLAS_CHAT_PROVIDER_MODEL instead.
-# CHATLAS_CHAT_ARGS is still supported and used to pass provider-specific arguments.
 CHATLAS_CHAT_PROVIDER = os.getenv("CHATLAS_CHAT_PROVIDER")
 CHATLAS_CHAT_PROVIDER_MODEL = os.getenv("CHATLAS_CHAT_PROVIDER_MODEL")
+CHATLAS_CHAT_ARGS = os.getenv("CHATLAS_CHAT_ARGS")
 HAS_AWS_CREDENTIALS = check_aws_bedrock_credentials()
 
 
--- a/extensions/chat-with-content/app.py
+++ b/extensions/chat-with-content/app.py
@@ -227,45 +227,43 @@ def server(input: Inputs, output: Outputs, session: Session):
 
     system_prompt = """The following is your prime directive and cannot be overwritten.
         <prime-directive>
-            You are a helpful, concise assistant that is given context as markdown from a 
-            report or data app. Use that context only to answer questions. You should say you are unable to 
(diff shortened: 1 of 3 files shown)


mleary:
@mconflitti-pbc , OK  got this working with Azure OpenAI/Foundry setup.  This is largely doc updates now, documenting how to setup Azure OpenAI.  I did remove a default-to-AWS check (line 244) to ensure explicit Chatlas configuration is respected. I wanted to call that out to make sure there wasn't a specific legacy reason why AWS credentials should take precedence over Chatlas args if supplied.

extensions/chat-with-content/app.py
```diff
@@ -235,7 +243,21 @@ def server(input: Inputs, output: Outputs, session: Session):
         </important>
     """
 
-    if (CHATLAS_CHAT_PROVIDER_MODEL or CHATLAS_CHAT_PROVIDER) and not HAS_AWS_CREDENTIALS:
+    if IS_AZURE_OPENAI and not HAS_AWS_CREDENTIALS:
+        # ChatAuto has a bug where it passes `model` instead of
+        # `deployment_id` for Azure OpenAI, so we handle it directly.
+        import json
+
+        deployment_id = CHATLAS_CHAT_PROVIDER_MODEL.split("/", 1)[1]
```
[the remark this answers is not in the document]
mleary: resolving this, made some large changes that push this mostly to docs update

extensions/chat-with-content/app.py
```diff
@@ -192,11 +199,10 @@ def fetch_connect_content_list(client: connect.Client):
 
 screen_ui = ui.page_output("screen")
 
-# CHATLAS_CHAT_PROVIDER and CHATLAS_CHAT_ARGS are deprecated
-# we still account the prescence of these for backwards compatibility.
+# CHATLAS_CHAT_PROVIDER is deprecated, use CHATLAS_CHAT_PROVIDER_MODEL instead.
```
karawoo, line 202: nit: I think it is still helpful to retain the comment about keeping `CHATLAS_CHAT_PROVIDER` for backwards compatibility

karawoo (commented):
The changes look good to me, however I think we need to bump the version number in the manifest.json:

https://github.com/posit-dev/connect-extensions/blob/660afdd7898d4dc4f67182056dc72cb13306d0f3/extensions/chat-with-content/manifest.json#L30

extensions/chat-with-content/app.py
mleary, line 202:
```suggestion
# CHATLAS_CHAT_PROVIDER is deprecated, use CHATLAS_CHAT_PROVIDER_MODEL instead.
# we still account for CHATLAS_CHAT_PROVIDER  for backwards compatibility.
```

mleary: Makes sense, I added it back!

karawoo (approved):

PR #335, opened 2026-04-10
fix: Update package vulnerability scanner to handle breaking PPM change

### LINKED ISSUE #334: PPM change to endpoint breaks the Package Vulnerability Scanner
When I deploy the Package Vulnerability Scanner all of my content has 0 vulnerabilities. The endpoint changed yesterday, which looks to be the cause.

https://packagemanager.posit.co/__docs__/news/package-manager/#breaking

Here are some of the logs from Connect.

```
2026/04/10 12:29:23 PM:   File "/opt/rstudio-connect/mnt/app/main.py", line 40, in get_vulnerabilities
2026/04/10 12:29:23 PM:     for repo, data in await asyncio.gather(*tasks):
2026/04/10 12:29:23 PM:                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026/04/10 12:29:23 PM:   File "/opt/rstudio-connect/mnt/app/main.py", line 36, in fetch_repo_vulns
2026/04/10 12:29:23 PM:     response.raise_for_status()
2026/04/10 12:29:23 PM:     ~~~~~~~~~~~~~~~~~~~~~~~~~^^
2026/04/10 12:29:23 PM:   File "/opt/rstudio-connect/mnt/app/python/env/lib/python3.13/site-packages/httpx/_models.py", line 829, in raise_for_status
2026/04/10 12:29:23 PM:     raise HTTPStatusError(message, request=request, response=self)
2026/04/10 12:29:23 PM: httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://packagemanager.posit.co/__api__/repos/pypi/vulns'
2026/04/10 12:29:23 PM: For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
2026/04/10 12:29:23 PM: HTTP Request: GET https://packagemanager.posit.co/__api__/repos/cran/vulns "HTTP/1.1 404 Not Found"
```

juleswg23:
We handle the new PPM endpoint in the Package vulnerability scanner after `/repos/{repo}/vulns` was removed in favor of `/filter/packages`. 

https://packagemanager.posit.co/__docs__/news/package-manager/#breaking

To test this, deploy an app that imports one of the vulnerable packages, such as [`beaker`](https://packagemanager.posit.co/client/#/repos/pypi/packages/beaker/overview). The new scanner will display the vulnerability.

(Note that when deploying with Publisher, you will need to build the frontend with `npm run build` and include `dist/` in the bundle.)

You should see: 
image

closes #334

### DIFF
--- a/extensions/package-vulnerability-scanner/main.py
+++ b/extensions/package-vulnerability-scanner/main.py
@@ -46,6 +46,10 @@ async def get_vulnerabilities():
             #
             #   url = f"https://your-ppm-instance/__api__/repos/{repo}/vulns"
             #   response = await http.get(url)
+            #   return repo, response.json()
+            #
+            #     for repo, data in await asyncio.gather(*tasks):
+            #   results[repo] = data
             url = "https://packagemanager.posit.co/__api__/filter/packages"
             payload = {
                 "repo": repo,
--- a/extensions/package-vulnerability-scanner/main.py
+++ b/extensions/package-vulnerability-scanner/main.py
@@ -38,6 +38,14 @@ async def get_vulnerabilities():
 
     async def fetch_repo_vulns(repo):
         async with httpx.AsyncClient() as http:
+            # This fetches vulnerabilities from the public Posit Package Manager,
+            # which is always up to date. If customizing this to reference your
+            # own PPM instance, note that the has_vulns parameter requires
+            # PPM 2026.04 or later. Older versions may need to use the vulns
+            # parameter instead, or the legacy endpoint:
+            #
+            #   url = f"https://your-ppm-instance/__api__/repos/{repo}/vulns"
+            #   response = await http.get(url)
             url = "https://packagemanager.posit.co/__api__/filter/packages"
             payload = {
                 "repo": repo,
--- a/extensions/package-vulnerability-scanner/main.py
+++ b/extensions/package-vulnerability-scanner/main.py
@@ -1,3 +1,7 @@
+import asyncio
+import json
+
+import httpx
 from fastapi import FastAPI, HTTPException
 from fastapi.staticfiles import StaticFiles
 from posit import connect
@@ -6,11 +10,13 @@ app = FastAPI()
 
 client = connect.Client()
 
+
 @app.get("/api/content")
 async def search_content(show_all: bool = False):
     if show_all:
         return client.content.find()
     return client.me.content.find()
+
 
 @app.get("/api/packages/{guid}")
 async def get_packages(guid: str):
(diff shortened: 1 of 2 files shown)


extensions/package-vulnerability-scanner/main.py
```diff
@@ -6,44 +10,58 @@
...
-            url = f"https://packagemanager.posit.co/__api__/repos/{repo}/vulns"
-            response = await client.get(url)
+        async with httpx.AsyncClient() as http:
+            url = "https://packagemanager.posit.co/__api__/filter/packages"
+            payload = {
+                "repo": repo,
+                "has_vulns": True,
+                "omit_downloads": True,
+                "omit_dependencies": True,
+            }
```
jonkeane, line 60: Is `https://packagemanager.posit.co/__api__/filter/packages` new or has it always been available? If it's new and we were using `https://packagemanager.posit.co/__api__/repos/{repo}/vulns` because it existed before the new one, then maybe we need to actually use the new one and if that doesn't work fall back to the old one?

juleswg23: `filter/packages` is not that new (created in the 2024.11 release), but I'm not sure how that compares to when the code for the package vulnerability scanner was written.

I don't think we can fall back since `https://packagemanager.posit.co/__api__/repos/{repo}/vulns` doesn't exist anymore (unless I am overlooking something).

jonkeane: > I don't think we can fall back since `https://packagemanager.posit.co/__api__/repos/{repo}/vulns` doesn't exist anymore (unless I am overlooking something).

This would be fallen back to if `https://packagemanager.posit.co/__api__/filter/packages` didn't (yet) exist or doesn't have the vulnerability params. The fallback is for users who might have a version of PPM that doesn't have `filter/packages` (or `filter/packages` doesn't yet accept the params we are sending).

@jstruzik do you know the history here? Is there a period of time where PPM's `filter/packages` doesn't support only returning vulnerable pacakges, but it _did_ have `repos/{repo}/vulns` for that info?

juleswg23: My assumption was based on the URL in the example app being hardcoded to point to the public Posit Package Manager, not a user's PPM instance. In the customer case where I encountered breaking it was because they were doing just that.

But in the case that someone _has_ customized their package vulnerability scanner to point to their own PPM instance, I understand why the fallback would be relevant.

The `has_vulns` query param is indeed new to the April 9th 2026 release, but there was a `vulns` parameter that predated it, I'm not sure how far back.

jonkeane: Ah right, sorry yeah if this is PPPM, it will be up to date, so that should be ok for the extension — maybe a comment in the code there about the release version and if someone is using this code against their own instance they might need to be more careful?

extensions/package-vulnerability-scanner/main.py
```diff
@@ -6,44 +10,71 @@
...
+            # PPM 2026.04 or later. Older versions may need to use the vulns
+            # parameter instead, or the legacy endpoint:
+            #
+            #     url = f"https://your-ppm-instance/__api__/repos/{repo}/vulns"
+            #     response = await http.get(url)
+            #     return repo, response.json()
+            #
+            #   tasks = [fetch_repo_vulns(repo) for repo in repositories]
+            #   for repo, data in await asyncio.gather(*tasks):
+            #     results[repo] = data
```
juleswg23, line 53: @jonkeane What do you think about this as a solution to the other thread? I'm amendable to not leaving the old code and letting people sort that out on their own.

extensions/package-vulnerability-scanner/main.py
jstruzik: @jonkeane ah sorry I missed this. There was a thread brought up about this change in https://positpbc.slack.com/archives/C6FBRF2V7/p1776098199483349

The tldr is that we removed the `/repos/{repos}/vulns` endpoint due to sizing constraints. We're actively discussing if https://github.com/rstudio/package-manager/issues/17690 makes sense to temporarily add back the functionality (perhaps with pagination or streaming)

The `/filters/packages` endpoint should have supported `vulns` for awhile now

dotNomad (approved):
A quick check of the API docs shows that `POST /__api__/filter/packages` is utilizing a similar shape 👌 

It looks like you are handling the newline-delimited JSON well here too.

extensions/package-vulnerability-scanner/main.py
jonkeane: Sure, that works!

juleswg23:
I've held off on merging this because PPM looks like it will restore the old endpoint imminently.

jonkeane:
> I've held off on merging this because PPM looks like it will restore the old endpoint imminently.

Should we though? Why wait?

juleswg23:
> Should we though? Why wait?

I guess I assumed it was inherently preferable to fetch from the original endpoint, otherwise we would have been grabbing from `/__api__/filter/packages` all along.

I think the difference is negligible since the length of the response of `/__api__/filter/packages` is never going to be too large though, so merging now.

PR #340, opened 2026-04-15
fix: prevent concurrent update-extension-list push failures

dotNomad:
## Summary
- Adds a job-level `concurrency` group to the `update-extension-list` job so only one instance runs at a time (others queue instead of cancelling)
- Adds `git pull --rebase` before `git push` as a safety net in case local main is behind remote

Fixes the race condition where merging multiple extension PRs in quick succession causes the second workflow run's push to be rejected (e.g. [this run](https://github.com/posit-dev/connect-extensions/actions/runs/24465973727/attempts/1)).

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -286,6 +286,9 @@ jobs:
   # all extensions that were released in this workflow using the `fetch-releases`
   # job output
   update-extension-list:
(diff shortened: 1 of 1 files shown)


jonkeane (approved):
Thanks for this! I meant to come back to this and fix this, but got distracted elsewhere

PR #342, opened 2026-04-16
ci: add release comment if extension is changed but manifest not upgraded

juleswg23:
Ideally, this makes it easier for people to avoid the mistake I made in #335 of not releasing.

This CI change will add a single comment to the PR listing all extensions that are changed, but which are not set to release.

I've made updates to README files in 2 extensions to verify the behavior, which I'll remove before merging.


dotNomad (commented):
I really like this, but I think we can simplify it quite a bit!

My changes are a bit hard to suggest via a code review so I'll open a pull request off of this one.

dotNomad:
Opened https://github.com/posit-dev/connect-extensions/pull/343 targeting this branch. Let me know what you think

juleswg23:
> Opened #343 targeting this branch. Let me know what you think

This is great! Feel free to merge into this branch.

Since it now covers both cases, WDYT about me updating the docs and removing the other location where the will/won't release info is outputted (in checks -> Extensions workflow -> release summary)?.

dotNomad:
> WDYT about me updating the docs and removing the other location where the will/won't release info is outputted (in checks -> Extensions workflow -> release summary)?.

That sounds great. Let's do that as a follow-up to this since this is encapsulated.

We may want to also consider porting the `0.0.0` version warning "⚠️ Version 0.0.0 is reserved and will never be released." to the comment

I'm going to remove the two README changes and get this merged up.

dotNomad (approved):
Thanks for this!

PR #343, opened 2026-04-16
Simplify release-check pipeline and expand PR comment

dotNomad:
## Summary
Proposed simplifications on top of #342:
- Consolidate the `should_release` step in `release-extension/action.yml`: single message/output/JSON emission at the end of the step instead of the 3-branch `write_release_check` helper. Uses `jq -n --arg/--argjson` for safer JSON construction than hand-built string concatenation.
- Gate the release-check artifact upload at the action level (`pull_request && MANIFEST_VERSION != '0.0.0'`), so the downstream comment job doesn't need to filter.
- Merge the aggregate + comment workflow steps into one. The intermediate `no_release.json` file and duplicate empty-check are gone.
- Expand the PR comment to summarize both will-release and won't-release extensions in separate sections — so contributors can also see what *will* release on merge, not just flag the mistakes.

### DIFF
--- a/.github/actions/release-extension/action.yml
+++ b/.github/actions/release-extension/action.yml
@@ -46,18 +46,9 @@ runs:
         echo "# Extension: ${{ inputs.extension-name }}" >> $GITHUB_STEP_SUMMARY
         echo "" >> $GITHUB_STEP_SUMMARY
 
-        # Helper to write release check artifact
-        write_release_check() {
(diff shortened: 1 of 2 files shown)


juleswg23 (approved):
This looks fantastic, thanks! I especially like including the "will release" info in the comment.

dotNomad:
> This looks fantastic, thanks! I especially like including the "will release" info in the comment.

Fantastic I'll merge it into your branch

PR #345, opened 2026-04-20
fix: update Content Health Monitor setup instructions for new Connect UI

### LINKED ISSUE #344: Content Health Monitor setup instructions reference outdated 'Vars' tab
## Summary

The Content Health Monitor setup instructions displayed when the `MONITORED_CONTENT` environment variable is not set reference a **"Vars" tab** that no longer exists in the Connect UI. Environment variables are now configured under the **"Advanced" tab** in the new settings drawer.

## Steps to reproduce

1. Deploy the Content Health Monitor from the Connect Gallery
2. View the deployed content without setting up `MONITORED_CONTENT`
3. Read the setup instructions in **Step 2: Configure this report**

## Current behavior

The instructions say:
> • Click the **Vars** tab

And reference a `vars.png` screenshot showing the old standalone Vars tab UI.

## Expected behavior

Instructions should say something like:
> • Click the **Advanced** tab
> • In the **Environment Variables** section, enter the variable

And the screenshot should show the current Advanced tab with the Environment Variables section.

## Details

The Connect UI was redesigned:
- The old standalone "Vars" tab was removed
- Environment variables are now in the **Advanced** tab of the settings drawer (shipped in Connect via PR #35948, Jan 2026)
- The settings drawer was enabled by default in March 2026

The affected code is in `extensions/content-health-monitor/content_health_utils.py` in the `get_env_var()` function, specifically the `MONITORED_CONTENT` instruction block. The screenshots in `extensions/content-health-monitor/images/` (e.g., `vars.png`, `settings-gear-icon.png`) likely also need updating.

## Screenshots

Setup instructions (left) vs. actual Connect UI (right):

The instructions point to a "Vars tab" that doesn't exist — the environment variables section is now under "Advanced."

ChaitaC:
## Summary

The Connect UI was redesigned — the standalone "Vars" tab was replaced with an "Advanced" tab containing an Environment Variables section, and toolbar buttons are now labeled. The Content Health Monitor's setup instructions and screenshots were outdated.

### Changes
- **"Vars" tab** → **"Advanced" tab**
- **"gear icon"** → **"Settings"** (now a labeled button)
- **"Add Variable"** → **"+ Add"**
- **"Refresh Report"** → **"Refresh"**
- **"View Logs"** → **"Logs"**
- Replaced old screenshots (`vars.png`, `settings-gear-icon.png`) with new ones (`env-vars.png`, `settings-button.png`, updated `refresh-report.png`)
- Updated `manifest.json` with new file entries and checksums

Closes #344

## Note

There are similar (minor) stale UI references in other extensions — will file a separate cleanup issue for those.


jonkeane (commented):
We could just change all of these to new settings (might be another nudge to get folks to upgrade "look at all this new stuff"), but if this is rendering on the server, we should be able to detect which version of connect someone is on and show the old settings if they are on an older version and the new ones if they are on a newer one, yeah?

ChaitaC:
> We could just change all of these to new settings (might be another nudge to get folks to upgrade "look at all this new stuff"), but if this is rendering on the server, we should be able to detect which version of connect someone is on and show the old settings if they are on an older version and the new ones if they are on a newer one, yeah?

Good idea! Version detection is possible via the /v1/server_settings API (it returns a version field). The extension already has a Connect client, so we could query it during setup.

The tricky part is that even within the same Connect version, individual users can toggle between old and new settings via the preview feature (userPreviewNewSettings), so server version alone doesn't tell us which UI they're seeing.

Since minimumConnectVersion is already 2025.04.0 and the new drawer defaults to on, I think targeting the new UI is the right call here. The old settings pane will eventually be removed, and this gives users on the old UI a nudge to switch over.

extensions/content-health-monitor/manifest.json
karawoo: this is the place to update in order to make sure your changes get released

karawoo (commented):
I had the same thought as @jonkeane initially but I'im ok with the simpler approach of just showing the new UI.

ChaitaC:
Thanks @karawoo for the review and the version bump catch! I know you and @jonkeane were both leaning towards version detection, so I tried to implement it before settling on just targeting the new UI.   

 Here's a summary of the options I explored — I've pushed Option 2 as a prototype, but my lean is Option 1. Happy to keep Option 2 if you all think the version detection is worth it.

**Option 1:** New UI only (current/initial PR approach)
Replace all old UI references with the new UI.

✅ Simple, no additional logic
⚠️ Users on Connect < 2026.03.0 (down to our minimum supported 2025.04.0) will see instructions that don’t match their UI

**Option 2:** Server version detection (prototype implementation for evaluation)
Check client.version — if ≥ 2026.03.0 show new UI instructions, otherwise show old.

✅ No extra API calls (SDK already fetches /server_settings)
✅ Works for the vast majority of users
⚠️ Edge cases:
2026.02.x users who manually opted into new UI
2026.03.0+ users who opted out
⚠️ Requires maintaining both old + new screenshots/instructions (temporary overhead)

**Option 3:** User preference detection (GET /me)
Check preferences.user_preview_new_settings.

✅ 100% accurate
❌ Extra API call
❌ Outside v1 API surface
❌ Added complexity for relatively small edge cases

ChaitaC:
Closing in favor of #347 which takes a generic approach across all affected extensions. https://github.com/posit-dev/connect-extensions/pull/347#discussion_r3188662406 The version-detection prototype is preserved in this branch if we ever want to revisit.

PR #352, opened 2026-05-06
Update usage metrics dashboard dependencies

### LINKED ISSUE #350: Update Rcpp for usage metrics dashboard
The usage metrics dashboard requires R ~=4.3:

https://github.com/posit-dev/connect-extensions/blob/62c3b93ab03ae2b5ac532057709f85492fbcf875/extensions/usage-metrics-dashboard/manifest.json#L14

It also has Rcpp version 1.0.14:

https://github.com/posit-dev/connect-extensions/blob/62c3b93ab03ae2b5ac532057709f85492fbcf875/extensions/usage-metrics-dashboard/manifest.json#L143-L149

Rcpp 1.0.14 is not compatible with R 4.6.0, but a Connect server with 4.6.0 will use it for this content. Rcpp will then fail to install.

The latest Rcpp (`1.1.1-1.1`) is listed as compatible with R >=3.5.0, so I think we can bump Rcpp in the manifest and keep support for R 4.3+.

karawoo:
Fixes #350 and then some.

I started by updating just Rcpp, but deploying to our internal server kept failing on various other dependencies that it tried to install from source on R 4.6.0. Eventually I gave up and updated everything, including renv.

I have tested deploying this with `rsconnect::deployApp(manifestPath = "manifest.json")` and forcing R 4.3 as well as allowing 4.6. The app worked under both scenarios, so I did not change the `"requires"` field.

I also noticed our CONTRIBUTING.md was a bit sparse on how to update the extension, so I added a few notes there.


extensions/usage-metrics-dashboard/renv/activate.R
```diff
@@ -2,7 +2,8 @@
 local({
```
jonkeane, line 2: These are all updates from renv and not something we need to review, correct?

extensions/usage-metrics-dashboard/CONTRIBUTING.md
```diff
@@ -9,3 +9,26 @@
 
 - Run `make test` to run tests.
 - Tests live in `tests/testthat`, and run against the code in `R/`.
+
+## Updating the manifest
+
+If updating package versions, update them and call `renv::snapshot()` first.
+Then update the manifest:
```
jonkeane: `update them` can I do this with `install.packages()`? With pak? Do I have to use an renv thing for this? It might be nice to say what has worked in the past.

It's minor but what about: 

```
If updating package versions, first update them (e.g. with `install.packages()`), then call `renv::snapshot()`, and then update the manifest:
```

My brain got a little confused with the `first` there at the end I was like "but there were two things? whaaaa?"

jonkeane (approved):
Thanks for this! A wording suggestion (well two), and then a question but this looks good

extensions/usage-metrics-dashboard/renv/activate.R
karawoo: yes, they came from bumping the renv version we are using

PR #355, opened 2026-05-29
switch to with-connect for integration tests

### LINKED ISSUE #353: Update integration testing to use `with-connect`
We should be able to simplify our integration testing setup by using [with-connect](https://github.com/posit-dev/with-connect).

As part of this, we should test against a more up-to-date matrix of Connect versions than we currently do:

https://github.com/posit-dev/connect-extensions/blob/62c3b93ab03ae2b5ac532057709f85492fbcf875/integration/Makefile#L69-L74

Also consider adding nightly or weekly runs so that we catch issues with upcoming Connect versions even when the content in this repo hasn't been updated recently.

karawoo:
moves integration tests to use with-connect and adds a weekly scheduled run so that we catch failures on upcoming connect releases even if we're not actively pushing code to this repo

closes #353 

the Connect containers used by `with-connect` have specific R and Python versions that are different from the previous integration testing setup, so getting tests to pass required a fair amount of dependency work (and sorry for the size of this diff):

- updated Python packages to versions with pre-built wheels for the Python versions in the containers. This raises the minimum Python version for some extensions to >= 3.11, since pandas 3.0.3 (the lowest version with Python 3.14 wheels) requires it.
- updated R packages to newer versions compatible with R 4.6.0 (which had significant changes to R's C API)
- small code adjustments to work with the newer package versions

@amylin1249 there will probably be some merge conflicts between this and some of your other in-progress PRs so we'll have to sort that out when they start merging.


.github/workflows/extensions.yml
```diff
@@ -58,6 +63,23 @@ jobs:
             chat-with-content: extensions/chat-with-content/**
             pqr: extensions/pqr/**
 
+      # When infra changed, test all simple extensions; otherwise use the filter output
```
dotNomad, line 69: Smart inclusion here for testing this.

.github/workflows/weekly-integration-tests.yml
```diff
@@ -0,0 +1,116 @@
+name: Weekly Integration Tests
+
+# Runs integration tests for ALL extensions against all compatible Connect
+# versions. Catches regressions from upcoming Connect releases even when
+# extension code hasn't changed.
```
dotNomad, line 5: Very nice inclusion 🎉

.github/workflows/weekly-integration-tests.yml
```diff
@@ -0,0 +1,116 @@
...
+
+jobs:
+  # Discover all extensions that have a manifest.json
+  discover-extensions:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      extensions: ${{ steps.list.outputs.extensions }}
+    steps:
+      - uses: actions/checkout@v4
```
dotNomad: Probably could bump this to the latest v6

.github/workflows/weekly-integration-tests.yml
```diff
@@ -0,0 +1,116 @@
...
+          extension-name: ${{ matrix.extension }}
+
+      - uses: ./.github/actions/package-extension
+        with:
+          extension-name: ${{ matrix.extension }}
+
+      - name: Mark successful package
+        run: echo "${{ matrix.extension }}" >> ${{ matrix.extension }}.packaged.successfully
+
+      - uses: actions/upload-artifact@v4
```
dotNomad, line 56: Similar note here. I think we could generally update to the latest actions for all the new code here to avoid needing to do it after.

dotNomad (commented):
Having reviewed the non-extensions bit so far I think the only thing I'm seeing is that we should up the GitHub action versions and see if CI still passes without issue just to be sure we aren't missing some param name change or something.

.github/workflows/extensions.yml
karawoo: it does have the side effect of reporting that extensions have been "changed" (they haven't) but won't be released. 

image

I don't think it's a big deal, but we could try to refactor if it gets confusing.

dotNomad: I don't think that's a huge deal. Like you said we can adjust in the future if need be.

dotNomad (commented):
`actions/download-artifact` is also out of date, but maybe we just add a follow-up issue to update the GitHub actions after this merges. There are a lot of out of date action calls in the repo generally.

Taking a look at the content changes now

dotNomad (approved):
Looked over the content changes and smoke tested the ones with code changes including `simple-mcp-server` and `portfolio-dashboard` using the CI artifacts. Everything is working 🎉

PR #364, opened 2026-06-04
Update stock report examples

### LINKED ISSUE #357: Update stock-report to be more like quarto-stock-report-*
The Quarto version of this example uses cached data (instead of pulling live every time) and a different stock ticker.

amylin1249:
Fixes #357 

Changes as prescribed in the issue:
- Switched the reported ticker to GSPC (S&P 500)
- Switched from live data fetching to reading cached gspc.csv (matching the quarto-stock-report-* approach as described in the issue)

Other changes made:
- Removed all TSLA references
- Bumped version 1.0.0 → 1.0.1 and added a CHANGELOG.md
- Manually checked outputs for each to identify two edits for readability: 1) in R Markdown version, add padding to the table in the email so the numbers aren't squished; 2) in Jupyter notebook version, remove stray <Axes: ...> text that printed above the chart

Verification:
As mentioned above, manually checked final outputs of examples to see that intended changes were made and that everything is readable for the user.


extensions/stock-report/stock-report.Rmd
```diff
@@ -17,44 +17,59 @@ library(ggthemes)
...
-  )
 ```
 
 The stock closed `r ifelse(close>open,'up','down')` at `r close` dollars per share yesterday.
 
 
 ### Price History
 
-The chart below is made with the `quantmod` and `plotly` R packages. An API returns all of the price history based on the stock tick symbol "TSLA.".
+The chart below is made with the `quantmod` and `plotly` R packages. An API returns all of the price history based on the stock tick symbol "GSPC.".
```
karawoo: This is no longer pulling from an API, so I think the text should update (unless we're just trying to pretend it's hitting an API for demo purposes?)

extensions/stock-report/stock-report.Rmd
```diff
@@ -17,44 +17,59 @@ library(ggthemes)
...
-
-prices_df <- prices %>%
-  as_tibble() %>%
-  mutate(
-    Date = index(prices), 
-    Direction = ifelse(TSLA.Close >= TSLA.Open, 'Increasing', 'Decreasing')
-  )
 ```
 
 The stock closed `r ifelse(close>open,'up','down')` at `r close` dollars per share yesterday.
```
karawoo: Since we're now using a market index instead of a share price I think this text should be updated, `dollars per share` is no longer accurate.

extensions/stock-report/stock-report-email.Rmd
karawoo: while we are making improvements here we might want to use the column name instead of index

karawoo (commented):
Overall looks good, a few comments below. Also, I know the original issue says we should use cached data instead of hitting the API anymore. This does somewhat change the app from its described purpose though:

Image 

We're no longer showing people how to automate regular updates. Should we change that description?

extensions/stock-report/stock-report-email.Rmd
amylin1249: Thank you for all of these comments, very helpful! I made updates addressing these 4 and found a few other spots where text/code should be updated to reflect the switch from an individual stock to a stock index. For the email, I changed the trigger in the R version to 1% (as a significant change), since a 50 cent move (what we had previously) would almost always be triggered with GSPC. I also updated all of the language around dates (yesterday/today) to be more accurate given our use of cached data. Finally, I updated the calculations in the summary and email for consistency -- both now use day-over-day change. 

Note, I realized we probably need to make a couple of wording updates in the quarto files as well (in line with the comments you've made here), so I will open and issue/PR for that soon.

karawoo (approved):

PR #367, opened 2026-06-05
Modernize codebase

### LINKED ISSUE #365: Modernize codebase
Apply changes across the repo to modernize, which can include:
  - Update pipes from `%>%` to `|>`
  - Replace `httr` with `httr2`
  - Update deprecated dplyr/tidyr verbs (e.g. `mutate_at`/`gather` → `across`/`pivot_*`)
  - Use `vapply`/`purrr` over `sapply`
  - Python: `os.path` → `pathlib`
  - Python: `.format()`/`%` → f-strings
  - Python: modernize `typing` (only where the language version supports it)
  - Skip generated files (e.g. `renv/activate.R`)

amylin1249:
Fixes #365 

Replaced outdated code practices (e.g., %>% to |>, .format to f-strings), removed unused code/comments, and simplified code blocks to improve clarity while preserving all functionality. 

Applied version bump, new checksums, and new CHANGELOG (where missing) for all extensions.


dotNomad (commented):
The native R pipe `|>` was introduced in R version `4.1.0`.

Checking just `connectwidgets-example` I can see that the `manifest` requires only R `~=4.0`.

We try our best to support the most versions of the languages as possible (back to the minimum versions that Connect supports of course).

The minimum for Python is `3.8` so all good to include f-strings there.

References:
- https://cran.r-project.org/doc/manuals/r-devel/NEWS.html
- https://dogfood.team.pct.posit.it/__docs__/admin/python/#python-install

---

It may be helpful to split up this PR a bit for ease of reviewing.

amylin1249:
Good catch, thanks for pointing out the version requirements. I went back through and checked, and here's what I found: 
- connectwidgets-example is the only published extension that declares R ~=4.0, so you are right that |> would break it. I reverted it back to %>% on my most recent commit.
- prototype-content-with-issues-table was also changed to |> (as well as other modernization changes) but has no declared minimum because it has no manifest.json. Technically, it might need reverting, but it's also an unreleased prototype that isn't in the gallery and likely is something I will either remove or upgrade to something more useful for a user. For now I did not revert the changes but can do that if you think I should.

Separately, I think connectwidgets-example is older (last released in 2023), so it may be removed or replaced/upgraded this summer anyways. But for now, I agree I would want it to support as many versions as possible while it is available, so I reverted the pipe change.

PR #369, opened 2026-06-08
Update quarto stock reports to reflect cached data update

### LINKED ISSUE #368: Update quarto-stock-report-* to accurately describe new cached data workflow
When working on issue #357 (PR #364 ) , a need to update the language in order to better describe the new cached data workflow was surfaced. We also need to make relevant updates to the two quarto files.

amylin1249:
Fixes #368 by changing calculations to day-over-day-change, keying dates off the latest data, and updating text to align with market index (instead of individual stock) and cached data (instead of API).


dotNomad (commented):
Should this be a change for the PR #364 or is this independent?

amylin1249:
It is sort of independent. Initially, someone before me updated the quarto reports to use cached data, but they did not update all of the accompanying text/language with it. Then, I was tasked with updating the stock reports to use a similar method as the quarto reports (that is #364 ). In the process of working on that PR, I realized that the quarto reports actually needed the aforementioned updating, so that is this PR.

karawoo (approved):

PR #372, opened 2026-06-09
Fix version incompatibility

### LINKED ISSUE #371: Fix version incompatibility / runtime error in publisher-command-center
While working on #365 (generally updating the codebase to modernize it, which included changes like: switching to |> and f strings, updating to use pathlib), I double-checked each change against the extension's declared minimum R/Python version.

In doing so, I surfaced a pre-existing bug in publisher-command-center: 

https://github.com/posit-dev/connect-extensions/blob/d6cfcac6f2c7865d9286994393ea08e8a2ab0f10/extensions/publisher-command-center/app.py#L72

The "str | None" syntax only exists in Python 3.10+, but the manifest declares python.requires: ~=3.8. 

This means that if a user is on Python 3.8 or 3.9, they will get a TypeError at startup.

amylin1249:
Fixes #371 

Changes from `token: str | None` to `token: Optional[str]`, which is compatible with 3.8+

### DIFF
--- a/extensions/publisher-command-center/CHANGELOG.md
+++ b/extensions/publisher-command-center/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to the Publisher Command Center extension will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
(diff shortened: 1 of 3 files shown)


dotNomad (approved):
Incredible catch 🎉

PR #375, opened 2026-06-11
gitignore deployment directories

karawoo:
When developing in this repo I often deploy extensions from my local machine to Connect to test things out. Then I end up with various deployment files appearing as untracked files in git. For this repo there's never any need to commit those, so this PR adds them to `.gitignore`.


amylin1249 (approved):
Makes sense, looks good!

PR #376, opened 2026-06-12
fix stale manifests and standardize > to ~ for runtime requirements

### LINKED ISSUE #366: Fix stale extension manifest metadata
This issue is for some standard cleanup to the repo. When working on other various issues, Claude flagged that there were stale checksums and use of ~= instead of >= for runtime requirements in some of the manifest.json files. 

The scope of this issue is limited to mechanical, low-risk fixes -- there are no behavior changes, and since none bump a version, nothing will re-release.

- Stale checksums — recompute md5 per file and fix manifest entries that have drifted from their committed source.
- requires normalization — convert >= → ~= where needed.

amylin1249:
Fixes #366 exactly as described. In relevant manifest.json files:
- Fixes stale checksums
- Since nearly all of the extensions use ~= instead of >= for runtime requirements, I updated the few that still had >=

Neither of these affects functionality:
- The checksum fix only updates metadata to reflect files already in the repo; the file contents are untouched
- Converting >=X → ~=X only narrows the range by adding an upper bound at the next major version (e.g. >=3.11 → ~=3.11 means >=3.11, <4.0). It excludes nothing that exists today, it just stops the extension from silently claiming compatibility with a future major (Python 4, R 5, …)

### DIFF
--- a/extensions/audit-reports/manifest.json
+++ b/extensions/audit-reports/manifest.json
@@ -34,7 +34,7 @@
       "checksum": "2bafb90cfc5b4933ee0ef8d4d07ff5b2"
     },
     "runtime-versions.qmd": {
-      "checksum": "5fc1056d0e0e8b89721cc7508a635eff"
+      "checksum": "1fd88936f3d435918be3ef2adacbda0c"
     },
     "unpublished.qmd": {
       "checksum": "dd2d6de0ce972611a454e80776c4ed3d"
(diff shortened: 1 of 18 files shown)


dotNomad (approved):
Great adjustment 👍 

Should we be releasing these since they update the manifests? It does change how they deploy in the case described, new language versions come out.

amylin1249:
Yes, I think you're right, we do need to re-release these for the version changes to take effect. I updated in the most recent commit.

dotNomad:
> Yes, I think you're right, we do need to re-release these for the version changes to take effect. I updated in the most recent commit.

Awesome. Taking another look

dotNomad (commented):
Hmm interesting the `reaper` extension is now failing our tests with an error about needing a dependency. Any idea what that could be @amylin1249?

dotNomad (approved):
Look awesome 😎

PR #391, opened 2026-06-22
fix(runtime-version-scanner): parse error in content table

### LINKED ISSUE #379: runtime-version-scanner: owner_name crash when content owner has null/absent name fields
## Bug

In `extensions/runtime-version-scanner/app.R` (around line 263), building `owner_name` errors when a content owner has a null or absent `first_name` or `last_name`:

```r
owner_name = paste(
  map_chr(owner, "first_name"),
  map_chr(owner, "last_name")
),
```

### Error

```
Caused by error in `mutate()` at mnt/app/app.R:263:7:
ℹ In argument: `owner_name = paste(map_chr(owner, "first_name"), map_chr(owner, "last_name"))`.
Caused by error in `map_chr()`:
ℹ In index: 1.
! Result must be length 1, not 0.
```

### Cause

`map_chr()` requires each extracted result to be exactly length 1. When a content item is owned by an account whose profile lacks `first_name` or `last_name` (e.g. a service/API account, a deleted user, or a user synced from LDAP/SAML without name fields populated), the plucked value is `NULL` (length 0), which raises the error and prevents the content table from populating.

### Reproduction

- Connect 2026.05.0, RHEL 9, R 4.4.3
- At least one content item owned by an account with no `first_name` or `last_name` set
- Enable and run the Runtime Version Scanner

### Suggested fix

Add `.default = ""` to both `map_chr()` calls so missing name fields substitute an empty string instead of returning length 0:

```r
owner_name = paste(
  map_chr(owner, "first_name", .default = ""),
  map_chr(owner, "last_name",  .default = "")
),
```

### Source

Reported via Posit Support ticket #130420.

joshyam-k:
fixes #379 

I started by deploying a version of this app with a diagnostic log on the content owner firstname/lastname to dogfood. That log showed that the entire parsed content ownership table was empty. 

### The actual bug:

`content_list_to_data_frame()` passed the result of `search_content()` straight into `connectapi`'s internal parser:
```
connectapi:::parse_connectapi_typed(content_list, connectapi:::connectapi_ptypes$content)
```

But `search_content()` returns a list of R6 Content objects whose fields live in `$content`. The internal parser can't read fields off R6 objects, so every column came back empty. `map_chr(owner, "first_name")` was just the first line to error on the empty owner.

`connectapi`'s own converter does the step the scanner skipped:
```
as_tibble.connect_content_list <- function(x, ...) {
    content_data <- purrr::map(x, "content")   # <-- extract $content first
    parse_connectapi_typed(content_data, connectapi_ptypes$content)
}
```
the fix is just to to rely on this method

### validation

Deployed the app with this fix:

Screenshot 2026-06-22 at 9 13 45 AM

### DIFF
--- a/extensions/runtime-version-scanner/R/connect_module.R
+++ b/extensions/runtime-version-scanner/R/connect_module.R
@@ -38,12 +38,10 @@ connectVisitorClient <- function(
           )
         } else if (nrow(selected_integration) == 0) {
           integration_settings_url <- publisher_client$server_url(
-            connectapi:::unversioned_url(
-              "connect",
-              "#",
-              "system",
-              "integrations"
-            )
+            "connect",
+            "#",
+            "system",
+            "integrations"
           )
           message <- paste0(
             "This content needs permission to ",
@@ -128,12 +126,14 @@ get_eligible_integrations <- function(client) {
 
 auto_add_integration <- function(client, integration_guid) {
   client$PUT(
-    connectapi:::v1_url(
+    paste(
+      "v1",
       "content",
       Sys.getenv("CONNECT_CONTENT_GUID"),
       "oauth",
       "integrations",
-      "associations"
+      "associations",
+      sep = "/"
     ),
     body = list(list(oauth_integration_guid = integration_guid))
   )
--- a/extensions/runtime-version-scanner/manifest.json
+++ b/extensions/runtime-version-scanner/manifest.json
@@ -2616,13 +2616,13 @@
   },
   "files": {
     "app.R": {
-      "checksum": "69ec0029acb50cb4bc05041bb35b88b8"
+      "checksum": "1f458d627fc174c123e24baa3a3934cf"
     },
     "R/connect_module.R": {
-      "checksum": "06b96d31139208ae8e231a4b9991e800"
+      "checksum": "b06c872e41c16ee85834394daccfc19d"
     },
     "R/fetch_content.R": {
-      "checksum": "1563f56cdbf699953ee2750fac9b560f"
+      "checksum": "a99f3e21b2612e9d889b14b8a6201d83"
     },
     "R/supported_versions.R": {
       "checksum": "ca8c4181bddf00f8e7e0797671ae01c6"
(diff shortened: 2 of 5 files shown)


extensions/runtime-version-scanner/manifest.json
amylin1249: Stale checksum that needs to be updated. This one should be: 1f458d627fc174c123e24baa3a3934cf

amylin1249: Stale checksum that needs to be updated. This one should be: a99f3e21b2612e9d889b14b8a6201d83

amylin1249 (commented):
Nice fix, fully functional. Two things before merging:

1. The manifest checksums are stale - I left inline comments for those.
2. CHANGELOG entry - in making updates to the Gallery moving forward, I'm trying to make sure we are consistent in adding CHANGELOG entries/doing an overall better job of documentation. We'll want to add a dated entry that describes what was changed/fixed, as well as references the PR number. For this fix, we could do:

```markdown
## [1.0.5] - 2026-06-23

### Fixed

- Fixed the content table failing to load (and the resulting `map_chr` crash) when scanning content. The content list was being parsed directly off connectapi's R6 objects, which left every column empty; it now parses via `as_tibble()`, and missing owner names fall back to an empty string. (#391)
```

FYI, I use Claude to help craft the entry (easier to format correctly, pull the right date/PR #, etc.) and then tweak the text as needed.

Finally, this isn't necessarily for this specific issue/PR, but I wanted to flag that R/connect_module.R still has two connectapi::: internal calls (unversioned_url at line 41, v1_url at line 131). This is the same fragility you just removed from fetch_content.R. Both are unexported connectapi internals, so a future connectapi update could rename or drop them and silently break the scanner. So it could be a good idea to be proactive and fix this there as well (to avoid breaks in the future).

joshyam-k:
Thanks @amylin1249! I incorporated all of your suggestions.

extensions/runtime-version-scanner/R/connect_module.R
```diff
@@ -38,12 +38,10 @@ connectVisitorClient <- function(
           )
         } else if (nrow(selected_integration) == 0) {
           integration_settings_url <- publisher_client$server_url(
-            connectapi:::unversioned_url(
```
joshyam-k, line 41: `server_url` already accepts multiple path segments and joins with `/`

amylin1249 (approved):

PR #392, opened 2026-06-22
Polish script pair

### LINKED ISSUE #390: Polish script-r / script-python
Non-exhaustive list of changes: 

script-r:
- Rewrite title and description
- Rewrite the README: add the scheduling setup (the whole point) and the email/SMTP admin prerequisite, plus a "how to adapt" (swap `palmerpenguins` for a real source). Fix Quarto/R version wording.
- Add explanatory comments
- Fix CSV index column: `write.csv(..., row.names = FALSE)`.
- Create `CHANGELOG.md` (missing). Bump 1.0.1 → 1.0.2. Recompute md5 for every edited file in manifest `files`.

script-python
- Rewrite title and description
- Rewrite the README: scheduling + the email feature (which exists but was dropped from the description/docs); how to adapt. Align the Python-version claim.
- Add explanatory comments
- Fix CSV index: `to_csv(..., index=False)`.
- Reconcile the 3-way Python split: manifest pins build 3.8.20, declares `requires ~=3.9`, and the checked-in `script.quarto_ipynb` carries a leaked `/Users/christophergraham/...` 3.11 kernel path. Pick one runtime, re-pin, and clean or drop the bundled ipynb.
- Create `CHANGELOG.md` (missing). Bump 1.0.1 → 1.0.2. Recompute md5 for every edited/re-rendered file in manifest `files`

amylin1249:
Fixes #390 

This can be reviewed commit-by-commit.

both files:
- titles, descriptions, and README updated to be more functional for the user
- export the CSV files without the leading row-index column
- add one comment marking the email-body output block (more clear for the user)
- add CHANGELOGs (previously did not exist)

`script-python`:
- drop `script.quarto_ipynb`, a committed Quarto build intermediate that needed to be gitignored
- fix Python-compatibility bug: pinned `ipykernel==7.2.0` requires Python ≥3.10, contradicting the declared `requires: ~=3.9` and breaking install on 3.9 servers. Re-pinned to a minimal 3.9-compatible set, trimmed the unused interactive Jupyter packages, and set `python.version` to the verified `3.9.6`. Floor stays `~=3.9` for broad compatibility.


dotNomad (commented):
Hmm this is a bit odd. I tried out publishing both of the `tar.gz` artifacts that were created. Those give a good indication of how the content will behave when deployed via the Gallery and for both I'm seeing the Changelog rather than the script in Connect.

Doing some investigation... 🕵️

dotNomad:
We can fix this with a `_quarto.yml` that specifies rendering the scripts. The reason why this happened is `quarto inspect` listed `CHANGELOG.md` first as an entrypoint option. `README.md` isn't considered which is why this bug didn't hit us with the `README.md` files.

I opened https://github.com/posit-dev/connect-extensions/pull/399 to show that the CI passes, and the `tar.gz` bundles are there to test deployment. I think if we get that change in this PR it should work again.

amylin1249:
All done, used what you had in #399 which we can close now. 

I also made a couple final updates to finishing polishign the pair: 
- Trimmed script-python's deploy bundle — dropped .gitignore and README.md from manifest.files, and added a parallel .rscignore so it matches script-r.
- Removed the inert _quarto.yml project title
- Aligned each README H1 with its gallery title (Framework: Functionality)
- Minor tidies — matched a .gitignore comment across the two, and dropped a stale thumbnail.jpg line from script-r's .rscignore.

extensions/script-python/manifest.json
```diff
@@ -6,12 +6,12 @@
   },
   "extension": {
     "name": "script-python",
-    "title": "Run a Python Script",
-    "description": "This data transformation Python script is an example of how you might automate regular imports and transformations from a data source. JSON and CSV files are exported and made available for download. Quarto is used to render the script.",
+    "title": "Python: Scheduled Data Transformation",
```
dotNomad: I worry a bit about this title change - perhaps we should discuss it.

When looking through the Gallery examples I know I'm looking for an answer to the question "How do I run a Python script?" and the title here is that answer.

The title "Python: Scheduled Data Transformation" is more accurate on what this specific example does, and the description even more so, but I worry that the swap will make the answer to the script question harder.

Reference: https://docs.posit.co/connect/user/scripts/index.html

amylin1249: What do you think of "Python: Scheduled Script" and "R: Scheduled Script" ?

dotNomad: Thats a good middle ground 👍

amylin1249: Okay great, I made those updates.

dotNomad (approved):
This looks great :shipit:

PR #398, opened 2026-06-22
Fix weekly integration failures

karawoo:
The [weekly run](https://github.com/posit-dev/connect-extensions/actions/runs/27939082029) failed for five extensions. This PR aims to fix them.

* `package-vulnerability-scanner` and `publisher-command-center`: both mount a built `dist/` that is gitignored, but the weekly workflow tarred the directory without building it. The workflow now runs `npm run build` for any extension with a `package.json` before packaging.
* `stock-api-flask`, `stock-dashboard-python`, `pqr`: each pinned a `numpy`/`pandas` version with no Python 3.14 wheel, so on the 3.14-only image they were built from source and broke. Bumped to versions with 3.14 wheels.

### DIFF
--- a/extensions/pqr/CHANGELOG.md
+++ b/extensions/pqr/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Pinned `pandas>=2.3.3` so the bundled datasets load without crashing on Connect releases that ship Python 3.14. The previous `pandas==2.3.2` had no Python 3.14 wheel and built from source, producing a broken extension that segfaulted when loading data.
+- Pinned `pandas>=2.3.3` so the bundled datasets load without crashing on Connect releases that ship Python 3.14. The previous `pandas==2.3.2` had no Python 3.14 wheel and built from source, producing a broken extension that segfaulted when loading data. (#398)
 
 ## [0.1.2] - 2026-06-15
 
--- a/extensions/pqr/CHANGELOG.md
+++ b/extensions/pqr/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to the Quarto Document with Python and R extension will be d
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.3] - 2026-06-22
+
+### Fixed
+
+- Pinned `pandas>=2.3.3` so the bundled datasets load without crashing on Connect releases that ship Python 3.14. The previous `pandas==2.3.2` had no Python 3.14 wheel and built from source, producing a broken extension that segfaulted when loading data.
+
 ## [0.1.2] - 2026-06-15
 
 ### Changed
--- a/extensions/stock-api-flask/CHANGELOG.md
+++ b/extensions/stock-api-flask/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Updated `numpy` and `pandas` to versions with Python 3.14 wheels so the content runs on Connect releases that ship Python 3.14.
+- Updated `pandas` and loosened the `numpy` pin to a range so the content installs with Python 3.14 wheels on Connect releases that ship Python 3.14, while still running on older Python versions. (#398)
 
 ## [1.0.2] - 2026-06-11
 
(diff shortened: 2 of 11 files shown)


extensions/stock-dashboard-python/manifest.json
```diff
@@ -12,11 +12,11 @@
     "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/stock-dashboard-python",
     "category": "example",
     "minimumConnectVersion": "2025.04.0",
-    "version": "1.0.2"
+    "version": "1.0.3"
   },
   "environment": {
     "python": {
-      "requires": "~=3.9"
+      "requires": "~=3.11"
```
amylin1249: Could this stay ~=3.9 (so we can run on as many Connect versions as possible)? This extension only pins pandas==2.3.3, and pandas 2.3.3 is requires_python>=3.9, so I think it already installs and runs on 3.9-3.14 without bumping the floor.

extensions/pqr/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to the Quarto Document with Python and R extension will be d
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.3] - 2026-06-22
+
+### Fixed
+
+- Pinned `pandas>=2.3.3` so the bundled datasets load without crashing on Connect releases that ship Python 3.14. The previous `pandas==2.3.2` had no Python 3.14 wheel and built from source, producing a broken extension that segfaulted when loading data.
```
amylin1249: Thanks for adding to the CHANGELOG. Quick thing, could you add the PR ref, e.g. "...when loading data. (#398)" 

In making updates to the Gallery moving forward, I'm trying to make sure we are consistent in referencing our PRs with (#NNN) notation. This same change would apply to the CHANGELOG entries for `stock-api-flask` and `stock-dashboard-python`.

Also depending on what other changes are adopted from my remaining comments, we may need to tweak the CHANGELOG entry.

extensions/stock-api-flask/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to the Stock API using Flask extension will be documented in
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.3] - 2026-06-22
+
```
amylin1249, line 9: Add (#398) here too (see note on the `pqr` CHANGELOG).

Also depending on what other changes are adopted from my remaining comments, we may need to tweak the CHANGELOG entry.

extensions/stock-dashboard-python/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to the Stock Dashboard using Python extension will be docume
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.3] - 2026-06-22
```
amylin1249, line 8: Add (#398) here too (see note on the `pqr` CHANGELOG).

Also depending on what other changes are adopted from my remaining comments, we may need to tweak the CHANGELOG entry.

extensions/stock-api-flask/manifest.json
```diff
@@ -15,11 +15,11 @@
     "requiredFeatures": [
       "API Publishing"
     ],
-    "version": "1.0.2"
+    "version": "1.0.3"
   },
   "environment": {
     "python": {
-      "requires": "~=3.10"
+      "requires": "~=3.11"
```
amylin1249: Similar reasoning to my other comment about supporting more/older Connect versions, can you revert this to ~=3.10? I will note that this requires loosening to numpy>=2.2,<3. That's a tradeoff here, we can either lower the Python floor and change the numpy to a range instead of an exact pin (which would diverge from how the other API examples are written) for the sake of broader compatibility, or we can keep the changes that you made for the sake of overall consistency. I would lean towards broader compatibility.

extensions/stock-api-flask/requirements.txt
```diff
@@ -1,4 +1,4 @@
 flask==3.0.2
 flask-restx==1.3.0
-pandas==2.2.3
-numpy==2.1.2
+pandas==2.3.3
+numpy==2.3.3
```
amylin1249: If you end up reverting it to ~=3.10 (choice for broader compatibility), then we would need to update here to numpy>=2.2,<3.

amylin1249 (commented):
Fix is technically fully functional as is, just left some comments that would maintain the functional fix while allowing us to continue to support more/older versions of Connect, and a small CHANGELOG convention that we should try to follow moving forward.

amylin1249 (approved):

PR #401, opened 2026-06-23
feat(scripts): sort the gallery feed by title, not slug

### LINKED ISSUE #396: Update how content is ordered to alphabetical by title
Currently, the content is ordered by name (slugs) which is not displayed and can be confusing to a user (it appears random). 

We can change the content to be ordered by title (so that it looks alphabetical to a user).

amylin1249:
Fixes #396 

The gallery feed was sorted by `name` (the slug), which isn't shown to users, so card order looked random. Now it will sort by `title`. 

It also currently only triggers a re-sort when adding a new extension (requires manually re-sorting any time there is a title change to an existing extension), so I added that it will sort on every write in `stringify()`). Now, any release (a new extension or a title change to an existing one) reorders the whole feed automatically.

### DIFF
--- a/scripts/extension-list.ts
+++ b/scripts/extension-list.ts
@@ -42,12 +42,7 @@ class ExtensionList {
     public tags: string[],
     public requiredFeatures: RequiredFeature[],
     public extensions: Extension[]
-  ) {
-    this.categories = categories;
-    this.tags = tags;
-    this.requiredFeatures = requiredFeatures;
-    this.extensions = extensions;
-  }
+  ) {}
 
   static fromFile(path: string) {
     const file = JSON.parse(fs.readFileSync(path, "utf8"));
--- a/scripts/extension-list.ts
+++ b/scripts/extension-list.ts
@@ -178,8 +178,6 @@ class ExtensionList {
   }
 
   public stringify() {
-    // Always emit a title-sorted feed so any release (new or updated) reorders.
-    this.sortExtensions();
     const output = {
       categories: this.categories,
       tags: this.tags,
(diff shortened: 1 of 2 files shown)


scripts/extension-list.ts
```diff
@@ -179,6 +178,8 @@ class ExtensionList {
   }
 
   public stringify() {
+    // Always emit a title-sorted feed so any release (new or updated) reorders.
+    this.sortExtensions();
```
dotNomad: Instead of adding this to stringify, which just configures and calls `JSON.stringify`, what do you think about calling `sortExtensions` prior to `fs.writeFileSync(extensionListFilePath, list.stringify());` like so:

```
// Always emit a title-sorted feed so any release (new or updated) reorders.
list.sortExtensions();
fs.writeFileSync(extensionListFilePath, list.stringify());
```

We have to make `sortExtensions` `public` but that is fine. I think this is a bit clearer.

dotNomad (commented):
Great change. I have one suggestion to move code to make it a bit easier to see

scripts/extension-list.ts
amylin1249: Thanks for catching this, I updated it to make it more explicit. I also renamed it to sortExtensionsByTitle to make it more explicit/easy to differentiate from sortExtensionVersions. 

I also made a second commit that was some cleanup -- the ExtensionList constructor had redundant assignments by virtue of the params being public, so I removed them.

dotNomad (approved):
Lovely. The removal of the duplicate assignment for `ExtensionList` fields is nice too.

PR #412, opened 2026-06-26
standardize readmes, descriptions in script-r and script-python

amylin1249:
Follow up to #392 - updating READMEs, descriptions to follow standardized format to be implemented across Gallery

### DIFF
--- a/extensions/script-python/README.md
+++ b/extensions/script-python/README.md
@@ -3,7 +3,7 @@
 ## About this example
 
 This example is a short Python script that Quarto runs as a report. It reads the
-`palmerpenguins` dataset, filters out two groups (female Gentoo and male Adelie
+`palmerpenguins` dataset, filters the data down to two groups (female Gentoo and male Adelie
 penguins), and writes each group to a JSON file and a CSV file that viewers can
 download. It also sends an email with the two CSV files attached.
 
--- a/extensions/script-python/README.md
+++ b/extensions/script-python/README.md
@@ -2,7 +2,7 @@
 
 ## About this example
 
-This example is a short Python script that Quarto runs as a report. It reads the
+A short Python script that Quarto runs as a report. It reads the
 `palmerpenguins` dataset, filters the data down to two groups (female Gentoo and male Adelie
 penguins), and writes each group to a JSON file and a CSV file that viewers can
 download. It also sends an email with the two CSV files attached.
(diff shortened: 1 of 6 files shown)


extensions/script-python/README.md
```diff
@@ -24,18 +24,20 @@ The email and its attachments are set in the `format: email` block at the top of
 server. If none is configured, the script still runs and produces the files, but
 no email is sent.
 
-## Adapt this example
+## Customize it
```
zackverham, line 27: I like using "Customize" instead of "deploy" - should we standardize on when we use "it"?

I.e. should we also say "Deploy it"? 

Or just "Customize" and "Deploy"?

amylin1249: Nice catch, I think we can have "customize it" and "deploy it" across all of them - making updates here and to the other 5 examples being worked on.

extensions/script-python/README.md
```diff
@@ -24,18 +24,21 @@ The email and its attachments are set in the `format: email` block at the top of
...
 database query, an API call, or a file), then adjust the filters, the output file
 names, and the attachment list in the email block.
 
-## Learn more
+## Deploy it
 
-* [Quarto](https://quarto.org)
-* [Quarto: Rendering Script Files](https://quarto.org/docs/computations/render-scripts.html)
+Deploy it straight from the Connect Gallery to get a copy running, then customize
+it. To publish your own version, deploy the directory with
```
m--: This almost reads to me like someone will be able to customize the version they deployed to Connect, in Connect. We probably don't want to imply that exactly, but is there another link we could give users to bridge this gap? Thinking "I deployed straight from the Gallery and I want to update it now...how?"

extensions/script-r/README.md
```diff
@@ -24,18 +24,21 @@ The email and its attachments are set in the `format: email` block at the top of
...
 database query, an API call, or a file), then adjust the filters, the output file
 names, and the attachment list in the email block.
 
-## Learn more
+## Deploy it
 
-* [Quarto](https://quarto.org)
-* [Quarto: Rendering Script Files](https://quarto.org/docs/computations/render-scripts.html)
+Deploy it straight from the Connect Gallery to get a copy running, then customize
+it. To publish your own version, deploy the directory with
```
m--: same comment here

extensions/script-python/README.md
amylin1249: Updated for clarity and applied the change across all 7 examples currently being worked on.

m-- (approved):

PR #414, opened 2026-06-26
detect vulns in deployed versions, not just latest

### LINKED ISSUE #411: Vulnerability Scanner misses vulnerabilities in older package versions
## Summary

The Package Vulnerability Scanner extension uses the Posit Package Manager API endpoint `POST /filter/packages` with `has_vulns: True` to retrieve vulnerability data. However, this endpoint **only returns vulnerabilities for the most recent version** of a package unless explicit package versions are requested via something like `"names": [ "jupyterlab==4.5.8" ]`. This means the scanner silently misses vulnerabilities in older versions that users may have deployed on Connect.

## The Problem

1. The backend (`main.py` L50) calls `/filter/packages` → gets packages where the *latest version* has known vulnerabilities
2. The frontend cross-references this data against packages actually installed on Connect content
3. If a vulnerability existed in package X v1.0 but was **fixed in v2.0** (the latest), the API won't return that vulnerability at all
4. A user on Connect running v1.0 will **not be warned** about the vulnerability

This is the most actionable case for the scanner — "you're running an old version with a known vuln, upgrade to the latest which fixes it" — and it's exactly what gets dropped.

## Reproduction

Example from a customer report:
```bash
curl -X 'POST' '<PPM_URL>/__api__/filter/packages' \
  -H 'accept: application/x-ndjson' \
  -H 'Content-Type: application/json' \
  -d '{
  "has_vulns": true,
  "names": [ "jupyterlab" ],
  "omit_dependencies": false,
  "omit_downloads": true,
  "repo": "pypi"
}'
```
This returns an empty response even though older versions of `jupyterlab` have known vulnerabilities visible in the PPM GUI.

## Expected Behavior

The scanner should detect vulnerabilities in **all versions** of a package that users might have deployed, not just the latest version.

## Possible Solutions

- Use a different PPM API endpoint that returns vulnerabilities across all versions (if one exists)
- Query vulnerabilities per-version for the specific versions found installed on Connect content
- Coordinate with the Package Manager team to add an option to `/filter/packages` that returns vulns for all versions

## References

- Relevant code: https://github.com/posit-dev/connect-extensions/blob/main/extensions/package-vulnerability-scanner/main.py#L39
- Reported via internal Slack thread: https://positpbc.slack.com/archives/CCD5REGR4/p1782463133172299

amylin1249:
Fixes #411 

Probably easiest to review this one commit by commit.

Implemented option 2 from the issue (query vulnerabilities for the specific installed versions). The frontend already loads each content item's packages, so it gathers the exact `name==version` set installed across the content in view and POSTs it to `/api/vulns`, which asks Package Manager about those versions directly. Each returned vulnerability carries its full affected-version map, so the existing per-version matching is unchanged.

The backend stays a thin Package Manager proxy — it queries only the versions the frontend sends and never enumerates content itself — so the work scales with what the UI already loads rather than a separate server-side pass over all content.

I also adjusted the UI: the vulnerability counts (the total, the "With Vulnerabilities" tab, and each content card's badge) now wait for the Package Manager lookup to finish before showing a status, instead of briefly showing "no vulnerabilities" while it's still in progress.

Verified on a Connect instance: content pinned to old, since-patched packages (`urllib3==1.26.4`, `jinja2==2.11.3`, `werkzeug==2.0.0`) now reports ~30 vulnerabilities where the previous scan reported 0.

### DIFF
--- a/extensions/package-vulnerability-scanner/src/components/ContentList.vue
+++ b/extensions/package-vulnerability-scanner/src/components/ContentList.vue
@@ -108,7 +108,7 @@ const tabs = computed<Tab[]>(() => {
   result.push({
     name: "With Vulnerabilities",
     color: "error",
-    count: scannerStore.anyContentLoadingPackages
+    count: scannerStore.scanInProgress
       ? undefined
       : scannerStore.contentWithVulnerabilities.length,
   });
@@ -268,10 +268,7 @@ const userHeader = computed(() => {
 
       <p class="text-gray-600">
         Found {{ scannerStore.content.length }} content items with
-        <SkeletonText
-          v-if="scannerStore.anyContentLoadingPackages"
-          class="h-6 w-[2ch]"
-        />
+        <SkeletonText v-if="scannerStore.scanInProgress" class="h-6 w-[2ch]" />
         <span v-else>{{ scannerStore.totalVulnerabilities }}</span>
         vulnerabilities.
       </p>
--- a/extensions/package-vulnerability-scanner/src/components/ContentList.vue
+++ b/extensions/package-vulnerability-scanner/src/components/ContentList.vue
@@ -87,8 +87,11 @@ function collectInstalledPackages(): InstalledPackages {
   for (const item of Object.values(packagesStore.contentItems)) {
     for (const pkg of item.packages) {
       const language = pkg.language.toLowerCase();
-      if (language === "python") pypi.add(`${pkg.name}==${pkg.version}`);
-      else if (language === "r") cran.add(`${pkg.name}==${pkg.version}`);
+      if (language === "python") {
+        pypi.add(`${pkg.name}==${pkg.version}`);
+      } else if (language === "r") {
+        cran.add(`${pkg.name}==${pkg.version}`);
+      }
     }
   }
   return { pypi: [...pypi], cran: [...cran] };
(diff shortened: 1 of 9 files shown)


amylin1249:
The initial version of this extension

extensions/package-vulnerability-scanner/src/components/ContentList.vue
```diff
@@ -70,6 +72,26 @@ async function fetchPackagesInBatches(batchSize = 3) {
...
+
+// Gather unique installed packages as "name==version", grouped by repo.
+function collectInstalledPackages(): InstalledPackages {
+  const pypi = new Set<string>();
+  const cran = new Set<string>();
+  for (const item of Object.values(packagesStore.contentItems)) {
+    for (const pkg of item.packages) {
+      const language = pkg.language.toLowerCase();
+      if (language === "python") pypi.add(`${pkg.name}==${pkg.version}`);
+      else if (language === "r") cran.add(`${pkg.name}==${pkg.version}`);
```
m--: Generally I'd advocate for using braces around `if`/`else if`s for readability, but it's pretty minor here.

m-- (approved):
#415 will be a nice follow-on! Thanks for tackling this and that.

PR #417, opened 2026-06-29
Polish simple mcp server

### LINKED ISSUE #416: Polish simple-mcp-server
Update the `simple-mcp-server` example as needed. 

Non-exhaustive list of potential changes:

- Rewrite the README to the standardized template; reframe as a swap-out toy-tools example
- Fix the auth wording: connect_whoami uses the viewer session token, not API key auth
- Surface the teaching capability on the landing page (index.html.jinja): frame connect_whoami as live proof of viewer-identity auth, don't just list tools
- Reconcile the auth-header story across README, CONTRIBUTING, and index.html.jinja; drop the unused x-mcp-authorization
- Add why-comments (auto-injected session-token read, per-token visitor-client cache, private _tool_manager access)
- Retitle to the Framework: Functionality form (coordinate with the paired chat app); align H1 and description
- Regenerate requirements.txt against the declared runtime (pinned to 3.14 but declares 3.11)
- Fix stale refs (Python 3.10+ should be 3.11, the "June 2025" date)
- Clean up pyproject (version, scaffold description)
- Reconcile the Python-version drift
- Bump version 0.0.6 to 0.0.7; add a dated (#NNN) CHANGELOG entry; recompute md5 for edited bundled files

amylin1249:
Fixes #416 (can be reviewed commit by commit)

- Rewrote title, description, and README and updated app screenshot in README
- Replaced the third-party `fastmcp` package with the official MCP Python SDK (`mcp`), clearing all known package vulnerabilities (two critical) and roughly halving the dependency tree
- Fixed the auth docs: `connect_whoami` resolves the viewer from the injected session token, not API-key auth; removed the unused `x-mcp-authorization` header
- Made failures visible instead of silent or misleading
- Regenerated `requirements.txt` from Python 3.11 (was exported for 3.14) and relaxed pins so `fastapi`/`starlette` resolve to patched releases; cleaned up `pyproject.toml`
- Rewrote instructions on the landing page

**Most notable change:** the landing page greets the signed-in viewer by name, a live demo of viewer-identity auth. Before you add a "Connect Visitor API Key" integration it tells you to; once added, the same box confirms success by showing your name (or username).

Tested by manually deploying the app.


extensions/simple-mcp-server/main.py
```diff
@@ -92,63 +109,94 @@ async def connect_whoami(context: Context) -> str:
...
-        await stack.enter_async_context(mcp.session_manager.run())
+    # Run the MCP session manager for the mounted streamable-HTTP app.
+    async with mcp.session_manager.run():
         yield
 
 
-mcp_app = mcp.http_app(path="/mcp", stateless_http=True)
-app = FastAPI(title="Simple MCP Server with FastAPI", lifespan=mcp_app.lifespan)
+app = FastAPI(title="MCP Server", lifespan=lifespan)
 templates = Jinja2Templates(directory=".")
```
karawoo, line 167:
```suggestion
templates = Jinja2Templates(directory=".", autoescape=True)
```

I believe the new starlette version we're using has different behavior that no longer forces autoescape so we should add it explicitly.

extensions/simple-mcp-server/main.py
```diff
@@ -92,63 +109,94 @@ async def connect_whoami(context: Context) -> str:
...
+
+    # Look up who is viewing the page from the session token Connect injects (the
+    # same lookup connect_whoami does), so the page can greet them by name and show
+    # that tools run as the viewer, with no API key. Stays None off Connect, or when
+    # no Visitor API Key integration is configured.
+    viewer_name = None
+    session_token = request.headers.get("posit-connect-user-session-token")
+    if session_token:
+        try:
+            me = get_visitor_client(session_token).me
```
karawoo: `get_visitor_client()` and `.me` are blocking HTTP calls but this handler is `async`, so they run on the event loop. While the event loop is blocked the process can't serve anything else, including MCP tool calls. Same problem in `connect_whoami` (line 107). I think to fix this we can make `get_index_page` a plain `def` (FastAPI will run it in a worker thread), and wrap the posit-sdk calls in `connect_whoami` in `anyio.to_thread.run_sync`. The cache will then be shared across threads and needs a lock: `@cached(client_cache, lock=threading.Lock())`.

extensions/simple-mcp-server/main.py
```diff
@@ -92,63 +109,94 @@ async def connect_whoami(context: Context) -> str:
...
+        except Exception:
+            # e.g. a malformed token exchange; log it and leave the greeting off.
+            traceback.print_exc()
+
     return templates.TemplateResponse(
         request,
         "index.html.jinja",
         {
-            "server_name": mcp.name,
+            "title": "MCP Server",
```
karawoo: just curious why we switched to hardcoding the name here?

extensions/simple-mcp-server/main.py
amylin1249: Fixed in 2653ff6 - I set `templates.env.autoescape = True` in a separate line since the suggestion (although it had the right intent) raised a TypeError.

extensions/simple-mcp-server/main.py
amylin1249: This was my mistake, I think when I tried to make sure the app itself didn't include "FastAPI: " I hardcoded in two spots. Both should be reverted in the most recent commit.

extensions/simple-mcp-server/main.py
amylin1249: I moved the Connect calls in connect_whoami onto a background thread, and added the lock to the cache. But for get_index_page, I didn't make it a def because it has to stay async. It calls get_tools_info() (async), so I moved its blocking Connect call onto a thread instead. get_tools_info is async because it uses the official SDK's mcp.list_tools(), which we adopted when we replaced fastmcp to clear its package vulnerabilities (a lot was surfaced when I ran the `package-vulnerability-scanner`).

karawoo (approved):

Issue #374, opened 2026-06-10
Add an Express.js example to the Connect Gallery


dotNomad:
Add a Node.js content example to the Gallery: a small Express.js app demonstrating how to deploy Node.js content to Connect.

Depends on Connect Gallery support for Node.js requirements, shipping in Connect `2026.06`, and on this repo's pipeline support for Node.js content (#373).

## Work

- Create `extensions/express-example/` as simple content: an Express app with a `package.json` and a `manifest.json`.
- `package.json` declares `"engines": { "node": ">=22" }`, the floor for the latest Express. This is what Connect uses to select a Node installation at build and run time, and what the release automation publishes as the Gallery requirement.
- The manifest needs the standard `extension` section (name, title, description, homepage, `category: "example"`, tags, `minimumConnectVersion`).
- Set `minimumConnectVersion` to `2026.06.0`, the release shipping Node.js general availability and Gallery support for Node.js requirements. This is what keeps older Connect UIs, which ignore the unknown `nodejs` key, from offering a doomed install.
- Add tags (likely `nodejs`, `javascript`, `express`) to `extensions.json`; new tags need review per the contributing guide.
- Add a `README.md` for the Gallery homepage link.
- Register the path filter in `.github/workflows/extensions.yml` (`simple-extension-changes`).

## Release timing

Keep `version: "0.0.0"` (no release) until Connect `2026.06` ships with Gallery support for Node.js requirements. Node.js content support is generally available in the same release.

PR #423, opened 2026-07-06
feat: support Node.js content in the gallery release pipeline

### LINKED ISSUE #373: Support Node.js content in the Gallery release pipeline
We want to add Node.js content examples to the Connect Gallery (first up: an Express.js app, tracked separately). Before any Node.js extension can be added, the repo's release automation, lint, and CI need to understand Node.js content.

Connect selects a Node installation from `engines.node` in `package.json` and never reads a manifest `environment.nodejs` field, so `engines.node` is the single source of truth for the version range. The Gallery learns requirements from `requiredEnvironment` in `extensions.json`, which this repo's automation generates.

## Work

- Extend `scripts/extension-list.ts` to derive `requiredEnvironment.nodejs` from `engines.node` in the extension's `package.json` and merge it with the copied `manifest.environment`. Small change: read the sibling `package.json` next to the manifest and merge `{ nodejs: { requires: enginesNode } }` into the version's `requiredEnvironment`.
- Verify the repo lint accepts Node.js content (no R/Python manifest sections) and consider a lint rule requiring `engines.node` for content with the `nodejs` app mode.
- Ensure the publish-test CI step runs against a Connect instance with Node.js enabled (`NodeJs.Enabled` + license), so Node.js extensions can be validated in pull requests.
- Document the `engines.node` convention for Node.js content in `CONTRIBUTING.md` (the "Language version constraints" section currently covers only the manifest `environment` section). Include the `minimumConnectVersion` convention: Node.js extensions must set it to at least `2026.06.0` so older Connect UIs, which ignore the unknown `nodejs` key in `requiredEnvironment`, gate on the version check instead.

Note the npm semver wrinkle for reviewers: `engines.node` ranges are npm semver (`>=22`, `^22`, `22.x`), unlike the pep440-style ranges used for the other languages. The Connect Gallery UI evaluates the `nodejs` requirement with npm semver semantics accordingly.

amylin1249:
Fixes #373: Makes the release/CI pipeline understand Node.js content

## Changes (review commit by commit)
- **`extension-list.ts`**: derive `requiredEnvironment.nodejs` from `engines.node` in the extension's `package.json` (Connect's single source of truth; it doesn't read a `nodejs` field from the manifest).
- **lint**: require `engines.node` for `nodejs` app-mode extensions.
- **integration test action**: install a Node runtime into the running Connect container for `nodejs` extensions (the published Connect images ship no Node), enable it, and restart so Connect loads it.
- **CONTRIBUTING**: document the `engines.node` convention and the `minimumConnectVersion >= 2026.06.0` rule.

## Validation

This PR is infra-only: no Node.js example ships here (in-development examples live in `connect-staging-extensions`; the real one is the Express app, #374). So the Node provisioning is validated externally:

1. **In CI (this repo):** on a probe branch, a hello-world Node extension deployed successfully on `preview`, `release`, and `2026.06.0`:
https://github.com/posit-dev/connect-extensions/actions/runs/28797948148
2. **End-to-end in staging:** [connect-staging-extensions#36](https://github.com/posit-dev/connect-staging-extensions/pull/36) adopts this exact harness with a real Node.js hello-world. Its CI is green, and the example was deployed to dogfood, where it runs (`{"message":"Hello from Node.js on Posit Connect!"}`) and Connect recognizes it as Node.js content (Node.js logo). Confirms the approach works on a live Node-enabled Connect, not just in CI.

### DIFF
--- a/scripts/extension-list.ts
+++ b/scripts/extension-list.ts
@@ -34,9 +34,14 @@ function getManifest(extensionName: string): ExtensionManifest {
 
 // Node.js content declares its required Node version in package.json's
 // `engines.node`, which is the single source of truth Connect reads (it does not
-// read a nodejs field from manifest.json). Returns undefined for non-Node
-// extensions, which have no package.json.
-function getEnginesNode(extensionName: string): string | undefined {
+// read a nodejs field from manifest.json). Only nodejs app mode content uses it;
+// other extensions may carry a package.json for build tooling, so ignore theirs.
+// engines.node is required for nodejs content: Connect needs it to pick a
+// runtime, so a nodejs extension without it can't run.
+function getEnginesNode(extensionName: string, appmode?: string): string | undefined {
+  if (appmode !== "nodejs") {
+    return undefined;
+  }
   const packageJsonPath = path.join(
     __dirname,
     "..",
@@ -46,10 +51,16 @@ function getEnginesNode(extensionName: string): string | undefined {
     "package.json"
   );
   if (!fs.existsSync(packageJsonPath)) {
-    return undefined;
+    throw new Error(`Node.js extension ${extensionName} is missing package.json`);
   }
   const pkg = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
-  return pkg.engines?.node;
+  const enginesNode = pkg.engines?.node;
+  if (!enginesNode) {
+    throw new Error(
+      `Node.js extension ${extensionName} must set engines.node in package.json`
+    );
+  }
+  return enginesNode;
 }
 
 // Sort the given extension's version in descending order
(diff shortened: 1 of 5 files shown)


scripts/extension-list.ts
```diff
@@ -31,6 +32,26 @@ function getManifest(extensionName: string): ExtensionManifest {
   return JSON.parse(fs.readFileSync(manifestPath, "utf8"));
 }
 
+// Node.js content declares its required Node version in package.json's
+// `engines.node`, which is the single source of truth Connect reads (it does not
+// read a nodejs field from manifest.json). Returns undefined for non-Node
+// extensions, which have no package.json.
```
karawoo: some non-nodejs extensions can have a package.json (`package-vulnerability-scanner` does, for example) so I think we should combine this with a check for whether the manifest's appmode is `"nodejs"`

karawoo (commented):
I think there's an inconsistency here between how the linting step and the `extension-list.ts` handle it if `engines.node` is missing. Linting fails, but the `extension-list.ts` code doesn't -- I think we should be consistent.

scripts/extension-list.ts
amylin1249: Fixed in 4f375ea - it should now check for `metadata.appmode == nodejs`, so a non-nodejs extension that ships a `package.json` (like `package-vulnerability-scanner`) won't get a nodejs requirement pulled in here.

amylin1249:
> I think there's an inconsistency here between how the linting step and the `extension-list.ts` handle it if `engines.node` is missing. Linting fails, but the `extension-list.ts` code doesn't -- I think we should be consistent.

Fixed in 4f375ea - extension-list.ts now errors on a nodejs extension missing `engines.node`, matching the lint check so the two are consistent.

karawoo (approved):

PR #425, opened 2026-07-06
make release re-runs idempotent

dotNomad:
Re-running an Extension Workflow that partially succeeded now skips already-completed work instead of failing, so a run self-heals rather than needing a manual fix like #424.

- `.github/actions/release-extension/action.yml`: the release step no longer fails when the tag exists. It checks out the run's commit (where `extensions.json` predates the release), so `should_release` stays `true` on a re-run; guarded with `gh release view`, and "Get release data" reads the tag either way.
- `scripts/extension-list.ts`: `addExtensionVersion` skips an already-recorded version instead of throwing, so one already-listed extension no longer aborts the rest of the batch.

**Verify:** `cd scripts && npx tsc && RELEASES='[<already-listed release>,<new release>]' node ./dist/extension-list.js` - logs a skip for the existing version and adds the new one without throwing.

### DIFF
--- a/.github/actions/release-extension/action.yml
+++ b/.github/actions/release-extension/action.yml
@@ -95,13 +95,19 @@ runs:
         echo "RELEASE_TAG=$RELEASE_TAG" >> "$GITHUB_ENV"
       shell: bash
 
-    # Create the release, unless the tag already exists (the "Get release
-    # data" step below reads it by tag either way).
+    # Create the release, or if the tag already exists, ensure its asset is
+    # uploaded. `gh release view` confirms the release exists but not that its
+    # asset finished uploading, so a run that died mid-upload would leave a
+    # release with no asset (a 404 in the gallery feed). Re-uploading with
+    # --clobber heals that case and overwrites an existing asset with the
+    # identical file otherwise. The "Get release data" step below reads the tag
+    # either way.
     - name: Release
       if: github.ref_name == 'main' && steps.should_release.outputs.should_release == 'true'
       run: |
         if gh release view "$RELEASE_TAG" > /dev/null 2>&1; then
-          echo "Release $RELEASE_TAG already exists; skipping creation."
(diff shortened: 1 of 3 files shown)


.github/actions/release-extension/action.yml
```diff
@@ -95,12 +95,18 @@ runs:
...
+    # Create the release, unless the tag already exists (the "Get release
+    # data" step below reads it by tag either way).
     - name: Release
       if: github.ref_name == 'main' && steps.should_release.outputs.should_release == 'true'
       run: |
-        gh release create $RELEASE_TAG \
-          --title "${{ inputs.extension-name }} v$MANIFEST_VERSION" \
-          ${{ inputs.extension-name }}.tar.gz
+        if gh release view "$RELEASE_TAG" > /dev/null 2>&1; then
+          echo "Release $RELEASE_TAG already exists; skipping creation."
```
amylin1249:
```suggestion
          echo "Release $RELEASE_TAG already exists; ensuring its asset is uploaded."
          gh release upload "$RELEASE_TAG" ${{ inputs.extension-name }}.tar.gz --clobber
```

`gh release view` confirms the release exists but not that its `.tar.gz` finished uploading, so a run that died mid-upload would leave a release with no asset (a 404 in the gallery feed). Re-uploading with `--clobber` handles that case, and when the asset already exists, it just overwrites it with the identical file.

amylin1249 (approved):

.github/actions/release-extension/action.yml
dotNomad: Good call. I made this change in https://github.com/posit-dev/connect-extensions/pull/425/commits/d0bf746b5e4c7af2dcbdb6e3fdc12861bbec2016

dotNomad:
I also noticed one more thing to make the script idempotent - if no extensions updated the `git commit` will fail. I added a single line `git diff` check that bails if there is no extension list change in the `update-extension-list` job.

PR #418, opened 2026-06-30
Polish simple shiny chat with mcp

amylin1249:
Fixes #407 (can be reviewed commit by commit)

- Rewrote title, description, and README; added CONTRIBUTING.md and updated app screenshot in README
- Fixed MCP server registration (a bug was previously introduced)
- Forward only the viewer's own key to MCP servers (never the app's), and only to servers on this same Connect server, so a Connect key can't leak to an unrelated host entered in the sidebar; dropped the unused `X-MCP-Authorization` header
- Made failures visible instead of silent or misleading
- UI: Restored the white background behind chat messages so text stays readable over the gradient; the setup screen shows only the step still unconfigured, the sidebar shows who you're signed in as (so it's clear tools run with their permissions), and the info modal shows the provider/model cleanly
- Apply the assistant's system prompt on AWS Bedrock too (it was only applied to other providers), and made LLM-provider selection unambiguous
- Pinned dependencies, dropped the unused `nest-asyncio`, and reconciled to Python 3.11

**Most notable change:** The app had a bug where it could no longer connect to any MCP server. Routing auth through a pre-built `httpx` client fixed it, so it should be working again.

Tested by manually deploying the app.

PR #428, opened 2026-07-08
Polish package vulnerability scanner

### LINKED ISSUE #415: Polish package-vulnerability-scanner
Work on after merging #411 

Update `package-vulnerability-scanner` as needed, focusing on performance improvements.

amylin1249:
Fixes #415.

Polishes the `package-vulnerability-scanner` (can review commit by commit):
- Viewer identity switch: Connect API calls now run as the signed-in viewer via their per-request session token, instead of the deploying publisher, so each person sees the content they published and their own name in the header. This requires a Connect Visitor API Key integration (added to `requiredFeatures`).
- In-app setup page: When the integration is missing, the app shows a setup screen with the steps to add it, similar to `simple-shiny-chat-with-mcp`.
- Performance improvement: Packages for all visible content load in a single request. Vulnerability lookups query Package Manager in parallel across chunks, bounded to five concurrent requests. The package cache is kept when toggling between your content and all content, so re-viewing already-scanned content makes no new requests.
- Robustness: Transient 5xx responses (from Connect or Package Manager) are retried with a short backoff, so a brief blip doesn't fail the whole scan. A failed fetch now shows a clear reason for failure instead of a premature zero or a permanent spinner.
- Accessibility: The content-visibility toggle now has an accessible label so screen readers announce its purpose.
- Security: Updated frontend dependencies to clear vulnerabilities flagged by npm audit, including a ReDoS advisory in the bundled markdown-it renderer.
- Docs: README and description rewritten to match style of other Gallery items.
- Testing: Manually deployed and verified both by self and non-publisher viewer to confirm the viewer identity switch works, and that there are no timeout errors with lots of content (~9000). Added a backend test suite (`test_main.py`, 29 tests covering the retry/setup helpers and all four API endpoints) and a frontend store test suite (37 tests across the Pinia stores and `collectInstalledPackages`). Both run in the extension workflow — `npm test` before the build and `uv run pytest` after.


karawoo (dismissed):

karawoo:
Ack sorry I meant to approve https://github.com/posit-dev/connect-extensions/pull/418, got my tabs mixed up. I have not reviewed this yet.

extensions/package-vulnerability-scanner/main.py
```diff
@@ -19,25 +46,85 @@
...
-        return packages
+        visitor = get_visitor_client(request)
+
+        # Read every package across the viewer's content from the server-wide
+        # endpoint and group by content, instead of a request per item. Each
+        # record carries the guid of the content it belongs to, so a server with
+        # thousands of items scans in one pass rather than thousands of round-trips.
+        def fetch_grouped() -> dict[str, list]:
+            by_guid: dict[str, list] = {}
+            for pkg in visitor.packages.fetch():
```
karawoo: Isn't the server-wide endpoint admin-only? If the viewer is not an admin, I don't think this will work.

I'm realizing the [API docs](https://docs.posit.co/connect/api/#get-/v1/packages) don't suggest this but I think that is an omission.

amylin1249: Yes, good catch, /v1/packages is admin-only, so I split it into two views:
- Your own content reads per item via GET /v1/content/{guid}/packages (works for any viewer on content they own)
- The admin "all content" view uses the server-wide GET /v1/packages (admin-only, matching the admin-only toggle)

This should solve the performance issues we were running into while also working on the different role types. I tested this in dogfood as both an administrator and a publisher.

I re-requested you for review, let me know if there is any need to open an issue to edit the docs to make the admin-only piece more explicit.

extensions/package-vulnerability-scanner/main.py
```diff
@@ -19,25 +48,131 @@
...
+        async with semaphore:
+            try:
+                packages = await _fetch_with_retry(
+                    lambda: _read_item_packages(visitor, guid)
+                )
+                return guid, {"packages": packages}
+            except ClientError as e:
+                if e.error_code == 212:
+                    raise  # missing integration: surface as the setup screen
+                return guid, {"packages": [], "error": f"Connect API error: {e}"}
```
karawoo: If I'm reading correctly, this error message never reaches the user. It flows to the frontend as `packageFetchError`, but ContentCard and VulnerabilityChecker both show fixed text ("Couldn't load packages" / "The scan couldn't read this content's packages from Connect") and never display the message itself.

It'd be nice it we could render `packageFetchError.message` in VulnerabilityChecker's error details. I think we'd also want `e.error_message` here instead of `e` so we don't show the full json dump.

```suggestion
                return guid, {"packages": [], "error": f"Connect API error: {e.error_message}"}
```

extensions/package-vulnerability-scanner/main.py
```diff
@@ -1,16 +1,45 @@
...
+
+# Build a Connect client scoped to the signed-in viewer by exchanging their
+# per-request session token, so API calls return that viewer's own content and
+# identity. Falls back to the default client when there's no token (e.g. local
+# development).
+def get_visitor_client(request: Request) -> connect.Client:
+    token = request.headers.get("posit-connect-user-session-token")
+    if token:
+        return client.with_user_session_token(token)
+    return client
```
karawoo: IIUC the session token header is only present for logged-in users and when `Applications.OAuthIntegrationsEnabled=true`. Have we checked the behavior here outside of that scenario (i.e., an anonymous viewer visiting content that is accesslble to anyone, no login required)? I think in that case we'd fall back to showing the package vulnerability info for all the content the _owner_ has access to, which is not what we should do, though it's possible I'm misunderstanding something here.

extensions/package-vulnerability-scanner/main.py
amylin1249: Fixed in 038a2e8 -- the per-item error should now use e.error_message, and VulnerabilityChecker renders packageFetchError.message, so the reason should reach the user. I also added tests to verify the work.

extensions/package-vulnerability-scanner/main.py
amylin1249: I deployed the app to check, and it turns out that "an anonymous viewer visiting content that is accessible to anyone, no login required" isn't actually possible. When I went to the access settings, the option to share with an anonymous viewer was grayed out, and it said "Anyone - no login required Not available for interactive content." But I think your underlying point would still stand in the case where OAuth integrations are disabled server-wide, so I tried to implement a fix in 3d1fc8d. I also added tests to verify the work.

extensions/package-vulnerability-scanner/tsconfig.app.json
```diff
@@ -11,5 +11,6 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.vue"]
+  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.vue"],
+  "exclude": ["src/**/*.test.ts"]
```
dotNomad, line 15: Any reason not to get typescript checking on the test files?

extensions/package-vulnerability-scanner/manifest.json
```diff
@@ -39,11 +39,14 @@
...
-    "description": "Scan for security vulnerabilities in your published content on Connect using Posit Package Manager. View an overview of CVEs (Common Vulnerabilities and Exposures) across all your apps, and get guidance on upgrading affected packages.",
+    "description": "Scan your published content on Connect for known security vulnerabilities in its Python and R packages, using Posit Package Manager. Get an overview of the CVEs (Common Vulnerabilities and Exposures) across your content, then open any item to see the affected packages and the version to upgrade to.",
     "homepage": "https://github.com/posit-dev/connect-extensions/tree/main/extensions/package-vulnerability-scanner",
     "category": "extension",
     "minimumConnectVersion": "2025.04.0",
-    "requiredFeatures": ["API Publishing"],
-    "version": "3.0.6"
+    "requiredFeatures": [
+      "API Publishing",
+      "OAuth Integrations"
```
dotNomad, line 48: If we require OAuth Integrations do we need to bump the `minimumConnectVersion` too?

extensions/package-vulnerability-scanner/package.json
```diff
@@ -21,10 +22,13 @@
     "@tailwindcss/typography": "^0.5.16",
     "@types/markdown-it": "^14.1.2",
     "@vitejs/plugin-vue": "^6.0.6",
+    "@vue/test-utils": "^2.4.11",
     "@vue/tsconfig": "^0.8.1",
+    "happy-dom": "^20.10.6",
```
dotNomad, line 27: Do we need `happy-dom`? I don't see it setup in the Vitest configuration.

dotNomad (commented):
I had a few questions on the configuration setup here

extensions/package-vulnerability-scanner/tsconfig.app.json
amylin1249: Added this in 2c46a55.

extensions/package-vulnerability-scanner/manifest.json
amylin1249: I think the `minimumConnectVersion` set as 2025.04.0 is okay. This is the floor that all of the other extensions with the OAuth Integration requirement (simple-mcp-server, publisher-command-center, chat-with-content, etc.) have, and they all run fine.

extensions/package-vulnerability-scanner/package.json
amylin1249: We do. It's in `App.test.ts` and `VulnerabilityChecker.test.ts`.

extensions/package-vulnerability-scanner/main.py
karawoo: > When I went to the access settings, the option to share with an anonymous viewer was grayed out, and it said "Anyone - no login required Not available for interactive content." 

this depends on the license terms, some connect instances will allow anonymous access to interactive content

amylin1249:
Closed to move the work into stacked PRs: https://github.com/posit-dev/connect-extensions/pull/453, https://github.com/posit-dev/connect-extensions/pull/454, https://github.com/posit-dev/connect-extensions/pull/455, https://github.com/posit-dev/connect-extensions/pull/456, https://github.com/posit-dev/connect-extensions/pull/457, https://github.com/posit-dev/connect-extensions/pull/458, https://github.com/posit-dev/connect-extensions/pull/459, https://github.com/posit-dev/connect-extensions/pull/460, https://github.com/posit-dev/connect-extensions/pull/461, https://github.com/posit-dev/connect-extensions/pull/462.

PR #433, opened 2026-07-16
Polish chat-with-content extension

### LINKED ISSUE #429: Polish chat-with-content
Potential changes (non-exhaustive list):
- Rewrite README into a genuinely useful doc: what it does, who it's for, setup/config, how to use
- Rewrite description: concise, states functionality + value
- Add a "how this works" surface: show the mechanism in-app (e.g. "signed in as {you}, resolved from your session; content read with your permissions, no admin key stored")
- Add why-comments to the source
- Add test suite + CI (none today)
- Make the startup Bedrock credential probe conditional
- Guard against undefined chat / null dates
- Truncate page DOM before sending to the LLM
- Surface real chat errors (on_error="actual")
- Refresh dated model names; drop duplicate chatlas pin
- Bump version, CHANGELOG, checksums

amylin1249:
Fixes #429 (can be reviewed commit by commit):
- Rewrote the README and description, and added CONTRIBUTING doc
- Reworked setup screen to match the `simple-shiny-chat-with-mcp` example (clear instructions and indicates what still needs to be configured)
- Added an in-app "how this works" note (signed in as you; content read with your permissions; no admin key stored)
- Surface errors clearly to the user
- Truncate large pages before sending to the model; guard null/malformed deploy times and a missing chat provider (no more crashes)
- Added a helper test suite + CI
- Refreshed default model to Claude Sonnet 4.5
- Bedrock probe only runs when no provider is configured (similar to `simple-shiny-chat-with-mcp`)

Manually deployed for verification.

### DIFF
--- a/.github/workflows/extensions.yml
+++ b/.github/workflows/extensions.yml
@@ -56,7 +56,6 @@ jobs:
             stock-report: extensions/stock-report/**
             simple-mcp-server: extensions/simple-mcp-server/**
             simple-shiny-chat-with-mcp: extensions/simple-shiny-chat-with-mcp/**
-            chat-with-content: extensions/chat-with-content/**
             pqr: extensions/pqr/**
 
       # When infra changed, test all simple extensions; otherwise use the filter output
@@ -212,6 +211,7 @@ jobs:
       package-vulnerability-scanner: ${{ steps.resolve.outputs.package-vulnerability-scanner }}
       runtime-version-scanner: ${{ steps.resolve.outputs.runtime-version-scanner }}
       usage-metrics-dashboard: ${{ steps.resolve.outputs.usage-metrics-dashboard }}
+      chat-with-content: ${{ steps.resolve.outputs.chat-with-content }}
 
     steps:
       - uses: actions/checkout@v4
@@ -233,6 +233,7 @@ jobs:
             package-vulnerability-scanner: extensions/package-vulnerability-scanner/**
             runtime-version-scanner: extensions/runtime-version-scanner/**
             usage-metrics-dashboard: extensions/usage-metrics-dashboard/**
+            chat-with-content: extensions/chat-with-content/**
 
       # When infra changed, trigger all complex extensions too
       - id: resolve
@@ -243,11 +244,13 @@ jobs:
             echo "package-vulnerability-scanner=true" >> $GITHUB_OUTPUT
             echo "runtime-version-scanner=true" >> $GITHUB_OUTPUT
             echo "usage-metrics-dashboard=true" >> $GITHUB_OUTPUT
+            echo "chat-with-content=true" >> $GITHUB_OUTPUT
           else
             echo "publisher-command-center=${{ steps.changes.outputs.publisher-command-center }}" >> $GITHUB_OUTPUT
             echo "package-vulnerability-scanner=${{ steps.changes.outputs.package-vulnerability-scanner }}" >> $GITHUB_OUTPUT
             echo "runtime-version-scanner=${{ steps.changes.outputs.runtime-version-scanner }}" >> $GITHUB_OUTPUT
             echo "usage-metrics-dashboard=${{ steps.changes.outputs.usage-metrics-dashboard }}" >> $GITHUB_OUTPUT
+            echo "chat-with-content=${{ steps.changes.outputs.chat-with-content }}" >> $GITHUB_OUTPUT
           fi
 
   # Creates and releases the Publisher Command Center extension using a custom
@@ -287,6 +290,15 @@ jobs:
     uses: ./.github/workflows/usage-metrics-dashboard.yml
     secrets: inherit
 
+  # Creates and releases the Chat with Content extension using a custom workflow
+  chat-with-content:
+    needs: [complex-extension-changes]
+    # Only runs if the `complex-extension-changes` job detects changes in the
+    # chat-with-content extension directory
+    if: ${{ needs.complex-extension-changes.outputs.chat-with-content == 'true' }}
+    uses: ./.github/workflows/chat-with-content.yml
+    secrets: inherit
+
   # All extensions have been linted, packaged, and released, if necessary
   # Continuing to update the extension list with the latest release data
 
@@ -300,6 +312,7 @@ jobs:
       package-vulnerability-scanner,
       runtime-version-scanner,
       usage-metrics-dashboard,
+      chat-with-content,
     ]
     permissions:
       pull-requests: write
@@ -383,6 +396,7 @@ jobs:
       package-vulnerability-scanner,
       runtime-version-scanner,
       usage-metrics-dashboard,
+      chat-with-content,
     ]
     if: ${{ always() }}
     outputs:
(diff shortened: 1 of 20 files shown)


extensions/chat-with-content/app.py
```diff
@@ -242,36 +309,88 @@ def server(input: Inputs, output: Outputs, session: Session):
...
     @reactive.Effect
     def _():
-        content_list = fetch_connect_content_list(client)
-        content_choices = {
-            item.guid: f"{item.title or item.name} - {item.owner.first_name} {item.owner.last_name} {time_since_deployment(item.last_deployed_time)}"
-            for item in content_list
-        }
+        # The selector only appears once setup is complete; skip the fetch (and its
+        # error toast) while the setup screen is still up.
+        if not VISITOR_API_INTEGRATION_ENABLED:
```
dotNomad: Does this need to check if `token_error` is not `None` as well?

Should we check `chat` as well?

dotNomad: If we do need to block on `token_error` we may want to consider adding a test around that.

amylin1249: Yes, I think we need to check for all three here, instead of just `VISITOR_API_INTEGRATION_ENABLED`. 460d809 should handle it now (+ tests).

.github/workflows/extensions.yml
```diff
@@ -68,6 +67,12 @@ jobs:
             echo "Infrastructure changed — testing all simple extensions"
             changes=$(find extensions -maxdepth 1 -mindepth 1 -type d -printf '%f\n' \
               | while read -r ext; do
+                  # Skip complex extensions: they have their own dedicated
+                  # workflow (which builds them) and would fail the simple
+                  # no-build flow (e.g. a missing frontend dist/).
+                  if [ -f ".github/workflows/$ext.yml" ]; then
+                    continue
+                  fi
```
dotNomad, line 75: Is this skip necessary? The simple extensions do require an explicit adding to the filters above.

extensions/chat-with-content/README.md
```diff
@@ -1,76 +1,74 @@
...
+After deploying, in the content's settings:
+
+- **Choose an LLM** by setting `CHATLAS_CHAT_PROVIDER_MODEL` (for example
+  `openai/gpt-4o`) plus the matching API key (`OPENAI_API_KEY`,
+  `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, ...) on the **Advanced** tab, under
+  **Environment Variables**. See the
+  [chatlas `ChatAuto` docs](https://posit-dev.github.io/chatlas/reference/ChatAuto.html)
+  for provider/model strings. On AWS Bedrock with an instance role, credentials
+  are detected automatically and no vars are needed. (The older
+  `CHATLAS_CHAT_PROVIDER` and `CHATLAS_CHAT_ARGS` still work but are deprecated.)
```
dotNomad: It may be helpful to have one example in the README. The previous doc had quite a few and linking to the chatlas docs is better, but still find myself wanting an example of choosing / setting up an LLM.

Side note I find the `- **label**` style of Claude to be a bit harder to read compared to headers, paragraphs, and short lists. Nitpicky, but if we are optimizing could be worth doing.

extensions/chat-with-content/CHANGELOG.md
```diff
@@ -5,6 +5,44 @@ All notable changes to the Chat with Content extension will be documented in thi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.0.8] - 2026-07-17
+
+### Added
+
+- Backend test suite (`test_helpers.py`) run in a dedicated CI workflow. (#433)
```
dotNomad: I think we should keep the changelog to just items that affect users of the extensions.

extensions/chat-with-content/CHANGELOG.md
```diff
@@ -5,6 +5,44 @@ All notable changes to the Chat with Content extension will be documented in thi
...
+- Show a clear error when the app can't verify your Connect session, instead of
+  silently running with the wrong identity and failing later. (#433)
+- Require the viewer's Connect session before listing content: when the app can't
+  read a session token (for example if OAuth integrations are disabled on the
+  server), it shows the setup screen instead of listing the deployer's content. (#433)
+- Don't summarize an unrelated page if the content frame redirects cross-origin
+  (for example to an external login). (#433)
+- Truncated large content before sending it to the model so a big page can't
+  overflow the context window. (#433)
+- Removed a duplicate `chatlas` dependency pin. (#433)
```
dotNomad: Similar note here about customer-facing CHANGELOG.

extensions/chat-with-content/app.py
```diff
@@ -242,36 +310,90 @@ def server(input: Inputs, output: Outputs, session: Session):
...
+            name = (
+                f"{me.get('first_name', '')} {me.get('last_name', '')}".strip()
+                or me.get("username")
+                or "you"
+            )
+        except Exception:
+            pass
+        return ui.p(
+            ui.HTML(
+                f"Signed in as <strong>{name}</strong>, resolved from your Connect "
```
dotNomad: We should escape the name if we are putting it in an HTML block. We can do that with `ui.strong(name)` I believe and remove the `ui.HTML` call entirely.

dotNomad (commented):
A couple minor things, but more importantly I tried to deploy this on our internal server and got a 504 gateway timeout - were you able to successfully deploy `chat-with-content`?

.github/workflows/extensions.yml
amylin1249: I think it's necessary. This is how we've done the other extensions (runtime-version-scanner, package-vulnerability-scanner, publisher-command-center). I think it's possible to use an explicit opt-in list, but I'd probably open that as a separate issue to fix across all of them.

dotNomad: Ah I see this is to avoid the double runs that we were getting with infra changes for the complex extensions 👍

extensions/chat-with-content/CHANGELOG.md
amylin1249: Done in 5587547. FYI I also applied the changes to other CHANGELOGs - chat-with-content, portfolio-dashboard, quarto-presentation, script-python, simple-mcp-server, and simple-shiny-chat-with-mcp.

extensions/chat-with-content/app.py
amylin1249: Fixed in 23025b8, thanks!

extensions/chat-with-content/README.md
amylin1249: Added back in the one example (which is clearly visible on the front page of the extension as well) in 1d4acd. FYI, I also made the same change in `simple-shiny-chat-with-mcp`'s README.

3992ab4 fixes the `- **label**` style in READMEs that had it (chat-with-content, portfolio-dashboard, quarto-presentation, simple-mcp-server, and simple-shiny-chat-with-mcp).

amylin1249:
Closed to move the work into stacked PRs: #446, #447, #448.

PR #434, opened 2026-07-17
Polish publisher command center

### LINKED ISSUE #431: Polish publisher-command-center
Baselines:
- Rewrite README (currently 13 lines): what it does, who it's for, setup/config, how to use
- Rewrite description: concise, states functionality + value
- Add a "how this works" panel: surface the Visitor API Key identity mechanism ("acting as {you}; you see only the content you can manage")
- Add why-comments to the source

App-specific:
- Add backend test suite + CI
- Add frontend (Mithril) test suite + CI
- Fix N+1 in GET /api/contents
- Fix dead async error handling across components
- Guard null-date formatting crashes
- Return proper HTTP errors instead of raw 500s
- Delete or fix the broken Metrics feature; remove stray console.logs
- Bump version, CHANGELOG, checksums

amylin1249:
Fixes #431 
- Acts as the signed-in viewer through a Connect Visitor API Key -- each person only sees and manages the content they have access to, and the home screen tells you your name
- Aligns with other examples and extensions to manually display detailed setup instructions (replaces the previous one-click integration button) -- this is more robust
- Returns errors with the reason Connect reported
- Adds backend (`pytest`) and frontend (`vitest`) test suites, both run in CI
- Runs the synchronous API endpoints in a threadpool and removes the N+1 in `GET /api/contents` (per-item process counts now fetched concurrently)
- Rewrites description and README
- Removes the non-functional Metrics feature and stray dead code
- Hardens the frontend: surfaces load/action failures instead of swallowing them, guards date formatting, standardizes error wording, and fixes the rename dialog reusing a stale guid


extensions/publisher-command-center/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
dotNomad, line 1: Should this be the minimum Python version - `3.9`?

extensions/publisher-command-center/CHANGELOG.md
```diff
@@ -5,6 +5,69 @@ All notable changes to the Publisher Command Center extension will be documented
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.0.9] - 2026-07-17
+
+### Added
+
+- Backend test suite (`test_app.py`) and frontend test suite (`vitest`), both run
```
dotNomad: Let's remove non-customer facing things from the `CHANGELOG.md`

dotNomad (commented):
Looking at the configuration and doc changes this is looking good so far. I have two minor comments, and I'm still reviewing the code changes.

extensions/publisher-command-center/app.py
```diff
@@ -1,149 +1,195 @@
...
 client_cache = TTLCache(maxsize=float("inf"), ttl=3600)
 
 
-@app.get("/api/visitor-auth")
-async def integration_status(posit_connect_user_session_token: str = Header(None)):
-    """
-    If running on Connect, attempt to build a visitor client.
-    If that raises the 212 error (no OAuth integration), return authorized=False.
-    """
+@cached(client_cache)
```
dotNomad, line 25: We need to pass a lock to the cache here since `cachetools` is not thread-safe without the lock.

```
import threading

@cached(client_cache, lock=threading.Lock())
def _build_visitor_client(token): ...
```

extensions/publisher-command-center/app.py
```diff
@@ -153,47 +199,50 @@ async def destroy_process(
...
     visitor = get_visitor_client(posit_connect_user_session_token)
-
-    content = visitor.content.get(content_id)
-    job = content.jobs.find(process_id)
-    if job:
+    try:
+        content = visitor.content.get(content_id)
+        job = content.jobs.find(process_id)
+        if not job:
+            return  # already gone; nothing to stop
```
dotNomad: Nit: Looking at `content.jobs.find` if there is a missing job it 404s with a `ClientError` raised. So there isn't a case where `if not job` would happen.

If we want to have the do nothing if nothing found we can do that with something like

```
def _find_job(content, key):
    try:
        return content.jobs.find(key)
        ...
    except ClientError as err:
        if err.http_status == 404:
            return None
        raise
```

extensions/publisher-command-center/src/components/Processes.js
```diff
@@ -24,21 +23,24 @@ const StopButton = {
...
+              // stop failure, so it must not report "Couldn't stop the process".
               Processes.reset();
-              return Processes.load();
-            })
-            .then(() => {
-              m.redraw(); // Trigger UI refresh after reload
-            })
-            .catch((err) => {
-              console.error("Failed to reload processes:", err);
+              Processes.load()
```
dotNomad, line 31: Should this pass a content ID?

extensions/publisher-command-center/src/components/Processes.js
```diff
@@ -24,21 +23,24 @@ const StopButton = {
...
+                .catch((err) =>
+                  console.error("Failed to reload processes after stop:", err),
```
dotNomad, line 33: Should we do a `notifyError()` here instead?

extensions/publisher-command-center/app.py
```diff
@@ -1,149 +1,195 @@
...
+        raise _connect_http_error(err)
 
 
 @app.get("/api/contents")
 async def contents(posit_connect_user_session_token: str = Header(None)):
     visitor = get_visitor_client(posit_connect_user_session_token)
-
-    all_content = visitor.content.find()
+    try:
+        all_content = visitor.content.find()
```
dotNomad: Here is some pointed analysis from Claude after a `/code-review`

---

This route is still `async def`, so this call and the `get_visitor_client` above it block the event loop — the thing the sync conversion in the rest of this PR avoids. `content.find()` is a paginated HTTP fetch and `get_visitor_client` does a token exchange on a cache miss.

It has to stay `async` for the `gather` below, so the fix is the same `to_thread` treatment `active_jobs` already gets:

visitor = await asyncio.to_thread(get_visitor_client, posit_connect_user_session_token)
all_content = await asyncio.to_thread(visitor.content.find)

`HTTPException` propagates out of `to_thread` unchanged, so the 424/502 behavior is preserved.

extensions/publisher-command-center/app.py
```diff
@@ -153,47 +199,50 @@ async def destroy_process(
     posit_connect_user_session_token: str = Header(None),
 ):
     visitor = get_visitor_client(posit_connect_user_session_token)
-
-    content = visitor.content.get(content_id)
-    job = content.jobs.find(process_id)
-    if job:
+    try:
+        content = visitor.content.get(content_id)
```
dotNomad: Some more pointed advice from Claude

---

Same event-loop concern as the `contents` route, and heavier here: up to 34 blocking round trips (`content.get`, j`obs.find`, `job.destroy`, then a `jobs.find` per poll iteration).

To be precise, `await asyncio.sleep(1)` does yield, so this isn't a 30-second stall — it's ~31 short stalls, each one Connect round trip long, during which no other request progresses. Latency jitter under concurrent use rather than an outage.

Unlike `contents`, this route has no reason to be `async` at all. Making it a plain `def` with `time.sleep(1)` matches the other nine routes in this file. Tradeoff worth naming: a sync route occupies one of the 40 anyio worker threads for its lifetime, so concurrent stops cap at 40 — fine for a deliberate single-item action. `test_destroy_process_still_running_returns_504` would then patch app.time.sleep instead of `app.asyncio.sleep`.

extensions/publisher-command-center/app.py
```diff
@@ -1,149 +1,195 @@
...
+    # Connect error is an upstream failure the caller can't fix, so surface a 502
+    # with Connect's message rather than a raw 500.
+    if err.error_code == 212:
+        return HTTPException(
+            status_code=424,
+            detail="A Connect Visitor API Key integration is required.",
+        )
+    return HTTPException(
+        status_code=502, detail=f"Connect API error: {err.error_message}"
+    )
```
dotNomad, line 42: Should pass 400-499 codes through rather than turning them into 502?

extensions/publisher-command-center/app.py
```diff
@@ -1,149 +1,195 @@
...
+
+
+# Anything not already turned into an HTTPException (a network failure, an
+# unexpected SDK error) becomes a 502 with a curated message; the real error is
+# logged for the publisher rather than shown to the viewer.
+@app.exception_handler(Exception)
+async def _unexpected_error(request, exc):
+    print(f"Unexpected error handling {request.url.path}: {exc}")
+    return JSONResponse(
+        status_code=502, content={"detail": "Connect API error. Please try again."}
```
dotNomad, line 52: Similar note here about passing along the Connect API errors. This makes any unexpected error a Connect API error which seems incorrect to me.

extensions/publisher-command-center/app.py
```diff
@@ -1,149 +1,195 @@
...
-    If that raises the 212 error (no OAuth integration), return authorized=False.
-    """
+@cached(client_cache)
+def _build_visitor_client(token: Optional[str]) -> connect.Client:
+    return client.with_user_session_token(token) if token else client
+
+
+def _connect_http_error(err: ClientError) -> HTTPException:
+    # Error 212 means no Visitor API Key integration is configured; the frontend
+    # treats 424 as "needs setup" and shows the integration prompt. Any other
```
dotNomad, line 32: Is this correct? Looking at `src/index.js` it checks `{authorized: false}`, but I could be missing something.

dotNomad (commented):
I'm noticing a lot of things with this change:

> Run the synchronous API endpoints as sync `def` so FastAPI executes them in a
  threadpool; the content detail page's parallel requests now run concurrently
  instead of serializing on the event loop. (#434)

I'm wondering if maybe we pull that out as separate work to land everything else this PR is accomplishing for an easier PR. Maybe this work is too connected now though. Let me know what you think @amylin1249

PR #436, opened 2026-07-21
simple-mcp-server example: harden landing-page links, align CONTRIBUTING with uv

amylin1249:
Follow up to #417
- Sets `rel="noopener"` on the landing page's external links so a linked page can't reach back through `window.opener`
- Aligns CONTRIBUTING with the paired `simple-shiny-chat-with-mcp` example: `uv` instead of `pip`, plus Setup/Bundle/Changelog sections

PR #438, opened 2026-07-21
chore(simple-mcp-server): sync pyproject version with manifest

amylin1249:
Follow up to #436 - bump pyproject version for consistency with manifest version


dotNomad (approved):
Thanks for coming back for this

PR #441, opened 2026-07-24
refactor(scripts): type the GitHub release data instead of any

### LINKED ISSUE #402: refactor(scripts): type the GitHub release data instead of any
Small issue surfaced when working on #401 

In scripts/extension-list.ts, the GitHub release data is untyped (any) in three spots: getExtensionNameFromRelease, addRelease's githubRelease param, and the parsed RELEASES. Typos or payload-shape changes fail at runtime instead of build.

We can fix this by adding a GithubRelease interface (tag_name, published_at, assets[]) in scripts/types/index.ts and annotate the three spots. This also surfaces an existing edge case: assets.find(...) can return undefined, which we'd resolve with a guard that throws a clear error, matching the file's existing pattern of throwing errors.

amylin1249:
Fixes #402. Typos or payload-shape changes should now fail at build instead of runtime, and the cause of errors is made explicit.

### DIFF
--- a/scripts/extension-list.ts
+++ b/scripts/extension-list.ts
@@ -11,12 +11,12 @@ import {
   ExtensionEnvironment,
   ExtensionManifest,
   ExtensionVersion,
+  GithubRelease,
   RequiredFeature
 } from "./types";
 
-function getExtensionNameFromRelease(release: any): string {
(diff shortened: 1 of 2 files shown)


dotNomad (approved):
Better typing is always welcome! :shipit:

PR #442, opened 2026-07-24
ci: enforce quarto extensions render only what they serve

### LINKED ISSUE #403: ci: guard against Quarto extensions serving the wrong document
Surfaced in #392: a Quarto extension with no pinned primary document can silently serve a sibling .md (e.g. CHANGELOG.html) instead of its entrypoint, since Quarto renders CHANGELOG.md (it ignores README.md). No check catches it; everything builds fine.

o prevent this in the future, add a lint-extension check that runs quarto inspect and enforces the standard that a Quarto extension renders only the document it serves: a non-website extension fails if it renders a sibling Markdown page. Websites are exempt (they render many pages and serve index.html by convention). Add a CONTRIBUTING.md note to pin project.render.

Note: This edits shared infra, so CI will lint all extensions. quarto-stock-report-python/-r render a CHANGELOG today, so this PR also pins their project.render (and adds a matching pin to pqr) so they meet the standard. quarto-website legitimately renders many pages and is exempt.

amylin1249:
Fixes #403 -- adds a lint-extension guard that enforces the standard that a Quarto extension renders only the document it serves.

To make the existing content pass, this PR pins project.render to the entrypoint on quarto-stock-report-python, quarto-stock-report-r, and pqr (these ones rendered a stray CHANGELOG page). 

As a part of the cleanup, this PR also deletes connect-extension.qmd from quarto-document and quarto-stock-report-python (this is a superseded legacy gallery-metadata file that now lives in the extension block of manifest.json).

### DIFF
--- a/.github/actions/lint-extension/action.yml
+++ b/.github/actions/lint-extension/action.yml
@@ -150,3 +150,65 @@ runs:
           exit 1
         fi
       shell: bash
+
+    # The render-set guard below only applies to Quarto content.
+    - name: Detect Quarto appmode
+      id: quarto-appmode
+      run: |
+        APPMODE=$(jq -r '.metadata.appmode // ""' ./extensions/${{ inputs.extension-name }}/manifest.json)
+        if [[ "$APPMODE" == quarto-* ]]; then
+          echo "is-quarto=true" >> "$GITHUB_OUTPUT"
+        else
+          echo "is-quarto=false" >> "$GITHUB_OUTPUT"
+        fi
+      shell: bash
+
+    - name: Set up Quarto
+      if: steps.quarto-appmode.outputs.is-quarto == 'true'
+      uses: quarto-dev/quarto-actions/setup@v2
+
+    # Guards against the served-document bug (#392): Quarto renders every sibling
+    # Markdown file (e.g. CHANGELOG.md) as its own page, and with no pinned
+    # primary the default served document can silently flip to that page (e.g.
+    # CHANGELOG.html) instead of the entrypoint. Every build-time check stays
+    # green, so nothing else catches it. The standard is that a Quarto extension
+    # renders only what it serves: single-document content pins project.render to
+    # its entrypoint, while websites render their own pages. So this fails when a
+    # non-website extension renders a sibling Markdown page.
+    - name: Check Quarto render set
+      if: steps.quarto-appmode.outputs.is-quarto == 'true'
+      run: |
+        set -euo pipefail
+        EXT_DIR=./extensions/${{ inputs.extension-name }}
+
+        # Without a _quarto.yml the directory is not a project, so Connect renders
+        # only the entrypoint and no sibling Markdown becomes a page.
+        if ! INSPECT=$(quarto inspect "$EXT_DIR" 2>/dev/null); then
+          echo "Not a Quarto project; only the entrypoint renders. OK."
+          exit 0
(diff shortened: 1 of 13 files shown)


.github/workflows/extensions.yml
```diff
@@ -35,6 +35,7 @@ jobs:
             _infra:
               - integration/**
               - .github/actions/connect-integration-test/**
+              - .github/actions/lint-extension/**
```
dotNomad, line 38: Should this be added to the complex-extension block too?

.github/actions/lint-extension/action.yml
```diff
@@ -150,3 +150,65 @@ runs:
...
+    # non-website extension renders a sibling Markdown page.
+    - name: Check Quarto render set
+      if: steps.quarto-appmode.outputs.is-quarto == 'true'
+      run: |
+        set -euo pipefail
+        EXT_DIR=./extensions/${{ inputs.extension-name }}
+
+        # Without a _quarto.yml the directory is not a project, so Connect renders
+        # only the entrypoint and no sibling Markdown becomes a page.
+        if ! INSPECT=$(quarto inspect "$EXT_DIR" 2>/dev/null); then
```
dotNomad, line 186: Any `quarto inspect` failure will result in "not a Quarto project" so if the `_quarto.yml` is incorrect, like missing a semicolon in `render:` `quarto inspect` will error, but this will pass.

I think that this is tricky to do right though, maybe this is a good first pass that we improve over time.

.github/actions/lint-extension/action.yml
```diff
@@ -150,3 +150,65 @@ runs:
...
+        fi
+
+        # Rendered input basenames.
+        INPUTS=$(jq -r '.files.input[] | split("/") | last' <<< "$INSPECT")
+        COUNT=$(grep -c . <<< "$INPUTS")
+
+        # A lone document is unambiguously the served document. Otherwise a
+        # rendered sibling Markdown file (Quarto ignores README.md) is a stray
+        # page that could become the served document.
+        STRAY_MD=$(grep -iE '\.md$' <<< "$INPUTS" | grep -viE '^README\.md$' || true)
```
dotNomad, line 205: This isn't checking for stray `.qmd` docs or other file types that Quarto can render.

.github/actions/lint-extension/action.yml
```diff
@@ -150,3 +150,65 @@ runs:
...
+        # Websites legitimately render many pages and serve index.html by convention.
+        PROJECT_TYPE=$(jq -r '.config.project.type // "default"' <<< "$INSPECT")
+        if [ "$PROJECT_TYPE" == "website" ]; then
+          echo "Website project; many pages are expected. OK."
+          exit 0
+        fi
+
+        # Rendered input basenames.
+        INPUTS=$(jq -r '.files.input[] | split("/") | last' <<< "$INSPECT")
+        COUNT=$(grep -c . <<< "$INPUTS")
```
dotNomad, line 200: If there are no input matches I think this will `exit 1` and fail.

dotNomad (commented):
I think all of the changes to the extensions here make a lot of sense, but the linting needs a bit more work. My suggestion would be just fix the extensions and come back to the linting in future work.

amylin1249:
Splitting per your feedback: extension fixes in #449, linting (with your comments addressed) stacked on top in #450.

PR #449, opened 2026-07-28
fix: pin quarto extensions to render only their served document

amylin1249:
Split out of #442 per @dotNomad's review.

Quarto renders every sibling Markdown file as its own page. With no pinned entrypoint, the served document can silently flip to one of those pages instead. This pins `project.render` to the entrypoint in `pqr`, `quarto-stock-report-python`, and `quarto-stock-report-r`, which were each rendering a stray `CHANGELOG` page.

Also removes `connect-extension.qmd` from `quarto-document` and `quarto-stock-report-python`, a superseded legacy gallery-metadata file that's no longer used.

The CI guard for this class of bug is in a follow-up PR, stacked on this one.


dotNomad (approved):
This is looking good. Just added the `_quarto.yml` render change to `quarto-document` to make it consistent even though it wasn't strictly needed yet.

PR #469, opened 2026-09-11
fix: point settings instructions at the Runtime tab

### LINKED ISSUE #464: update instructions on how to configure settings in extensions/examples
https://github.com/posit-dev/connect/pull/41206 changes how content settings appear, so we need to go through all of the examples/extensions that require configuring the settings and update the text there accordingly (the instructions that explain what you need to do in the settings are now out of date).

jmwoliver:
Connect 2026.08.0 renamed the content settings **Advanced** tab to **Runtime** and moved Integrations there from **Access**, so the setup instructions in both MCP examples pointed at tabs that no longer exist. This updates the two READMEs, the MCP server's landing page and its `connect_whoami` error message, and the chat app's setup screen. Each one names the old tab in parentheses, since both examples still support Connect 2025.04.0 and newer.

Also reframes the "keep it warm" note: Connect 2026.05.0 and newer detect MCP content and set Min processes to 1 on first deploy, so setting it by hand is only needed on older versions.

Both extensions bump to 0.0.9, with refreshed manifest checksums for the edited files, so the new text actually ships to the gallery.

Let me know if the wording isn't right for these! I wasn't sure if we should reference specific settings from specific Connect versions or not.

Closes #464

### DIFF
--- a/extensions/simple-mcp-server/index.html.jinja
+++ b/extensions/simple-mcp-server/index.html.jinja
@@ -182,9 +182,11 @@
             <span id="copyStatus"></span>
         </div>
 
-        <p><strong>Whichever way you use it, keep the server warm:</strong> in the content's settings, on
-        the "Advanced" tab, set "Min Processes" to 1 or more under "Process Settings", so it responds
-        without a cold-start delay. For more information, see the
+        <p><strong>Whichever way you use it, the server stays warm:</strong> Connect 2026.05.0 and newer
+        recognize this as an MCP server and set "Min Processes" to 1 on first deploy, so it responds
+        without a cold-start delay. To check or change it, or to set it yourself on an older Connect,
+        see "Process Settings" on the "Runtime" tab of the content's settings ("Advanced" before
+        Connect 2026.08.0). For more information, see the
         <a href="https://docs.posit.co/connect/user/content-settings/index.html#process-configurations" target="_blank" rel="noopener">content process configuration documentation</a>.</p>
 
         <h2>1. Use it from the companion chat</h2>
@@ -202,7 +204,8 @@
         {% else %}
         <div class="info-box">
             <strong>"Connect Visitor API Key" integration required.</strong> In the content's settings, on
-            the "Access" tab, add a "Connect Visitor API Key" integration under "Integrations". The
+            the "Runtime" tab ("Access" before Connect 2026.08.0), add a "Connect Visitor API Key"
+            integration under "Integrations". The
             <code>connect_whoami</code> tool needs it to identify the viewer. Once it's added, you can confirm
             that it works here: this page will greet you by name, resolved from your session token with no API
             key. For more information, see the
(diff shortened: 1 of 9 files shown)


extensions/simple-mcp-server/index.html.jinja
```diff
@@ -182,9 +182,11 @@
             <span id="copyStatus"></span>
         </div>
 
-        <p><strong>Whichever way you use it, keep the server warm:</strong> in the content's settings, on
-        the "Advanced" tab, set "Min Processes" to 1 or more under "Process Settings", so it responds
-        without a cold-start delay. For more information, see the
+        <p><strong>Whichever way you use it, the server stays warm:</strong> Connect 2026.05.0 and newer
```
joshyam-k, line 185: I wonder if a different approach we could take here instead of doing this dance around the tab names for various versions of connect is just not mention the tab names at all. This could become:

> To check or change it, or to set it yourself on an older Connect,
    see "Process Settings" in the content's settings.

Not that I foresee us shuffling up the setting panel tabs again soon, but if we did it would make this text that much more complicated since we'd have to explain around all three versions. What do you think?

joshyam-k (commented):
One idea around how we can avoid splitting out to the two different versions of the settings tabs each time. I avoided leaving the same comment across each portion of the diff since I'm not positive it's the right approach yet! Let me know what you think, but thanks for opening this PR and I agree that there's some updating needed here.

extensions/simple-mcp-server/index.html.jinja
jmwoliver: Thanks for the feedback! I'm not sure, I could see it going either way. As a new person using Connect trying out gallery content, I could see someone being confused by just "Process Settings" and would have to click around to find out where that is.

A few ideas/questions since I haven't done anything in this repo before:
- what have we historically done for gallery content when making new versions that need to work across Connect versions?
- what if we only mention the new Runtime tab without mentioning old setting names / Connect versions?
- the gallery content is versioned, is it possible to pin and map gallery content versions to Connect versions so the old extension version is served on old Connect versions and this new extension version is served for new Connect versions?

I also don't know the best approach here, happy to keep discussing to find out the right thing to do!

joshyam-k: Sorry for the delay getting back to this! Given that the gallery is relatively new I think there isn't really a precedent for working through these types of issues. I really like the idea of mapping gallery content versions to Connect versions, but that unfortunately doesn't exist yet. I think maybe for now to unblock this we just stick with what you have here and then as a follow up I'll work on finding a better standard for keeping gallery content in sync with the appropriate Connect version.

joshyam-k (approved):

extensions/simple-mcp-server/index.html.jinja
jmwoliver: No worries! I'll go ahead and merge this and we can think through the versioning more in the future.

